PR #3, opened 2018-05-07
Add .prow.sh script

jlebon:
This works in tandem with the changes in
https://github.com/openshift/release/pull/841.

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -xeuo pipefail
+
+yum -y install jq
+for jsonfile in *.json; do
+    echo "Checking JSON syntax for $jsonfile"
+    jq < $jsonfile . >/dev/null
+done
+
+imagebuild -privileged .


.prow.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -xeuo pipefail
+
+yum -y install jq
+for jsonfile in *.json; do
+    jq < $jsonfile >/dev/null
```
peterbaouoft: Hi, might be a dumb question, but do you mind elaborating a bit what does this line do?(Does it just check the validity of each json files). I was also confused about the usage of `jq < $file` vs `jq $file`, do you have pointers to look at for usage of `<` in bash? Thanks, =)

jlebon: Exactly! It'll check that the JSON files are valid. Just a simple check that can be used for now until we have more sophisticated PR tests. `<` means feeding the file as the program's standard input: http://tldp.org/HOWTO/Bash-Prog-Intro-HOWTO-3.html

.prow.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -xeuo pipefail
+
+yum -y install jq
+for jsonfile in *.json; do
+    echo "Checking JSON syntax for $jsonfile"
+    jq < $jsonfile . >/dev/null
```
jlebon: The `.` here is needed to work around a bug where `jq` will refuse to work if redirecting output to something that's not a TTY: https://github.com/stedolan/jq/issues/1351.

PR #5, opened 2018-05-08
Add Jenkinsfile compose pipeline

jlebon:
Add a Jenkinsfile that runs the treecompose and publishes it. Right now,
this is only running the `rhcos` variant, so it's publishing to some
place private. The next steps are to also build cloud images, and to
also build the `origin` CentOS-based variant and push that out somewhere
public.

I brought back the RHEL repos here for simplicity; I'd like to keep as
much of the logic as possible right here. Though I'm open to changing
this back.


Jenkinsfile
```diff
@@ -0,0 +1,73 @@
+def DOCKER_IMG = "registry.fedoraproject.org/fedora:28"
+def DOCKER_ARGS = "--net=host -v /srv:/srv --privileged"
+
+// this var conveniently refers to a location on the server as well as the local dir we sync to/from
+def repo = "${env.ARTIFACT_SERVER_DIR}/repo"
```
peterbaouoft: quick question, where does this env variable get defined? Is it on the jenkins side? =)

jlebon: That's correct, this is defined on the Jenkins side since it depends on where it runs. We'll need to make this more flexible if we want to support e.g. GCS buckets.

host-base.json
```diff
@@ -110,7 +110,8 @@
         "sg3_utils",
         "glusterfs-fuse",
         "dnsmasq",
-        "oci-umount"
+        "oci-umount",
+        "origin-node"
```
ashcrow: :+1: for adding ``origin-node`` at this stage in development. Reminder to us all that this will be replaced with a node that is built upon upstream merges later.

Jenkinsfile
```diff
@@ -0,0 +1,73 @@
...
+}
+def manifest = "host-${env.BUILD_TYPE}.json"
+
+node(env.NODE) {
+    checkout scm
+
+    docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
+        stage("Provision") {
+            sh "dnf install -y git rpm-ostree rsync openssh-clients"
+            sh "cp RPM-GPG-* /etc/pki/rpm-gpg/"
```
ashcrow: This looks similar to [what's happening in the Dockerfile](https://github.com/openshift/os/blob/master/Dockerfile#L3) but in this case it looks like it's doing it on the node. Am I reading this right?

jlebon: That's correct. Since we're running in a Fedora container, we're not gonna have those keys available. I just copy them all since I'd like to generalize the Jenkinsfile to build both the `rhcos` and the `origin` variants.

host-base.json
jlebon: To clarify, `origin-node` was in the main manifest before. This is a small revert of #2, which changed that to `atomic-openshift-node` for RHEL. So we might need to revert this -- the OpenShift repo we're using right now does have it though.

PR #6, opened 2018-05-09
Initial overlay

ashcrow:
Place to build on for generating RPMs for each upstream repo that we then layer into the compose.


overlay.yml
```diff
@@ -0,0 +1,24 @@
...
+distgit:
+  prefix: fedorapkgs
+  branch: master
+  
+root:
+  mock: fedora-28-$arch
+
+components:
+  - src: github:projectatomic/docker
+    branch: docker-1.13.1-rhel
```
miabbott: @ashcrow, @peterbaouoft and myself tested this overlay file and it looks like we need to override the version here with `override-version: "1.13.1"`

ashcrow: Updated

overlay.yml
```diff
@@ -0,0 +1,25 @@
+aliases: 
+  - name: github
+    url: https://github.com/
+
+  - name: fedorapkgs
```
peterbaouoft: Minor: seems like if we plan to build all rpms from git repos, we don't need the fedorapkgs(as we are not fetching from it)? I happened to not include the cert and it passes. Only a minor tho :).

ashcrow: The fedorapkgs is used by the distgit section but you're correct that if we had spec files in each repo we wouldn't need it.

ashcrow:
Can someone else lgtm?

PR #9, opened 2018-05-10
rdgo: Switch to centos 7 mock root

ashcrow:
Tested by building ``cri-o``. No issues.


miabbott:
> Nice! Have you tried the other components yet? If it works, then WDYT of just merging this overlay into the CAHC one for now? E.g. we'll also probably want git master rpm-ostree/ostree too, which is already building there. Though I agree there's an appeal to keeping this definition here.

You know, merging with CAHC would probably get us to PoC status faster than building out our own jobs.  I'm not opposed to this idea.  cc: @mike-nguyen

PR #11, opened 2018-05-11
rdgo: Add openshift repo and required mock packages

ashcrow:


ashcrow:
Set to hold while I try to find the proper way to **not** build the clients in the rpm for other arch/os's as it causes failures.

@dgoodwin if you have any pointers let me know :smile:

peterbaouoft:
Thanks @ashcrow  :), and for more context, we are using [origin.spec](https://github.com/openshift/origin/blob/master/origin.spec) for building package, 
 below is the output when running build cross on a fedora mock: Looks like they have `OS_ONLY_BUILD_PLATFORMS=linux/amd64` set so we skip other distro builds. 
```
hack/build-cross.sh
++ Building go targets for linux/amd64: images/pod
++ Building go targets for linux/amd64: pkg/network/sdn-cni-plugin vendor/github.com/containernetworking/plugins/plugins/ipam/host-local vendor/github.com/containernetworking/plugins/plugins/main/loopback
++ Building go targets for linux/amd64: cmd/hypershift cmd/openshift cmd/oc cmd/oadm cmd/template-service-broker cmd/openshift-node-config vendor/k8s.io/kubernetes/cmd/hyperkube
++ Building go targets for linux/amd64: test/extended/extended.test
hack/build-cross.sh took 458 seconds
+ OS_ONLY_BUILD_PLATFORMS=linux/amd64
+ OS_GIT_VERSION=
+ OS_GIT_COMMIT=
+ OS_GIT_MAJOR=
+ OS_GIT_MINOR=
+ OS_GIT_TREE_STATE=
+ hack/build-go.sh vendor/github.com/onsi/ginkgo/ginkgo
++ Building go targets for linux/amd64: vendor/github.com/onsi/ginkgo/ginkgo
```

But for centos mock, when building cross, we are building for all distros, and that leads to an error due to an unbound variable: (shown below).
```
+ cd /builddir/build/BUILD
+ cd origin-v3.10.0-alpha.0-1098-7cfc7eaadd23a5afc722b6d0dfcb7eddaf6caf63
+ OS_GIT_VERSION=
+ OS_GIT_COMMIT=
+ OS_GIT_MAJOR=
+ OS_GIT_MINOR=
+ OS_GIT_TREE_STATE=
+ OS_BUILD_RELEASE_ARCHIVES=n
+ make build-cross
hack/build-cross.sh
++ Building go targets for linux/amd64: images/pod
++ Building go targets for linux/amd64: pkg/network/sdn-cni-plugin vendor/github.com/containernetworking/plugins/plugins/ipam/host-local vendor/github.com/containernetworking/plugins/plugins/main/loopback
++ Building go targets for linux/amd64: cmd/hypershift cmd/openshift cmd/oc cmd/oadm cmd/template-service-broker cmd/openshift-node-config vendor/k8s.io/kubernetes/cmd/hyperkube
++ Building go targets for darwin/amd64: cmd/hypershift cmd/openshift cmd/oc cmd/oadm cmd/template-service-broker cmd/openshift-node-config vendor/k8s.io/kubernetes/cmd/hyperkube
++ Building go targets for windows/amd64: cmd/hypershift cmd/openshift cmd/oc cmd/oadm cmd/template-service-broker cmd/openshift-node-config vendor/k8s.io/kubernetes/cmd/hyperkube
/builddir/build/BUILD/origin-v3.10.0-alpha.0-1098-7cfc7eaadd23a5afc722b6d0dfcb7eddaf6caf63/hack/lib/constants.sh: line 253: OS_GIT_PATCH: unbound variable
[ERROR] [13:46:35+0000] PID 13387: hack/lib/build/binaries.sh:150: `local -a binaries=("$@")` exited with status 1.
[INFO] [13:46:35+0000] 		Stack Trace: 
[INFO] [13:46:35+0000] 		  1: hack/lib/build/binaries.sh:150: `local -a binaries=("$@")`
[INFO] [13:46:35+0000] 		  2: hack/build-cross.sh:76: os::build::build_binaries
[INFO] [13:46:35+0000]   Exiting with code 1.
make: *** [build-cross] Error 1
```

I am thinking it is because OS_GIT_PATCH was never set in origin.spec, and therefore not get passed in with `os_git_vars` defined [here](https://github.com/openshift/origin/blob/master/origin.spec#L25). And eventually, when executing, the code fails in [here](https://github.com/openshift/origin/blob/master/hack/lib/constants.sh#L253) due to `OS_GIT_PATCH` not being set.

ashcrow:
To keep this updated:

```OS_ONLY_BUILD_PLATFORMS=linux/amd64``` sadly didn't solve the problem. It seems like the only way around this will be to make a PR to ``origin`` with a way to skip redistr via a variable.

**Comment by peterbaouoft** [comment]
@ashcrow, thanks for trying it out =). Meanwhile, I saw this commit https://github.com/openshift/origin/commit/793ea5bb42c33aa409aba9e86f03a655e5be27bd. Sadly, it is related to tito atm after some digging. An example usage of tito with the option set is the following: https://github.com/openshift/origin/blob/12f0ec1763f15dd30ed786a023dc7b92df74ab97/hack/build-rpm-release.sh#L37

**Comment by peterbaouoft** [comment]
Update: those commits may be useful :)
https://github.com/openshift/origin/commit/3f055a9d1199b669997cedfb851b55379878e7d3
https://github.com/openshift/origin/commit/663a36af5c3df42299d5246e9e1654e45cfb70a5
and `make build-rpms-redistributable` in origin repo can create rpms(without relearning all tito related syntaxes) 

**Current status**: i am still looking to see if we can pass those flags without using tito, tito is one known way to do so. I think I will ask on openshift-origin to see it works.

**Comment by miabbott** [comment]
@ashcrow @peterbaouoft Is there a known good commit of the spec file that doesn't include changes which are breaking the build for us?

We might be able to pin the spec file to that commit like so:

```
- src: github:openshift/origin
  spec: internal
    freeze:  <commit id>
```

I know that kind of syntax is supported by the `distgit` key - https://github.com/cgwalters/rpmdistro-gitoverlay/blob/master/doc/example-overlay.yml#L54

**Comment by peterbaouoft** [comment]
@miabbott, https://github.com/openshift/origin/commit/c8c201d2e965bca2fcfd1e3a036b49eb50e4d9e6, seems to be the culprit. But ..it is when openshift is at 3.6.0. If we try to track the git upstream changes, we would want this build to be fixed anyways right? 

I can quickly create a hacky version of rpmdistro-gitoverlay for you to try it out tho :).

**Comment by ashcrow** [comment]
@miabbott I think the answer lies in @peterbaouoft's work to update `rpmdistro-overlay` to allow ``define``.

**Comment by peterbaouoft** [comment]
@ashcrow, https://github.com/cgwalters/rpmdistro-gitoverlay/issues/44, I am thinking about this.

**Comment by ashcrow** [comment]
@jlebon the ``generate-openshift-repo`` would need to run before the compose, but it would provide us with the latest builds. PTAL.

File: centos-7-and-extras-x86_64.cfg [hunk 1]
```diff
@@ -1,7 +1,7 @@
 config_opts['root'] = 'centos-and-extras-7-x86_64'
 config_opts['target_arch'] = 'x86_64'
 config_opts['legal_host_arches'] = ('x86_64',)
-config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils system-release findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar yum-utils unzip util-linux which xz python'
+config_opts['chroot_setup_cmd'] = 'install bash bzip2 coreutils cpio diffutils system-release findutils gawk gcc gcc-c++ grep gzip info make patch redhat-rpm-config rpm-build sed shadow-utils tar yum-utils unzip util-linux which xz python goversioninfo'
```
peterbaouoft: do we still need this if we don't build origin rpm?

host-base.json
```diff
@@ -5,6 +5,7 @@
     "repos": [
         "atomic-centos-continuous",
         "origin-repo",
+	"openshift",
```
peterbaouoft: minor: an extra space here?

centos-7-and-extras-x86_64.cfg
```diff
@@ -45,4 +45,13 @@ failovermethod=priority
 exclude=docker-python python-docker-py
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-7
 gpgcheck=1
+
+[openshift]
+name=openshift
```
peterbaouoft: ditto.

cgwalters: I don't think it's needed in the base buildroot.

generate-openshift-repo
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
cgwalters: `set -xeuo pipefail`

cgwalters: Also maybe we should have a `make refresh` that calls this?

centos-7-and-extras-x86_64.cfg
ashcrow: We do not. Will remove.

centos-7-and-extras-x86_64.cfg
ashcrow: Will remove.

ashcrow:
Updated per reviews.

PR #12, opened 2018-05-11
manifest: Add CAHC repo

jlebon:
In preparation for adding more of the packages of interest there[1]. But
even right now, this will allow us to track ostree and rpm-ostree git
master.

[1] https://github.com/CentOS/sig-atomic-buildscripts/pull/326

### DIFF
--- a/host-base.json
+++ b/host-base.json
@@ -3,7 +3,7 @@
     "osname": "openshift-host",
     "ref": "openshift/3.10/x86_64/os",
     "repos": [
-        "ostree-master",
+        "atomic-centos-continuous",
         "origin-repo",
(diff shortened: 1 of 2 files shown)


host-base.json
```diff
@@ -5,7 +5,8 @@
     "repos": [
         "ostree-master",
         "origin-repo",
-        "dustymabe-ignition"
+        "dustymabe-ignition",
```
miabbott: Do we want these in the base or in `host-rhcos.json`?

jlebon: I think these are all RPMs that would be shared between the two variants, so it makes more sense in the base?

PR #13, opened 2018-05-14
treecompose: Generate openshift repo before composing

jlebon:
The treecompose has been erroring out on the fact that it's not finding
an "openshift" repo. Fix this by running `repo-refresh` during the
provisioning phase.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,9 @@ syntax-check:
 	done
 
 .PHONY: container
-container:
+container: repo-refresh
 	imagebuild -privileged .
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -6,10 +6,10 @@ syntax-check:
 		echo "OK"; \
 	done
 
-.PHONY: container
+.PHONY: container repo-refresh
```
ashcrow: Why is repo-refresh twice ``PHONY``?

jlebon: Thanks, fixed!

PR #14, opened 2018-05-14
rdgo: Set cri-o to use master and force version

ashcrow:


overlay.yml
```diff
@@ -20,8 +20,7 @@ components:
 
   # NOTE: master=1.11.0 which requires a different spec
```
jlebon: We should delete this note now, right?

ashcrow: Yeah :-)

ashcrow: done

overlay.yml
```diff
@@ -18,10 +18,8 @@ components:
     override-version: "1.13.1"
     branch: docker-1.13.1-rhel
 
-  # NOTE: master=1.11.0 which requires a different spec
   - src: github:kubernetes-incubator/cri-o
-    override-version: "1.10.1"
-    branch: release-1.10
+    override-version: "1.11"
```
cgwalters: Unfortunate we still need this; I guess it's an rdgo issue where it should better handle projects which do releases from branches.  For my own reference, today cri-o git master:

```
$ git describe --tags --always --abbrev=42
v1.9.0-beta.2-433-g55a3daaf3dcfe290aba6c621b4fd73db7709072e
```

PR #15, opened 2018-05-14
rdgo: Add ignition via fake distgit

ashcrow:


cgwalters:
We need to delete the ignition COPR repo too right?

ashcrow:
Yeah, we should. Doing so now.

overlay.yml
```diff
@@ -21,5 +21,10 @@ components:
   - src: github:kubernetes-incubator/cri-o
     override-version: "1.11"
 
-#   - src: github:coreos/ignition
-#   TODO: use copr for distgit here OR get spec in the ignition repo
+  - src: github:coreos/ignition
+    override-version: "0.24.0"
```
jlebon: We shouldn't need this, right? Looks like the tag is on the master branch. Would be good to keep those overrides to a minimum to keep maintenance low.

jlebon:
```
$ git describe --tags --always --abbrev=42
v0.24.0-18-g042a00e218eb4a49eb22f053077c4e3152c8caed
```

ashcrow:
Removed from the json file. Leaving the actual repo file there since it's still used in the OTA stuff for the moment. We'll rework that to take advantage of the stream later.

overlay.yml
ashcrow: I'll double check but you're likely correct.

ashcrow: You are correct. I've removed it.

jlebon:
Nice! /lgtm

cgwalters:
@jlebon /lgtm needs to be on a new line I think

jlebon:
Ahh, let me try that again.
/lgtm

PR #16, opened 2018-05-14
Add a rdgo pipeline

mike-nguyen:
Add yet another Jenkins pipeline `Jenkinsfile.overlay` for rdgo built
packages.

### DIFF
--- a/Jenkinsfile.rdgo
+++ b/Jenkinsfile.rdgo
@@ -30,16 +30,24 @@ node(env.NODE) {
             }
         }
 
-        stage("Build Overlay Packages") {
+        stage("Setup and Initialize RDGO") {
             sh """
                 ln -sf $WORKSPACE/overlay.yml ${rdgo}/
+                cd ${rdgo}
+                rm -f rdgo.stamp
                 rpmdistro-gitoverlay init
-                rpmdistro-gitoverlay resolve --fetch-all
-                rpmdistro-gitoverlay build --touch-if-changed rdgo.stamp --logdir=log
             """
         }
 
-        if (!fileExists("${rdgo}/rdgo.stamp})) {
+        stage("Resolve and Fetch Sources") {
+            sh "cd ${rdgo} && rpmdistro-gitoverlay resolve --fetch-all"
+        }
+
+        stage("Build Overlay Packages") {
+            sh "cd ${rdgo} && rpmdistro-gitoverlay build --touch-if-changed rdgo.stamp --logdir=log"
+        }
+
+        if (!fileExists("${rdgo}/rdgo.stamp")) {
             currentBuild.result = 'SUCCESS'
             return
         }
--- a/Jenkinsfile.rdgo
+++ b/Jenkinsfile.rdgo
@@ -8,10 +8,12 @@ node(env.NODE) {
 
     docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
         stage("Provision") {
-            sh "dnf install -y git rsync openssh-clients dnf-plugins-core fedpkg"
-            sh "cp RPM-GPG-* /etc/pki/rpm-gpg/"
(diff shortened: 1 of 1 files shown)


Jenkinsfile.overlay
```diff
@@ -0,0 +1,53 @@
...
+            sh "ln -sf $WORKSPACE/overlay.yml ${rdgo}/"
+            if (!fileExists("${rdgo}/src")) {
+                sh "cd ${rdgo} && rpmdistro-gitoverlay init"
+            }
+            sh "cd ${rdgo} && rpmdistro-gitoverlay resolve --fetch-all"
+            sh "cd ${rdgo} && rpmdistro-gitoverlay build --touch-if-changed rdgo.stamp --logdir=log"
+        }
+
+        stage("Sync Out") {
+            if (fileExists("${rdgo}/rdgo.stamp")) {
```
jlebon: How about we move this out of the stage:

```
if (!fileExists("${rdgo}/rdgo.stamp")) {
        currentBuild.result = 'SUCCESS'
        return
}

stage("Sync Out") {
...
}
```

This will make it more obvious in the pipeline view which runs actually had builds.

Jenkinsfile.overlay
```diff
@@ -0,0 +1,53 @@
...
+                           -o StrictHostKeyChecking=no' \
+                        ${env.ARTIFACT_SERVER}:${rdgo}/ ${rdgo}
+                """
+            }
+        }
+
+        stage("Build Overlay Packages") {
+            sh "ln -sf $WORKSPACE/overlay.yml ${rdgo}/"
+            if (!fileExists("${rdgo}/src")) {
+                sh "cd ${rdgo} && rpmdistro-gitoverlay init"
```
jlebon: Hmm, this operation is already idempotent, so let's just take it out of the `if`? I think we start from scratch anyway, so we'd always hit this path.

cgwalters: Depends on if we have a previous build.

Jenkinsfile.rdgo
```diff
@@ -0,0 +1,53 @@
...
+                """
+            }
+        }
+
+        stage("Build Overlay Packages") {
+            sh "ln -sf $WORKSPACE/overlay.yml ${rdgo}/"
+            if (!fileExists("${rdgo}/src")) {
+                sh "cd ${rdgo} && rpmdistro-gitoverlay init"
+            }
+            sh "cd ${rdgo} && rpmdistro-gitoverlay resolve --fetch-all"
```
cgwalters: I'd make this a separate stage.  (And this implies a bit making the `ln -sf` and `init` bits part of "Sync In and Init" or so)

The rationale here is that source fetching can be distinct from building.

Jenkinsfile.rdgo
```diff
@@ -0,0 +1,53 @@
...
+            sh "cd ${rdgo} && rpmdistro-gitoverlay build --touch-if-changed rdgo.stamp --logdir=log"
+        }
+
+        stage("Sync Out") {
+            if (fileExists("${rdgo}/rdgo.stamp")) {
```
cgwalters: I think this could use [a when conditional](https://jenkins.io/doc/book/pipeline/syntax/#when) but eh, fine as is now.

Jenkinsfile.overlay
mike-nguyen: The operation is idempotent so I removed the check for `${rdgo}/src` and just ran `rpmdistro-gitoverlay init` as suggested.

Jenkinsfile.rdgo
mike-nguyen: I ended up separating into three stages so it is easier to distinguish which part failed in the pipeline view.

mike-nguyen:
Pushed a fixup  with changes for the comments.

jlebon:
Let's do this!
/lgtm

PR #18, opened 2018-05-15
Makefile more

cgwalters:


peterbaouoft:
/me is testing the content out :). Also, seems like rebase is needed after #17 has been merged?

jlebon:
LGTM too, though we should be working towards embedding an existing commit into the container, right?

jlebon:
I guess we want both: new tree in a container for testing out stuff, but slurping up an existing tree for actual pipeline integration.

ashcrow:
> I guess we want both: new tree in a container for testing out stuff, but slurping up an existing tree for actual pipeline integration.

I'd like the same content used in both eventually, but this works for the moment.

peterbaouoft:
Ah, nvm, the same commit was included from #17, and looks like robot can recognize & ignore the same commit. Interesting and good to know =).

PR #20, opened 2018-05-15
Add credentials to sync out

mike-nguyen:
The sync out was missing credentials for the rsync function and also
fixed a typo with the UserKnowHostsFile.


jlebon:
Thanks!
/lgtm

jlebon:
Hmm, though note we don't have homu here. We should look into whether prow supports autosquashing. Anyway, we already have a couple of `fixup!` commits in master now.
/lgtm

cgwalters:
Prow doesn't do `fixup!` BTW unfortunately.

PR #21, opened 2018-05-15
cloud.ks: Switch to RHELAH kickstart

jlebon:
I'm not sure why I went with the Fedora kickstart in the original patch
that introduced `cloud.ks`. Given that the userspace is
RHEL/CentOS-based, it makes more sense to iterate from that instead.

I did do a few minor tweaks from the RHELAH one:

- drop `%pre` hack for mounting ostree repo locally
- drop `rsyslog`, re-enable persistent journal
- drop ostree remote manipulations for subscription-manager


cloud.ks
```diff
@@ -11,118 +10,98 @@ rootpw --lock --iscrypted locked
...
-# https://pagure.io/atomic-wg/issue/299
-reqpart --add-boot
+clearpart --initlabel --all
+bootloader --timeout=1 --append="no_timer_check console=tty1 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0"
+
+part /boot --size=300 --fstype="xfs"
 part pv.01 --grow
 volgroup atomicos pv.01
-# Start from 3GB as we did before, since we just need a size.  But we do --grow to fill all space.
-logvol / --size=3000 --grow --fstype="xfs" --name=root --vgname=atomicos
```
jlebon: Should we still stick with the Fedora storage configuration here? I *think* they're equivalent, except in RHEL we rely on c-s-s to do the growing for us in `%post`?

cloud.ks
```diff
@@ -11,118 +10,98 @@ rootpw --lock --iscrypted locked
...
-#
-# Use reqpart to create hardware platform specific partitions
-# https://pagure.io/atomic-wg/issue/299
-reqpart --add-boot
+clearpart --initlabel --all
+bootloader --timeout=1 --append="no_timer_check console=tty1 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0"
+
+part /boot --size=300 --fstype="xfs"
 part pv.01 --grow
 volgroup atomicos pv.01
```
cgwalters: While we have the patient open, `s/atomicos/coreos`?

cgwalters: And let's not block on this but @bgilbert and @ajeddeloh noted that Ignition today does not support LVM.

cloud.ks
```diff
@@ -11,118 +10,98 @@ rootpw --lock --iscrypted locked
...
-# Anaconda is writing an /etc/resolv.conf from the install environment.
-# The system should start out with an empty file, otherwise cloud-init
-# will try to use this information and may error:
-# https://bugs.launchpad.net/cloud-init/+bug/1670052
+# Additional drivers for vmware and hyperv
+echo 'add_drivers+="mptspi vmw_pvscsi "'    > /etc/dracut.conf.d/vmware-fusion-drivers.conf
+echo 'add_drivers+="hv_storvsc hv_netvsc "' > /etc/dracut.conf.d/hyperv-drivers.conf
+# Rerun dracut for the installed kernel (not the running kernel):
+KERNEL_VERSION=$(rpm -q kernel --qf '%{V}-%{R}.%{arch}\n')
+dracut -f /boot/initramfs-$KERNEL_VERSION.img $KERNEL_VERSION
```
cgwalters: This whole section never worked...but I guess we should fix that upstream first.

cloud.ks
ashcrow: For now I'd say stick with the RHEL way since we know it works as is with CentOS and RHEL. We can enhance it once we have a base to work off.

cgwalters: Both use `GROWPART=true` with c-s-s.

ashcrow:
Will wait for @cgwalters to :+1: and merge.

cloud.ks
jlebon: Right but in Fedora, does `GROWPART=true` actually do anything given that we already have `--grow`?

jlebon:
> While we have the patient open, `s/atomicos/coreos`?

Done! I just rolled it into the commit -- let's just move away from `fixup!` for this repo for now.

PR #23, opened 2018-05-15
base: Enable ostree-remount

cgwalters:
See https://src.fedoraproject.org/rpms/ostree/c/82d1f941ed1504840a63b5635e1051ccd2774b07?branch=master
We need a new `-release` package, in the meantime let's do this
quick hack.

### DIFF
--- a/host-base.json
+++ b/host-base.json
@@ -125,6 +125,7 @@
     "units": [
         "docker.service",
         "tuned.service",
+        "ostree-remount.service",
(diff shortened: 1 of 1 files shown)


jlebon:
Ahh I see, we're pulling in `redhat-release-server` instead of `redhat-release-atomic-host`, which does have the fix for this.

Anyway, that works for me! Good catch!
/lgtm

PR #25, opened 2018-05-15
Add rhcos-rdgo rpmmd repo

cgwalters:
Otherwise we won't have ignition.


jlebon:
We discussed hosting this publicly in S3, right? Working on that right now.

cgwalters:
OK, though that's an orthogonal improvement right?

jlebon:
https://github.com/openshift/os/pull/27

ashcrow:
Since #27 merged is this safe to close?

PR #26, opened 2018-05-15
host: Add origin* packages

cgwalters:
We need hyperkube which contains the kubelet, and we really
might as well ship the clients.

### DIFF
--- a/host-base.json
+++ b/host-base.json
@@ -110,7 +110,9 @@
         "glusterfs-fuse",
         "dnsmasq",
         "oci-umount",
-        "origin-node"
+        "origin-node",
+        "origin-hyperkube",
+        "origin-clients"
     ],
     "remove-from-packages": [
         [


host-base.json
```diff
@@ -110,7 +110,9 @@
         "glusterfs-fuse",
         "dnsmasq",
         "oci-umount",
-        "origin-node"
+        "origin-node",
+        "origin-hyperkube",
+        "origin"
```
smarterclayton: Use origin-clients, not origin.  Origin brings in 800M of stuff.

Hyperkube is already brought in via node.

cgwalters: OK, done :arrow_down:

jlebon: Though `origin` is getting pulled in right now. Assuming those origin packages are built from the `origin.spec` in openshift/origin, then I'd guess it's https://github.com/openshift/origin/blob/e00803cbc67b24a154a9dd74672e46f3ebe9e249/origin.spec#L138. Should that be `origin-hyperkube` then?

jlebon: But then again, we're not pulling in `origin-hyperkube`, so this is probably not the spec we're using.

ashcrow: @jlebon do you mean ``origin.spec`` is not producing ``origin-hyperkube``?

jlebon: It does. I'm just referring to Clayton's comment that we don't want `origin`. We're currently pulling it in though, so even if we remove it from this patch, we'll probably still be pulling it in.

jlebon:
```
$ sudo dnf repoquery --repofrompath=asdf,https://storage.googleapis.com/origin-ci-test/logs/test_branch_origin_extended_conformance_gce/3033/artifacts/rpms --requires origin-node | grep origin
origin = 3.10.0-0.alpha.0.1141.6c675da
origin = 3.6.0-1.fc27
origin = 3.9.0-1.fc28
origin-hyperkube = 3.10.0-0.alpha.0.1141.6c675da
tuned-profiles-origin-node = 3.6.0-1.fc27
```

smarterclayton: The only package you should be pulling in is origin-node, which does the right thing for you.

jlebon: @smarterclayton Which spec file is used for those `https://storage.googleapis.com/origin-ci-test/releases/openshift/origin` RPMs?

ashcrow: @smarterclayton hrm. That is what we were doing before :slightly_frowning_face:.

PR #27, opened 2018-05-15
rdgo: Publish repo to S3

jlebon:
These are based on the git master of various projects and thus should
just be made public for easier consumption.

### DIFF
--- a/Jenkinsfile.rdgo
+++ b/Jenkinsfile.rdgo
@@ -9,7 +9,7 @@ node(env.NODE) {
     docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
         stage("Provision") {
             sh """
-                dnf install -y git rsync openssh-clients dnf-plugins-core fedpkg dnf-utils
+                dnf install -y git rsync openssh-clients dnf-plugins-core fedpkg dnf-utils awscli
                 cp RPM-GPG-* /etc/pki/rpm-gpg/
                 dnf copr -y enable walters/buildtools-fedora
                 dnf install -y rpmdistro-gitoverlay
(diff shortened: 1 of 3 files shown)


Jenkinsfile.rdgo
```diff
@@ -13,6 +13,7 @@ node(env.NODE) {
                 cp RPM-GPG-* /etc/pki/rpm-gpg/
                 dnf copr -y enable walters/buildtools-fedora
                 dnf install -y rpmdistro-gitoverlay
+                pip3 install awscli==1.15.20
```
cgwalters: FWIW this also exists as `awscli` as an RPM.

Jenkinsfile.rdgo
```diff
@@ -63,6 +64,14 @@ node(env.NODE) {
                         ${rdgo}/ ${env.ARTIFACT_SERVER}:${rdgo}
                 """
             }
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                              credentialsId: env['AWS_CREDENTIALS'],
+                              accessKeyVariable: 'AWS_ACCESS_KEY_ID',
+                              secretKeyVariable: 'AWS_SECRET_ACCESS_KEY']]) {
+                sh """
+                    aws s3 sync ${rdgo}/build/ s3://aos-ci/rhcos/rdgo
```
cgwalters: I notice we aren't using `--delete`...that's probably OK short term.  But I feel like we should set up something like Pulp or whatever to push to - that also applies to origin RPMs.

cgwalters:
Basically seems OK, but leaving unmerged in case you want to change anything.  If not then I'll lgtm.

Jenkinsfile.rdgo
jlebon: Ahh, nice. I had checked `python-awscli` and `python3-awscli`, but not `awscli` directly.

Jenkinsfile.rdgo
jlebon: Good point, the CLI supports `--delete`, so added in :arrow_down:

jlebon:
I also dropped the

```
accessKeyVariable: 'AWS_ACCESS_KEY_ID',
secretKeyVariable: 'AWS_SECRET_ACCESS_KEY'
```

bits since those are the default values already.

PR #28, opened 2018-05-15
host-base.json: Add cri-o

ashcrow:


jlebon:
Looks like this is causing issues:

```
Resolving dependencies... failed
0. nothing provides containernetworking-plugins >= 0.5.2-7 needed by cri-o-1.11-72cb4483f7564cb057d9e22174090ec222071c37.ce0b9dc836a9b3a878f56f7480d802c391103c9c.el7.centos.x86_64
make: *** [Makefile:33: rpmostree-compose] Error 1
```

ashcrow:
If need be we can revert this commit until we add in the missing dep to the rpmdistro-overlay build.

PR #31, opened 2018-05-15
Dockerfile: Make nginx work again

cgwalters:
I broke this with the previous changes.  We do need to install
nginx in the final target image, not the build container.

Also add an `EXPOSE` so `docker run -P` etc. does something useful.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,14 +4,16 @@ FROM registry.fedoraproject.org/fedora:28 AS build
 COPY RPM-GPG-KEY-* /etc/pki/rpm-gpg/
 COPY . /srv/tree/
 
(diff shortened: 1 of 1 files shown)


ashcrow:
Failure is due to a missing ``cri-o`` dependency and is a known issue.

```
0. nothing provides containernetworking-plugins >= 0.5.2-7 needed by cri-o-1.11-72cb4483f7564cb057d9e22174090ec222071c37.ce0b9dc836a9b3a878f56f7480d802c391103c9c.el7.centos.x86_64
```

ashcrow:
Hopefully https://github.com/openshift/os/pull/32 will fix the build issue as it should provided the needed dep for ``cri-o``.

PR #32, opened 2018-05-16
Build cni plugins

ashcrow:
Add ``containernetworking-cni`` which provides ``containernetworking-pluigins`` for ``cri-o``.

### DIFF
--- a/overlay.yml
+++ b/overlay.yml
@@ -27,3 +27,8 @@ components:
         name: ignition
         patches: drop
 
+  # Required as a cri-o dependency
+  - src: https://github.com/containernetworking/plugins.git
+    distgit:
+        name: containernetworking-cni
+        patches: drop


host-base.json
```diff
@@ -114,6 +114,7 @@
         "origin-node",
         "origin-hyperkube",
         "origin-clients",
+        "containernetworking-cni",
```
cgwalters: Since this is an RPM dependency today, do you see value in explicitly listing it?

ashcrow: Hrm, I guess not. My half awake brain added it because ``containernetworking-cni`` provides ``containernetworking-plugins``, but it should solve without explicit addition like this. I'll drop it from this PR.

ashcrow: Done

ashcrow:
``
0. nothing provides containernetworking-plugins >= 0.5.2-7 needed by cri-o-1.11-72cb4483f7564cb057d9e22174090ec222071c37.92e65ab06f5d2eee9cf6dcc76f64f696c2cb3e89.el7.centos.x86_64
``

I *think* this is because the ``rpmdistro-overlay`` builds and the test are separated. @jlebon can you confirm?

cgwalters:
Yeah, it's a circular dependency.

/merge

cgwalters:
Basically true multi-component cross-repository CI/CD is hard; we have this same setup today with PAPR+rdgo for ostree/rpm-ostree.  Building multiple dependent packages in Koji sucks.

I think for now let's neuter the prow check?

ashcrow:
@cgwalters SGTM

ashcrow:
Manually merged.

PR #34, opened 2018-05-16
Add redhat-release-coreos

cgwalters:
We want to have our own release for many reasons.  This also
fixes the issue with having `ostree-remount.service` enabled,
so drop our treefile change for it.

### DIFF
--- a/overlay.yml
+++ b/overlay.yml
@@ -32,3 +32,6 @@ components:
     distgit:
         name: containernetworking-cni
         patches: drop
+
+  - src: github:openshift/redhat-release-coreos
+    spec: internal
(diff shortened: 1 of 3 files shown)


overlay.yml
```diff
@@ -32,3 +32,6 @@ components:
     distgit:
         name: containernetworking-cni
         patches: drop
+
+  - src: github:cgwalters/redhat-release-coreos
```
cgwalters: We should obviously figure out hosting for this... @smarterclayton WDYT about `github.com/openshift/redhat-release-coreos` ?

ashcrow: WFM

smarterclayton: WFM - transfer it to me and i'll set up permissions

cgwalters:
/hold cancel

Looks like the repo was imported, updated :arrow_up:

PR #38, opened 2018-05-16
treecompose: Write commitid to workspace too

cgwalters:
Same as the previous fix.  Not sure why this was working in my
test pipeline.

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -48,7 +48,7 @@ node(env.NODE) {
         }
 
         stage("Compose Tree") {
-            sh "rpm-ostree compose tree --repo=${repo} ${manifest} --write-commitid-to commit.txt"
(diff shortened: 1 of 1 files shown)


Jenkinsfile.treecompose
```diff
@@ -48,7 +48,7 @@ node(env.NODE) {
         }
 
         stage("Compose Tree") {
-            sh "rpm-ostree compose tree --repo=${repo} ${manifest} --write-commitid-to commit.txt"
+            sh "rpm-ostree compose tree --repo=${repo} ${manifest} --write-commitid-to $WORKSPACE/commit.txt"
             currentBuild.description = '🆕 commit ' + readFile('commit.txt');
```
miabbott: Should the `readFile()` also specify `$WORKSPACE`?

jlebon: We should find out in a few minutes :)

PR #39, opened 2018-05-16
Default to rhcos, move host-origin to contrib/

cgwalters:
This will be the focus, let's ensure the `Makefile` defaults
to it.  And since we don't have a lot of bandwidth to maintain
multiple things here, let's move the origin one to contrib/ for now.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -1,14 +1,7 @@
 // this var conveniently refers to a location on the server as well as the local dir we sync to/from
 def output = "${env.ARTIFACT_SERVER_DIR}/images"
 
-def ref
-if (env.BUILD_TYPE == 'origin') {
-    ref = "origin/3.10/x86_64/os"
-} else if (env.BUILD_TYPE == 'rhcos') {
-    ref = "openshift/3.10/x86_64/os"
-} else {
-    assert false
-}
+def ref = "openshift/3.10/x86_64/os";
 
 node(env.NODE) {
     checkout scm
--- a/.prow.sh
+++ b/.prow.sh
@@ -3,4 +3,5 @@ set -xeuo pipefail
 
 yum -y install jq
 make syntax-check
-make container
+# And disable the container again until we figure out about Prow -> internal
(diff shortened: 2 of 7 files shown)


Jenkinsfile.cloud
```diff
@@ -1,14 +1,7 @@
...
 
-def ref
-if (env.BUILD_TYPE == 'origin') {
-    ref = "origin/3.10/x86_64/os"
-} else if (env.BUILD_TYPE == 'rhcos') {
-    ref = "openshift/3.10/x86_64/os"
-} else {
-    assert false
-}
+def ref = "openshift/3.10/x86_64/os";
```
jlebon: There's some similar logic in the other Jenkinsfiles we may want to delete as well.

cgwalters: I didn't see anything in the other `Jenkinsfiles`?  But I did find other things to fix like the `README.md`, done below.

jlebon:
```
error: cannot update repo 'rhel-7.5-server': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Curl error (6): Couldn't resolve host name for http://cdn.stage.redhat.com/content/dist/rhel/server/7/7.5/x86_64/os/repodata/repomd.xml [Could not resolve host: cdn.stage.redhat.com]
make: *** [Makefile:33: rpmostree-compose] Error 1
```

Hmm OK, maybe let's just comment out the imagebuilder step for now? We'll need to wire this up a different way.

Jenkinsfile.cloud
jlebon: I meant for the treecompose one, e.g. something like:

```groovy
-if (env.BUILD_TYPE != 'origin' && env.BUILD_TYPE != 'rhcos') {
-    assert false
-}
-def manifest = "host-${env.BUILD_TYPE}.json"
+def manifest = "host-rhcos.json"
```

?

cgwalters:
I disabled the imagebuilder for now :arrow_up:

PR #40, opened 2018-05-16
treecompose: Add db-diff

cgwalters:
I found this useful for FAHC.  Also...Groovy quoting vs shell
quoting is a huge mess.  I found
http://mrhaki.blogspot.com/2009/08/groovy-goodness-string-strings-strings.html
https://gist.github.com/Faheetah/e11bd0315c34ed32e681616e41279ef4
etc. very useful.

We should probably change our pipelines to use `chdir` and avoid
using source directories as strings.  Or maybe inject them
into the real environment?  Basically it feels saner to me if
we always use `sh('command')` and not `sh("command")`.

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -5,6 +5,7 @@ def DOCKER_ARGS = "--net=host -v /srv:/srv --privileged"
 def repo = "${env.ARTIFACT_SERVER_DIR}/repo"
 
 def manifest = "host-rhcos.json"
+def ref = "openshift/3.10/x86_64/os";
 
 node(env.NODE) {
     checkout scm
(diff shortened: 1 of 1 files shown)


Jenkinsfile.treecompose
```diff
@@ -50,6 +50,7 @@ node(env.NODE) {
         stage("Compose Tree") {
             sh "rpm-ostree compose tree --repo=${repo} ${manifest} --write-commitid-to $WORKSPACE/commit.txt"
             currentBuild.description = '🆕 commit ' + readFile('commit.txt');
+            sh 'rev=$(cat commit.txt) && rpm-ostree db diff ${rev}^ ${rev}'
```
miabbott: Would `$WORKSPACE/commit.txt` be better here, since we've been bitten in the past when we didn't use that?

cgwalters:
/hold

Actually I broke things with `--write-commit-id` since we stopped setting the ref.  Going to fix that with tihs.

Jenkinsfile.treecompose
cgwalters: I think the issues with needing `$WORKSPACE` elsewhere is because of `cd`; in the rdgo code we do:
`            sh "cd ${rdgo} && rpmdistro-gitoverlay build --touch-if-changed $WORKSPACE/rdgo.stamp --logdir=log"`

But we're not doing that here.

jlebon:
/lgtm

> I think the issues with needing `$WORKSPACE` elsewhere is because of `cd`

Oh shoot, mystery solved!

PR #44, opened 2018-05-18
cloud: Generate AMIs

jlebon:
Start generating AMIs from the qcow2. To do this, we first start by also
converting the imagefactory output to an optimized vmdk (AWS doesn't
accept qcow2s and raw would just be too wasteful). Then we upload that
image using `ore`.

I had initially written a fully functional Python script to do this,
thinking after looking at both `ore` and `fedimg` that neither will do.
The knowledge I gained from the exercise made me realize `ore` indeed
does what we need.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -1,3 +1,6 @@
+def DOCKER_IMG = "registry.fedoraproject.org/fedora:28"
+def DOCKER_ARGS = "--net=host -v /srv:/srv --privileged"
+
 // this var conveniently refers to a location on the server as well as the local dir we sync to/from
 def server_dir = "${env.ARTIFACT_SERVER_DIR}"
 def output = "${server_dir}/images"
@@ -64,10 +67,12 @@ node(env.NODE) {
         def dirname = (new Date()).format("YYYY-MM-dd-HH-mm-ss-") + commit
         dirpath = "${output}/cloud/${dirname}"
         qcow = "${dirpath}/rhcos.qcow2"
+        vmdk = "${dirpath}/rhcos.vmdk"
         sh "mkdir -p ${dirpath}"
 
         sh "qemu-img convert -f raw -O qcow2 ${image} ${qcow}"
         sh "gzip ${qcow}"
+        sh "qemu-img convert -f raw -O vmdk ${image} -o adapter_type=lsilogic,subformat=streamOptimized,compat6 ${vmdk}"
 
         sh "ln -sfn ${dirname} ${output}/cloud/latest"
         // just keep the last 2 (+ latest symlink)
@@ -87,6 +92,7 @@ node(env.NODE) {
                 ${prev_commit} ${commit} > ${dirpath}/${commit}.pkgdiff
         """
         sh "sha256sum ${qcow}.gz | awk '{print \$1}' > ${qcow}.gz.sha256sum"
+        sh "sha256sum ${vmdk} | awk '{print \$1}' > ${vmdk}.sha256sum"
     }
 
     stage("Sync Out") {
@@ -99,6 +105,30 @@ node(env.NODE) {
                             -o StrictHostKeyChecking=no' \
                     ${output}/ ${env.ARTIFACT_SERVER}:${output}
             """
+        }
+    }
+
+    stage("Create AMI") {
+        docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
+            // not packaged yet, just build from src for now
+            sh "dnf install -y git golang"
+            sh "git clone https://github.com/coreos/mantle"
+            sh "mantle/build ore"
+
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                              credentialsId: env['AWS_CREDENTIALS']]]) {
+                sh """
+                    # use a symlink so that our uploaded filename is unique
+                    ln -sf ${output}/cloud/latest/rhcos.vmdk rhcos-${commit}.vmdk
+                    mantle/bin/ore aws upload --region us-east-1 \
+                        --ami-name 'rhcos_dev_${commit[0..6]}' \
+                        --ami-description 'Red Hat CoreOS ${commit}' \
+                        --bucket 's3://${env.S3_PRIVATE_BUCKET}/rhcos/cloud' \
+                        --file rhcos-${commit}.vmdk \
+                        --name "rhcos_dev_${commit[0..6]}" \
+                        --delete-object # already default, just being explicit
+                """
+            }
         }
     }
 }
--- a/rhcos.tdl
+++ b/rhcos.tdl
@@ -9,5 +9,8 @@
         </install>
         <rootpw>none</rootpw>
     </os>
+    <disk>
+        <size>6G</size>
+    </disk>
 </template>
 


Jenkinsfile.cloud
```diff
@@ -69,4 +73,26 @@ node(env.NODE) {
...
+            sh "dnf install -y git golang"
+            sh "git clone https://github.com/coreos/mantle"
+            sh "mantle/build ore"
+
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                              credentialsId: env['AWS_CREDENTIALS']]]) {
+                sh """
+                    mantle/bin/ore aws upload --region us-east-1 \
+                        --ami-name 'rhcos_dev_${commit[0..6]}' \
+                        --ami-description 'Red Hat CoreOS ${commit}' \
```
jlebon: Clearly, a better label here is the version number. We can do that as a follow-up step. There's a lot of metadata missing in general to help with debugging, discoverability, etc...

Jenkinsfile.cloud
```diff
@@ -69,4 +73,26 @@ node(env.NODE) {
...
+
+    stage("Create AMI") {
+        docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
+            // not packaged yet, just build from src for now
+            sh "dnf install -y git golang"
+            sh "git clone https://github.com/coreos/mantle"
+            sh "mantle/build ore"
+
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                              credentialsId: env['AWS_CREDENTIALS']]]) {
```
peterbaouoft: So, does this Credentials link to a AWS EC2 VM? Where does it get consumed? Is jenkins able to link the credential ID to a EC2 VM or...? /me asking with almost no context tho :p.

ashcrow: @peterbaouoft Jenkins provides said ``AWS_CREDENTIALS`` environment variable. The value is stored in jenkins itself. Using it allows further AWS commands to be used with said credentials.

Jenkinsfile.cloud
```diff
@@ -69,4 +73,26 @@ node(env.NODE) {
...
+                sh """
+                    mantle/bin/ore aws upload --region us-east-1 \
+                        --ami-name 'rhcos_dev_${commit[0..6]}' \
+                        --ami-description 'Red Hat CoreOS ${commit}' \
+                        --bucket 's3://${env.S3_PRIVATE_BUCKET}/rhcos' \
+                        --delete-object --force \
```
bgilbert: `--delete-object` is the default and can be omitted.

`--force` is slightly misdocumented.  Without it, the existing S3 object will be reused (and then deleted after the snapshot is created).  With it, the existing object is clobbered, which could race with another build using the same object.  Preferably there's a 1:1 mapping between image bits and filename (so reusing an object from a previously aborted run is safe) and force is not set.

Along those lines, there's a bug in ore and a bug here.  The bug in ore is that `--bucket` is supposed to be a prefix with the basename of `--file` appended, but instead we're using `--bucket` as the complete object name.  The bug here is that `--bucket + basename(--file)` is always the same value, so if there are two simultaneous runs, one will clobber the other.

Jenkinsfile.cloud
peterbaouoft: Ah, I think I get it now. I was unsure what s3 was b4, that is why I kinda have the confusion :p. It seems like the credentials was used for access for s3(hopefully). Thanks for the clarification =).

Jenkinsfile.cloud
bgilbert: ore bugfix at https://github.com/coreos/mantle/pull/854.

jlebon: Thanks for the review! I dropped `--force` and based the filename on the commit. Also added a tweak to the TDL to lower the virtual disk size.

jlebon: > `--delete-object` is the default and can be omitted.

Yeah, I pass it just to be more explicit. I added a comment for that.

Jenkinsfile.cloud
```diff
@@ -69,4 +73,28 @@ node(env.NODE) {
...
+            sh "git clone https://github.com/coreos/mantle"
+            sh "mantle/build ore"
+
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding',
+                              credentialsId: env['AWS_CREDENTIALS']]]) {
+                sh """
+                    # use a symlink so that our uploaded filename is unique
+                    ln -sf ${output}/cloud/latest/rhcos.vmdk rhcos-${commit}.vmdk
+                    mantle/bin/ore aws upload --region us-east-1 \
+                        --ami-name 'rhcos_dev_${commit[0..6]}' \
```
dobbymoodge: The pruner looks at the **Tag** `Name`. Can `ore` take an argument to set the tag as well?

ashcrow: I took a quick look and it doesn't look like it uses the tag today:

- https://github.com/coreos/mantle/blob/master/cmd/ore/aws/aws.go#L50-L55
- https://github.com/coreos/mantle/blob/master/cli/cli.go#L55-L60

bgilbert: Currently, mantle's tagging is pretty hardcoded, but we can add functionality as needed.  How does the pruner use the `Name` tag?

ashcrow: @bgilbert can you verify?

ashcrow: @dobbymoodge :arrow_up:

ashcrow:
Adding in @imcleod as well in case he hasn't seen this.

ashcrow:
It looks the following needs to be done:

- [x] Make changes requested by initial review 
- [ ] Find out how to name/tag to work with the pruner
- [ ] Rebase on top of master

Jenkinsfile.cloud
dobbymoodge: @bgilbert It matches against the Name tag as a first step in classifying the AMI. Very roughly, something like `AMI_WILDCARDS["rhcos_dev_*"] = {:keep => 5}` means to deregister AMIs prefixed with "`rhcos_dev_`", except the 5 most recent.

ashcrow: That sounds like it would work to me.

jlebon:
Rebased!

ashcrow:
https://github.com/coreos/mantle/pull/856

Jenkinsfile.cloud
dobbymoodge: @jlebon Ah yeah, you're right. That was an extended brain fart I had there, sorry. Regardless, the more sophisticated pruning logic **does** key off of the tags, and that's the direction we have planned to go in for future changes, so that mantle PR makes sense. Thanks, and sorry about the confusion.

peterbaouoft:
>so if everyone is happy, then let's merge this in?

I wants to :+1: :)

Jenkinsfile.cloud
```diff
@@ -101,4 +107,28 @@ node(env.NODE) {
...
+                        --ami-description 'Red Hat CoreOS ${commit}' \
+                        --bucket 's3://${env.S3_PRIVATE_BUCKET}/rhcos/cloud' \
+                        --file rhcos-${commit}.vmdk \
+                        --name "rhcos_dev_${commit[0..6]}" \
```
cgwalters: We don't need to change this right now, but I think down the line we should prefer version numbers over hashes, particularly truncated hashes.

jlebon: Right now, we're only publishing this metadata internally. We should also sync it to an S3 bucket or something?

PR #46, opened 2018-05-21
cloud: fix sha256sum generation

miabbott:
When the 'metadata' stage went to production, it blew up on the
generating `sha256sum` step with an error like this:

```
java.lang.NoSuchMethodError: No such DSL method 'sha256sum
/srv/rhcos/output/images/cloud/...
```

Since it worked on my test Jenkins, I'm just taking a guess that the
use of double-quotes around the `awk` command confused the DSL parser.

So this might fix it, but it might continue to fail...

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -86,7 +86,7 @@ node(env.NODE) {
             rpm-ostree db diff --repo=${repo} \
                 ${prev_commit} ${commit} > ${dirpath}/${commit}.pkgdiff
         """
-        sh "sha256sum ${qcow}.gz | awk "{print $1}" > ${qcow}.gz.sha256sum"
+        sh "sha256sum ${qcow}.gz | awk '{print $1}' > ${qcow}.gz.sha256sum"
     }
 
     stage("Sync Out") {


ashcrow:
Understood

Jenkinsfile.cloud
```diff
@@ -86,7 +86,7 @@ node(env.NODE) {
             rpm-ostree db diff --repo=${repo} \
                 ${prev_commit} ${commit} > ${dirpath}/${commit}.pkgdiff
         """
-        sh "sha256sum ${qcow}.gz | awk "{print $1}" > ${qcow}.gz.sha256sum"
+        sh "sha256sum ${qcow}.gz | awk '{print $1}' > ${qcow}.gz.sha256sum"
```
cgwalters: I spent a while reading http://groovy-lang.org/syntax.html#_string_interpolation while working on this project.   Anyways, this looks a lot more right, except based on my testing, we need `\$1` instead of `$1` to tell Groovy not to look for a variable named `1`.  The single quotes ensure `/bin/sh` won't try to look it up, because we want the *third* language involved here (awk) to find the `1` variable.  Phew!

miabbott: groooooooooovy

PR #48, opened 2018-05-21
cloud: remove sha256sum generation

miabbott:
I don't want to block any pipeline progress, so let's remove the
`sha256sum` stuff for now.

I'll figure out a better way to do this in a separate PR.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -86,7 +86,6 @@ node(env.NODE) {
             rpm-ostree db diff --repo=${repo} \
                 ${prev_commit} ${commit} > ${dirpath}/${commit}.pkgdiff
         """
-        sh "sha256sum ${qcow}.gz | awk '{print $1}' > ${qcow}.gz.sha256sum"
     }
 
     stage("Sync Out") {


miabbott:
FWIW, #46 caused the pipeline to break in a new and exciting way:

```
org.codehaus.groovy.control.MultipleCompilationErrorsException: startup failed:
WorkflowScript: 89: illegal string body character after dollar sign;
   solution: either escape a literal dollar sign "\$5" or bracket the value expression "${5}" @ line 89, column 29.
           sh "sha256sum ${qcow}.gz | awk '{print $1}' > ${qcow}.gz.sha256sum"
                               ^
```

cgwalters:
See https://github.com/openshift/os/pull/46#discussion_r189708068

ashcrow:
@miabbott looks like @cgwalters has a proposed way to fix. Should we hold on this so you can implement his suggestion?

miabbott:
/hold

Will push the fix suggested by @cgwalters

miabbott:
Have a look at #49 

@ashcrow @cgwalters

PR #50, opened 2018-05-22
cloud: Add more tags to AMI and snapshot

jlebon:
Follow up to #44 and #45. See individual commit messages for details.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -39,6 +39,23 @@ node(env.NODE) {
         }
     }
 
+    // Check if there's a new version out.
+    // Really, we should be checksumming e.g. the ks and tdl too.
+    def latest_version, last_build_version
+    stage("Check for Changes") {
+        latest_version = sh(returnStdout: true,
+                            script: "ostree show --repo=${repo} --print-metadata-key=version ${ref}").trim()
+        latest_version = latest_version[1..-2] // trim single quotes
+        last_build_version = sh(returnStdout: true,
+                                script: "cat ${output}/cloud/version.txt || :").trim()
+    }
+
+    if (latest_version == last_build_version) {
+        currentBuild.result = 'SUCCESS'
+        currentBuild.description = '(No changes)'
+        return
+    }
+
     stage("Prepare Configs") {
         sh "sed -i 's,\\(<url>\\).*\\(<\\/url\\),\\1${env.INSTALLER_TREE_URL}\\2,' rhcos.tdl"
         sh "sed -i 's,@@OSTREE_INSTALL_URL@@,${env.OSTREE_INSTALL_URL},' cloud.ks"
@@ -74,6 +91,9 @@ node(env.NODE) {
         qcow = "${dirpath}/rhcos.qcow2"
         vmdk = "${dirpath}/rhcos.vmdk"
         sh "mkdir -p ${dirpath}"
+        // this belongs better in a JSON file, but for now just use a file;
+        // this is used higher up to determine no-op changes
+        sh "echo '${version}' > ${dirpath}/version.txt"
 
         sh "qemu-img convert -f raw -O qcow2 ${image} ${qcow}"
         sh "gzip ${qcow}"
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -85,11 +85,11 @@ node(env.NODE) {
 
         sh """
             rpm-ostree db list --repo=${repo} ${commit} > \
-                /${dirpath}/${commit}.rpmlist
+                /${dirpath}/pkglist.txt
         """
         sh """
             rpm-ostree db diff --repo=${repo} \
-                ${prev_commit} ${commit} > ${dirpath}/${commit}.pkgdiff
+                ${prev_commit} ${commit} > ${dirpath}/pkgdiff.txt
         """
         sh "sha256sum ${qcow}.gz | awk '{print \$1}' > ${qcow}.gz.sha256sum"
         sh "sha256sum ${vmdk} | awk '{print \$1}' > ${vmdk}.sha256sum"
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -116,7 +116,7 @@ node(env.NODE) {
     stage("Create AMI") {
         docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
             // not packaged yet, just build from src for now
-            sh "dnf install -y git golang"
+            sh "dnf install -y git golang awscli jq"
             sh "git clone https://github.com/coreos/mantle"
             sh "mantle/build ore"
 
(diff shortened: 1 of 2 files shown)


Jenkinsfile.cloud
```diff
@@ -55,6 +55,21 @@ node(env.NODE) {
...
+    stage("Resync Repo") {
+        withCredentials([sshUserPrivateKey(credentialsId: env['ARTIFACT_SSH_CREDS_ID'],
+                                           keyFileVariable: 'KEY_FILE')]) {
+            sh """
+                mkdir -p ${output}/repo
+                rsync -Hrlpt --stats \
+                    -e 'ssh -i ${env.KEY_FILE} \
+                            -o UserKnownHostsFile=/dev/null \
+                            -o StrictHostKeyChecking=no' \
+                    ${env.ARTIFACT_SERVER}:${server_dir}/repo ${server_dir}
```
cgwalters: I'm confused...aren't we pointing anaconda at the remote ostree-repo?  Why do we need to sync it locally?  What's `${output}/repo` versus `${server_dir}`?

Jenkinsfile.cloud
```diff
@@ -70,16 +70,21 @@ node(env.NODE) {
...
+    def commit, version, dirpath, qcow, vmdk
     stage("Finalizing") {
         def image = sh(returnStdout: true, script: "ls /var/lib/imagefactory/storage/*.body").trim()
         // just introspect after the fact to avoid race conditions
         commit = sh(returnStdout: true,
                     script: """LIBGUESTFS_BACKEND=direct virt-cat -a ${image} -m /dev/coreos/root:/ \
                                     /ostree/repo/refs/remotes/rhcos/${ref}""").trim()
-        currentBuild.description = 'commit ' + commit
+        version = sh(returnStdout: true,
+                     script: "ostree show --repo=${repo} --print-metadata-key=version ${commit}").trim()
```
cgwalters: Ah wow, that's some heroics!  I was going to reply that it was racy, but I see what you're doing now.

Jenkinsfile.cloud
```diff
@@ -142,11 +142,20 @@ node(env.NODE) {
...
+                        --delete-object | tee out.json # already default, just being explicit
+
+                    # Add the version and commit as tags to both the AMI and the underlying snapshot.
+                    # We should teach mantle to apply extra tags.
+                    ami=\$(jq -r .HVM out.json)
+                    snapshot=\$(jq -r .SnapshotID out.json)
+                    AWS_DEFAULT_REGION=us-east-1 aws ec2 create-tags \
+                        --resources \${ami} \${snapshot} \
+                        --tags Key=ostree_commit,Value=${commit} \
+                               Key=ostree_version,Value=${version}
```
cgwalters: Neato.

Jenkinsfile.cloud
```diff
@@ -39,6 +39,23 @@ node(env.NODE) {
...
+    def latest_version, last_build_version
+    stage("Check for Changes") {
+        latest_version = sh(returnStdout: true,
+                            script: "ostree show --repo=${repo} --print-metadata-key=version ${ref}").trim()
+        latest_version = latest_version[1..-2] // trim single quotes
+        last_build_version = sh(returnStdout: true,
+                                script: "readlink ${output}/cloud/latest").trim()
+    }
+
+    if (latest_version == last_build_version) {
```
cgwalters: What I've done elsewhere is have a "blow the cache" key in metadata.  If Pagure was up, I'd link to what https://pagure.io/fedora-atomic-host-continuous is doing, but basically: Let's have a yaml file like:

`build-caching.yml`:
```yaml
cloud-force-rebuild: 3.10-7
````

Then we add a condition here like `if (latest_version != cloud_force_rebuild && latest_version == last_build_version` or so, where we parse `cloud_force_rebuild` from the yaml in the job.

File: Jenkinsfile.cloud (hunk 1 above)
Comment (cgwalters) [reply Jenkinsfile.cloud]: Now I understand this, although I think it'd be a lot less expensive to just do an `ostree pull --commit-metdata-only` or so later.

File: Jenkinsfile.cloud [hunk 5]
```diff
@@ -55,20 +81,32 @@ node(env.NODE) {
...
+        // fetch parent too for the later db diff
+        sh "ostree --repo=repo pull --depth=1 --commit-metadata-only tmp ${commit}"
+        version = sh(returnStdout: true,
+                     script: "ostree show --repo=repo --print-metadata-key=version ${commit}").trim()
+        assert (version.startsWith("'") && version.endsWith("'"))
+        version = version[1..-2] // trim single quotes
+
+        currentBuild.description = "${version} (${commit})"
 
         def dirname = (new Date()).format("YYYY-MM-dd-HH-mm-ss-") + commit
```
jlebon: One tricky bit here is that I left the dirname alone instead of using the version string as before. The reason is that if we force-nocache to rebuild the same version, I still wanted it to be output in a new directory instead of overwriting.

Jenkinsfile.cloud
```diff
@@ -55,20 +81,32 @@ node(env.NODE) {
...
         dirpath = "${output}/cloud/${dirname}"
         qcow = "${dirpath}/rhcos.qcow2"
         vmdk = "${dirpath}/rhcos.vmdk"
         sh "mkdir -p ${dirpath}"
+        // this belongs better in a JSON file, but for now just use a file;
+        // this is used higher up to determine no-op changes
+        sh "echo '${version}' > ${dirpath}/version.txt"
```
jlebon: I didn't want to make the PR even longer, but in a future patch we can just output a mini JSON file with all the metadata in there.

Jenkinsfile.cloud
```diff
@@ -105,19 +115,26 @@ node(env.NODE) {
     }
 
     stage("Generate Metadata") {
-        def prev_commit = sh(returnStdout: true,
-                             script: "ostree --repo=${repo} rev-parse ${commit}^").trim()
+        // we're stuck with an older host because imagefactory produces kernel
+        // panics on newer hosts, so let's use a container to have access to a
```
cgwalters: Nonblocking: Bug/reference?

Jenkinsfile.cloud
```diff
@@ -105,19 +115,26 @@ node(env.NODE) {
...
+        // newer rpm-ostree that understands the new pkglist metadata
+        docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
+            sh "dnf install -y rpm-ostree ostree"
```
cgwalters: Nonblocking: After the composer container change lands, this could use that container.

Jenkinsfile.cloud
```diff
@@ -80,19 +118,26 @@ node(env.NODE) {
...
```
miabbott: Maybe we should think about publishing a tools container that contains all the goodies we need:  `ostree`, `rpm-ostree`, `git`,`golang`, `aws-cli`, `coreos-assembler` etc etc

PR #51, opened 2018-05-23
rdgo: Force cri-o to 1.10.x

ashcrow:
cri-o needs to be pinned to the OpenShift release. (1.10 to 3.10).
The spec in fedora distgit's master branch doesn't work with 3.10.
Because of this we need to switch to the fedora distgit branch for
f28 which builds properly for cri-o and still allows proper building of
the other packages.

/cc @derekwaynecarr


ashcrow:
Flake maybe? Says the job is still running and Jenkins can't find it ...

/retest

PR #52, opened 2018-05-23
Convert to YAML definition, use coreos-assembler

cgwalters:
This uses https://github.com/cgwalters/coreos-assembler
which today mostly converts YAML treefiles into JSON.  It also helps
avoid provisioning time.


host.yaml
```diff
@@ -0,0 +1,89 @@
...
+ - setools-console
+ # System setup
+ - ignition cloud-init cloud-utils-growpart
+ - dracut-network
+ # Bootloader
+ - grub2 grub2-efi ostree-grub2 efibootmgr shim
+ # SSH
+ - openssh-server openssh-clients
+ # Containers
+ - docker podman skopeo runc bubblewrap cri-o
```
cgwalters: This apparently also snuck in a change to add `podman` here.

cgwalters:
One downside here is that I haven't figured out if it's possible to teach the Jenkins docker pipeline bits to do the Kubernetes equivalent of `imagePullPolicy: Always`.

host.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+remove-from-packages: 
+  - - yum
+    - "/usr/bin/.*"
+  - - filesystem
+    - "/usr/share/backgrounds"
+units:
+  - "docker.service"
+  - "tuned.service"
+  - "ostree-remount.service"
```
jlebon: Minor: we shouldn't need this anymore right?

host.yaml
```diff
@@ -0,0 +1,89 @@
...
+ - authconfig sssd shadow-utils kexec-tools
+ # tuned
+ - tuned tuned-profiles-atomic
+ # Used by admins interactively
+ - sudo coreutils less tar xz gzip bzip2 rsync tmux
+ - nmap-ncat net-tools bind-utils
+ # Editors
+ - vim-minimal nano
+ # OpenShift
+ - origin-node origin-hyperkube origin-clients
```
jlebon: This list is missing `redhat-release-coreos`, no?

miabbott:
> One downside here is that I haven't figured out if it's possible to teach the Jenkins docker pipeline bits to do the Kubernetes equivalent of imagePullPolicy: Always.

I think it is a two stage operation:

```
docker.image(DOCKER_IMG).pull()
docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
```

cgwalters:
> docker.image(DOCKER_IMG).pull()

Yep, tested that, works.  Updated, thanks!

host.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - root
+ignore-removed-groups:
+  - root
+check-passwd:
+  type: "file"
+  filename: "passwd"
+check-groups:
+  type: "file"
+  filename: "group"
+packages:
```
jlebon: I really like the package breakdown like this!

cgwalters: One other thing I did in `coreos-assembler` is that not only is it YAML, each line in the `packages` section can be space-separated too.  It feels slightly dirty but since package names can't contain whitespace...

host.yaml
cgwalters: Yes, good catch!  For reference here's a diff vs the base packages:

```
--- packages-old.txt	2018-05-23 15:32:18.246115057 -0400
+++ packages-new.txt	2018-05-23 15:32:43.910079944 -0400
@@ -1,8 +1,8 @@
 authconfig
-bash-completion
 bind-utils
 biosdevname
 bridge-utils
+bubblewrap
 bzip2
 ceph-common
 chrony
@@ -19,8 +19,6 @@
 dracut-network
 e2fsprogs
 efibootmgr
-filesystem
-glibc
 glusterfs-fuse
 grub2
 grub2-efi
@@ -32,7 +30,6 @@
 kernel
 kexec-tools
 less
-logrotate
 lvm2
 mdadm
 microcode_ctl
@@ -49,7 +46,7 @@
 origin-hyperkube
 origin-node
 ostree-grub2
-passwd
+podman
 policycoreutils-python
 rpm-ostree
 rsync
@@ -65,7 +62,6 @@
 subscription-manager-plugin-container
 subscription-manager-plugin-ostree
 sudo
-systemd
 tar
 tmux
 tuned
```

I squashed in most of the changes, although I dropped explicitly listing `filesystem glibc` because we really only want them via deps.

host.yaml
```diff
@@ -0,0 +1,92 @@
...
+  - rhel-7.5-server-extras
+  - rhel-7.5-atomic
+install_langs:
+ - en_US
+documentation: false
+initramfs-args:
+  - "--no-hostonly"
+  - "--add"
+  - "iscsi"
+automatic_version_prefix: "3.10-7.1803"
```
jlebon: Minor: let's stick with `3.10-7.5.0`? (I'm guessing this was just forgotten during the transition from inheriting JSON files to a single YAML).

host.yaml
jlebon: That diff looks good to me!

host.yaml
cgwalters: I changed to use `7.5` since...the `.0` there feels like it's claiming something stronger than we're really saying.

jlebon:
Just needs a rebase!

cgwalters:
Rebased :surfing_man:

PR #56, opened 2018-05-23
rdgo: Fix branch for ignition

cgwalters:
Followup fix to https://github.com/openshift/os/pull/51
The problem was changing the dist-git branch globally to `f28`.


cgwalters:
Yeah, I think you'd have had to do `resolve --fetch-all` to notice this.

PR #64, opened 2018-05-25
Add pipeline-utils.groovy local lib

jlebon:
Add a separate groovy file and factor out a few useful functions common
to multiple pipelines.

This is more of an `#include` type thing rather than a Jenkins pipeline
shared library[1]. It doesn't feel like it's quite worth it yet.

[1] https://jenkins.io/doc/book/pipeline/shared-libraries/

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -1,15 +1,17 @@
 def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler"
 def DOCKER_ARGS = "--net=host -v /srv:/srv --privileged"
 
-// this var conveniently refers to a location on the server as well as the local dir we sync to/from
-def server_dir = "${env.ARTIFACT_SERVER_DIR}"
-def output = "${server_dir}/images"
+// this var conveniently refers to a location on the server as well as the
+// local dir we sync to/from
+def output = "${env.ARTIFACT_SERVER_DIR}/images"
 
 def ref = "openshift/3.10/x86_64/os";
 
 node(env.NODE) {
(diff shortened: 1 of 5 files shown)


cgwalters:
Is there a way we can use this to share common properties among the different files using the `properties` definition?

smarterclayton:
/test all

(just verifying

jlebon:
> Is there a way we can use this to share common properties among the different files using the `properties` definition?

Did this in the second commit above by shoving all the internal goop into credentials! :arrow_up:
I only tested this for the cloud pipeline, though it should work OK for the other ones as well (and we can just follow up with fixups if not).

cgwalters:
Hm, using credentials feels weird; might be simpler to have one credential with all of the config as a newline-separated `FOO=BAR\nBAZ=BLAH\n` etc and then we have a helper and parse it with [withEnv](https://jenkins.io/doc/pipeline/steps/workflow-basic-steps/#code-withenv-code-set-environment-variables)  or something? 

But eh, not worth reworking, so I'm OK to merge this as is.

ashcrow:
I'll defer to @cgwalters for merge.

jlebon:
> Hm, using credentials feels weird; might be simpler to have one credential with all of the config

Or maybe we can make the credential a URL to a groovy script that does `env.FOO=BAR`?

Hmm, I'm a bit torn between that, and the clarity that keeping them separate allows us. I do agree using credentials is a bit odd, though OTOH, being able to have a description for each one makes it easier to understand what they are and where they come from.

Anyway, we can easily revisit this later!

PR #65, opened 2018-05-25
Dockerfile,Makefile: Use coreos-assembler

cgwalters:
Fixes the (we need a name for this I'll just make one up) "all-in-one-Dockerfile"
path again, which I broke when doing the -assembler switch to use yaml.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
 # Use fedora as we maintain our tools a bit better there
-FROM registry.fedoraproject.org/fedora:28 AS build
+FROM quay.io/cgwalters/coreos-assembler AS build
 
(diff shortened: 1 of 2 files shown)


smarterclayton:
https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_os/65/pull-ci-os-container-build/1000115800117350400/

cgwalters:
Hm...is something in the prow job overriding the repos now?  I submitted https://github.com/openshift/os/pull/66

smarterclayton:
That passed

smarterclayton:
Yes, the prow job is now using the mirrors.  I'm going to do it in the job because i don't want the repo to be public.  See openshift/release#884

PR #66, opened 2018-05-25
build: Use OSORGCI-maintained repo mirrors

cgwalters:
Pulls from http://rpm-mirror-ci.svc.ci.openshift.org/


smarterclayton:
I'm going to do this in the prow job because i don't want to expose the secrets to the repo and i don't want to expose the mirror outside the cluster firewall for now.  See openshift/release#884 which adds the push job and the PR job

cgwalters:
Ah.  OK.

PR #67, opened 2018-05-28
cloud: Fix AMI upload regression

jlebon:
Minor regression from #63. `ore` is now in the `$PATH`, so just call it
directly.


peterbaouoft:
Makes sense, 
/lgtm

PR #69, opened 2018-05-29
oscontainer: Fix credentials passing for registry

jlebon:
The `docker.withRegistry` syntax already expects a credential ID rather
than a username/password. So skip wrapping it with `withCredentials` and
just pass the ID directly to `docker.withRegistry` instead.


jlebon:
Weird, this one seems stuck.

PR #71, opened 2018-05-31
Set timer interval directly from pipeline

jlebon:
This moves the last bit of configuration from the internal JJB to the
pipeline. Another reason for doing this now is that I've noticed the
timer set by JJB is actually getting overwritten by the `properties`
call we do here. The end result is that the timers were completely
turned off.


miabbott:
LGTM, but I wonder if some of the properties/parameters could live in the `pipeline-utils.groovy` so we don't have to change them in multiple places?

jlebon:
Can you expand on that? That's what we do already with the shared `define_properties`, right?

PR #75, opened 2018-06-04
Attempt to get ignition working in stream

dustymabe:
- disable rdgo for ignition for now
- add boot args for ignition
- target just ec2 for now

### DIFF
--- a/cloud.ks
+++ b/cloud.ks
@@ -16,7 +16,13 @@ services --disabled=network,avahi-daemon
 
 zerombr
 clearpart --initlabel --all
-bootloader --timeout=1 --append="no_timer_check console=tty1 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0"
+# Add the following to kernel boot args:
+#  - ip=dhcp           # how to get network
+#  - rd.neednet=1      # tell dracut we need network
+#  - enforcing=0       # ignition + selinux doesn't work
+#  - coreos.first_boot # tell ignition to run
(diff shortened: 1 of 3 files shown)


cloud.ks
```diff
@@ -16,7 +16,9 @@ services --disabled=network,avahi-daemon
 
 zerombr
 clearpart --initlabel --all
-bootloader --timeout=1 --append="no_timer_check console=tty1 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0"
+# Add "ip=dhcp rd.neednet=1 enforcing=0 coreos.first_boot" for ignition for now
+# Add "coreos.oem.id=ec2" for ec2 target for now
```
ashcrow: Looks like ``coreos.oem.id=ec2`` is also added here. No biggie. If you want to add it to the above line we'll wait to merge, but it's not a blocker.

ashcrow:
Failure is:
``
error: No package matches 'ignition-dracut'
``

should be fixed once the package is in @dustymabe's copr.

cloud.ks
dustymabe: not sure what you mean exactly - i broke the two comments out on purpose but maybe I should break each arg out and give a comment about each one. will do that now.

dustymabe:
> It just hit me I think a change to https://github.com/openshift/os/blob/master/host.yaml is also needed to get the repo added.

yeah for some reason I thought that was already done.

dustymabe:
addressed comments. PTAL

PR #76, opened 2018-06-04
Remove harcoding of EC2 target

dustymabe:
Did some rudimentary logic in the dracut modules for now.


dustymabe:
link to PR with logic: https://github.com/dustymabe/bootengine/pull/2

PR #81, opened 2018-06-05
Add developmentPipeline, DRY_RUN

cgwalters:
This makes it easier to set up a development instance; fork
the repo, add a commit which sets `developmentPipeline=true`, then
your work commits on top.  The build timers will be disabled, and
you won't push artifacts over the production ones.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -137,6 +137,13 @@ node(NODE) {
         }
     }
 
+   if (params.DRY_RUN) {
+       echo "DRY_RUN set, skipping push"
+       currentBuild.result = 'SUCCESS'
+       currentBuild.description = '(dry run)'
+       return
+   }
+
     stage("Sync Out") {
         withCredentials([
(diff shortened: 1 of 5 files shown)


Jenkinsfile.cloud
```diff
@@ -137,6 +137,13 @@ node(NODE) {
         }
     }
 
+   if (params.DRY_RUN) {
```
peterbaouoft: I am still catching up with new changes, so might of miss something :p.  But a small question, this section seems to be reused four times, is it an option to put this section into the groovy file?

cgwalters: Hm, maybe - a bit tricky as we need the `return` here but we could possibly make it something like:

`if (utils.dry_run()) return;` ?

ashcrow: Makes sense.

jlebon: Good idea.
This PR looks good to me otherwise!

miabbott:
/lgtm

No one seems opposed and the code looks sane, so here we go!

PR #83, opened 2018-06-06
 Jenkinsfile.cloud: Use qcow2-to-vagrant to make vagrant-libvirt box

cgwalters:
I find Vagrant useful.  Though it'd probably be a lot more useful
to have a vagrant-virtualbox version...

Doing that type of thing is probably going to require separate ImageFactory
runs.  I may investigate that at some point.

Also I snuck in here some usage of `parallel` which is fairly cool.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -12,6 +12,10 @@ def ref = "openshift/3.10/x86_64/os";
 node(NODE) {
     docker.image(DOCKER_IMG).pull()
 
+    stage("Clean workspace") {
+       step([$class: 'WsCleanup'])
+    }
+
     checkout scm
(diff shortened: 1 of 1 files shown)


Jenkinsfile.cloud
```diff
@@ -28,7 +33,12 @@ node(NODE) {
                               rpm-ostree \
                               rsync
            """
-    }
+    } }
+    par_stages["clone qcow2-to-vagrant"] = { -> stage("Clone qcow2-to-vagrant") {
+        // I tried to use git() but not sure where that went, it isn't the workspace
+        sh('git clone https://github.com/cgwalters/qcow2-to-vagrant')
+    } }
+    parallel par_stages; par_stages = [:]
```
ashcrow: Neat!

PR #88, opened 2018-06-08
disable cloud-init

dustymabe:
I don't think we want cloud-init and ignition, but I could be wrong.

### DIFF
--- a/cloud.ks
+++ b/cloud.ks
@@ -11,8 +11,8 @@ rootpw --lock --iscrypted locked
 firewall --disabled
 
 network --bootproto=dhcp --onboot=on
-services --enabled=sshd,cloud-init,cloud-init-local,cloud-config,cloud-final
-services --disabled=network,avahi-daemon
+services --enabled=sshd
+services --disabled=network,avahi-daemon,cloud-init,cloud-config,cloud-final,cloud-init-local
 
 zerombr
 clearpart --initlabel --all


cgwalters:
(If we do this there's no point shipping it in the tree either)

That said we still need either cloud-init or [coreos-metadata](https://github.com/coreos/coreos-metadata) I believe - people expect ssh key updates to work with AWS etc.

cgwalters:
>  people expect ssh key updates to work with AWS etc.

Or, maybe that was a transitional thing - maybe we say we're pure Ignition now.

ashcrow:
I personally think we should move to `ignition`.

lucab:
For context, ignition does the initial setup of ssh_keys (from userdata, on first boot) while coreos-metadata does recurring ssh_keys refreshes (from the dedicated `openssh-key` endpoint, on every boot).

:+1: on disabling cloud-init here, but there will likely be some fallout (does is still take care of dhcp or do we use networkd here too?)

dustymabe:
> Or, maybe that was a transitional thing - maybe we say we're pure Ignition now.

yeah this was mostly meant to be transitional (i booted a local qcow and didn't attach the user data ISO and saw failures and this was a knee jerk PR). Sorry if it was a bit pre-mature.

cgwalters:
Not a blocker but note-to-self: [qcow2-to-vagrant](https://github.com/cgwalters/qcow2-to-vagrant) may have to learn about ignition too.

dustymabe:
> qcow2-to-vagrant

is that better than the two kickstart model we've been using today (where we have one for vagrant and one for base) ?

cgwalters:
> is that better than the two kickstart model 

It's not better, it just has the value of being able to work on an existing qcow2 - like for RHELAH.  But yeah, we can also switch to the two-image build, and that would also give us vagrant-virtualbox which is a huge plus.

cgwalters:
> (does is still take care of dhcp or do we use networkd here too?)

dhcp is enabled by default in the kickstart files we use today.  networkd is a whole other discussion...one that will almost certainly come up in CL compatibility.

ashcrow:
> networkd is a an excellent example of a dividing line between "more Fedora" vs "more CL".

Yeah. I assume(d) we'd run with `networkmanager` rather than `networkd` to skew closer to the RH family of OS's. But if there is a good reason to switch we should document it and present it.

cgwalters:
One thing that wasn't entirely clear to me at first...this PR seems to depend on/be entangled with the [cloud provider detection](https://github.com/openshift/os/issues/101) issue - if we nuke cloud-init, but don't enable Ignition for e.g. OpenStack.  Similar for other providers that use cloud-init.

dustymabe:
will cancel PR. it was premature.

PR #90, opened 2018-06-08
jenkins/treecompose: Add version to build description too

cgwalters:
This is what the cloud bits do; should help in lining up
treecompose vs images.

(Though I don't know why there are two checksums appearing both
 in treecompose and cloud today)


cgwalters:
Not tested locally.

PR #91, opened 2018-06-08
cloud/treecompose: Log more when doing dry runs

cgwalters:
I also tweaked things to avoid the `tmp` naming in the local mirror
repo - if we mirror with `--mirror` then we don't need a prefix.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -57,17 +57,18 @@ node(NODE) {
         ]) {
             sh """
                 rm -rf repo && ostree init --repo=repo --mode=bare-user
-                ostree --repo=repo remote add tmp --no-gpg-verify ${OSTREE_INSTALL_URL}
-                ostree --repo=repo pull --commit-metadata-only tmp ${ref}
+                ostree --repo=repo remote add rhcos --no-gpg-verify ${OSTREE_INSTALL_URL}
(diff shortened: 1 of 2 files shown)


cgwalters:
Not tested locally.

Jenkinsfile.cloud
```diff
@@ -57,24 +57,26 @@ node(NODE) {
...
+        latest_commit = utils.sh_capture("ostree --repo=repo rev-parse ${ref}")
+        latest_version = utils.get_rev_version("repo", "${latest_commit}")
         if (fileExists('force-nocache-build')) {
             force_nocache = readFile('force-nocache-build').trim();
         }
         last_build_version = utils.sh_capture("cat ${images}/cloud/latest/version.txt || :")
     }
 
     if (latest_version == last_build_version && latest_version != force_nocache) {
+        echo "Last built ${latest_version} (${latest_commit}) - no changes"
```
jlebon: These need to be `sh "echo ..."` or alternatively `print[ln]`

cgwalters: https://jenkins.io/doc/pipeline/steps/workflow-basic-steps/#echo-print-message right?  We're using it elsewhere.

jlebon: Woah, I could've sworn I've hit an issue with that in the past. Cool!

PR #95, opened 2018-06-11
WIP: os-content/pivot: New script

cgwalters:
This is a WIP/RFC for discussion.  The script seems to work well, though I have some work I want to land in rpm-ostree to [make this nicer](https://github.com/projectatomic/rpm-ostree/pull/1396).

This PR proposes to ship the script in this repo itself - we'd do something like turn this repo into an RPM too.   Just a bit easier than making a whole other repo for the script, but I have no opposition to doing that too.  (Or we could stick it in the release repo)

Any comments on the design/code are fine too!

### DIFF
--- a/os-content/pivot.py
+++ b/os-content/pivot.py
@@ -1,0 +1,91 @@
+#!/usr/bin/python
+# Execute an OS pivot - see https://github.com/openshift/os
+
+import os,sys,argparse,json,subprocess,time
+
+PIVOT_DONE_PATH = '/etc/os-container-pivot.stamp'
+PIVOT_NAME = 'ostree-container-pivot'
+
+def fatal(msg):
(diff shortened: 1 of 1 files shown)


ashcrow:
If this is going to run on RHCOS itself (and not in a container) it may make sense to prototype it in Python, etc.. but deliver it in a compiled language. We could avoid hard requirements for RHCOS **itself** to require language runtimes.

cgwalters:
Yeah, I agree - though not shipping Python is going to be rough for Atomic users...we haven't crossed that bridge really yet.

It argues for a separate repo anyways.  I'll do that.

ashcrow:
FWIW https://gist.github.com/ashcrow/4013030f3268b38a19bff5a1dddc4bca

PR #98, opened 2018-06-13
Add pivot in rdgo

ashcrow:
Signed-off-by: Steve Milner <smilner@redhat.com>

### DIFF
--- a/overlay.yml
+++ b/overlay.yml
@@ -38,3 +38,6 @@ components:
 
   - src: github:openshift/redhat-release-coreos
     spec: internal
+
+  - src: github:ashcrow/pivot
+    spec: internal


ashcrow:
This should be merged AFTER the command is added to RHCOS. Working on that now.

ashcrow:
Error is expected. ``pivot`` is not currently built as it's part of this PR.

cgwalters:
That said...we may have to either neuter the Prow job temporarily, or do this in two commits/PRs.  Seems like the latter shouldn't be too hard right?

ashcrow:
@cgwalters this PR now simply adds pivot to rdgo. https://github.com/openshift/os/pull/103 adds doc update and inclusion into the compose.

PR #99, opened 2018-06-13
cloud.ks: Remove nonfunctional initramfs regeneration

cgwalters:
This code has never worked; ostree stores initramfs in a separate
location, and this extra initramfs is just taking up space.

If we need these changes, I think we should just include them in
every initramfs.


jlebon:
Oh yeah, good catch. Hmm, we should probably remove this from the RHELAH kickstarts as well.
/lgtm

PR #100, opened 2018-06-13
Don't wait for triggered jobs to finish

jlebon:
Let's make the `treecompose` pipeline simply trigger new `cloud` jobs
instead of also waiting for it to finish. Similarly for `rdgo` and
`treecompose.


cgwalters:
Oops, yes!

/lgtm

PR #103, opened 2018-06-14
Include pivot

ashcrow:
Requires #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,13 +20,23 @@ Turn a [CentOS Atomic](https://wiki.centos.org/SpecialInterestGroup/Atomic/Downl
 $ lvm lvextend -r -l +25%FREE atomicos/root
 ```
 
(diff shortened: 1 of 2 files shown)


cgwalters:
The rdgo build is failing https://continuous-infra-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/rhcos/job/coreos-rhcos-rdgo/1301/artifact/log/failed/pivot/build.log

It looks like pivot assumes it's in a git repo.

jlebon:
A bit late to the party, but why do we need to include this in the host? We just need it to get to the host, right? Or is RHCOS also on double duty as the bootstrap OS for now?

ashcrow:
> Or is RHCOS also on double duty as the bootstrap OS for now?

Right :-). The ``README.md`` already explains how to do it manually. This makes it easier.

Issue #101, opened 2018-06-13
Ignition handling of various cloud providers


jlebon:
Right now, we have some magic in [bootengine](https://github.com/dustymabe/bootengine/) for detecting AWS based on Xen and OpenStack based on `dmidecode` output, though this is clearly not ideal. IIUC, CL had different OEM packages for each cloud that drops in a small grub cfg script which is used at runtime to build the kernel cmdline, e.g. https://github.com/coreos/coreos-overlay/blob/master/coreos-base/oem-gce/files/grub.cfg. 

One idea is to have a small libguestfs script that does something similar for each cloud-specific image we provide before uploading them?

cgwalters:
> One idea is to have a small libguestfs script that does something similar for each cloud-specific image we provide before uploading them?

That doesn't seem too bad for handling the AWS/Azure/GCE cases + OpenStack, but gets uglier the farther we take it; the more that literally the same disk image works for both AWS and OpenStack and VMWare, the better.

Eh.  I'm fine with that approach since others seemed negative on the dynamic detection but...it feels to me we can make something like the `dmidecode` fairly reliable in some cases - if it breaks we're going to notice.

bgilbert:
If platform detection breaks, it may break only on certain instance types or with certain platform configurations, which may take a while to notice.

cgwalters:
That's a valid concern, in fact the change Dusty made to do "xen == ec2" definitely breaks with the new KVM instances, however:

```
[root@ip-172-18-6-44 ~]# systemd-detect-virt 
kvm
[root@ip-172-18-6-44 ~]# dmidecode |grep 'Amazon EC2'
        Vendor: Amazon EC2
        Manufacturer: Amazon EC2
        Manufacturer: Amazon EC2
        Manufacturer: Amazon EC2
        Asset Tag: Amazon EC2
```

But eh...I'm fine switching to postprocessing too.

jlebon:
My issue with using output from `dmidecode` is that it's implicit. It's not really a stable "API" we can bank on. I see that e.g. AWS docs themselves promote using `/sys/hypervisor/uuid` or `dmidecode` (but only for HVM instances): https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/identify_ec2_instances.html, and even then, there are caveats: https://serverfault.com/a/700771. From a quick search, I couldn't find such firm declarations from the other clouds.

[...]

PR #106, opened 2018-06-14
WIP: Stamp coreos.oem.id in images

cgwalters:
For https://github.com/openshift/os/issues/101


cgwalters:
Untested: I'll polish this off if people are OK with the approach.

scripts/coreos-oemid
```diff
@@ -0,0 +1,61 @@
...
+src=$1
+dest=$2
+oemid=$3
+
+fatal() {
+    echo "error: $@" 1>&2
+    exit 1
+}
+
+tmpd=$(mktemp -d /tmp/qcow2-to-vagrant.XXXXXX)
```
cgwalters: I forked this from https://github.com/cgwalters/qcow2-to-vagrant btw.

Jenkinsfile.cloud
```diff
@@ -141,7 +143,9 @@ node(NODE) {
     parallel par_stages; par_stages = [:]
 
     stage("Finalizing, generate metadata") {
-        sh "gzip ${qcow}"
+        sh """env LIBGUESTFS_BACKEND=direct ${WORKSPACE}/scripts/coreos-oemid ${qcow} ${qcow}.openstack
```
jlebon: This is missing a final `openstack` argument, right?

scripts/coreos-oemid
```diff
@@ -0,0 +1,61 @@
...
+gf list-filesystems > ${tmpd}/filesystems.txt
+vg=/dev/coreos/root
+if ! grep -qFe "${vg}" ${tmpd}/filesystems.txt; then
+    sed -e 's,^,# ,' < ${tmpd}/filesystems.txt
+    fatal "Missing LVM VG ${vg} in filesystems"
+fi
+gf mount "${vg}" /
+gf mount "/dev/sda1" /boot
+stateroot=/ostree/deploy/$(gf ls /ostree/deploy)
+rootdir=${stateroot}/deploy/$(gf ls ${stateroot}/deploy | grep -v \.origin)
```
jlebon: Wait, why do we need `stateroot` and `rootdir`?

cgwalters: We don't.  I kept it since it's convenient if we need to change something outside of `/boot`, but I can just comment it out for now.

cgwalters:
OK, so we also need to back out the patches in bootengine.  @dustymabe agreed with all of this?

dustymabe:
> OK, so we also need to back out the patches in bootengine. 

yeah we can do that but it's not a blocker since it will still use what was provided on the kernel command line if it was provided there. 

> @dustymabe agreed with all of this?

seems ok

jlebon:
Once this is in, I can take care of verifying it works on OpenStack.

cgwalters:
I'm testing this out now in my staging job.

cgwalters:
OK, this builds although I didn't test the result as I was using `DRY_RUN`; need to tweak the code to support dumping artifacts to a non-default location.

jlebon:
Let's just get it in and test it out!
/lgtm

PR #107, opened 2018-06-15
Initial how to build info and Make targets

ashcrow:
A question has come up over how one would build RHCOS. The original Container Linux has [this](https://coreos.com/os/docs/latest/sdk-modifying-coreos.html) document explaining how to do so. Today the closest we have is for one to look at ``Jenkinsfile.*``s.

This would let people run one or more ``make`` targets locally to build a local development version of RHCOS.

See https://github.com/openshift/os/issues/108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,11 @@ COMPOSEFLAGS ?=
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 CACHE_ARGS := $(shell if test -d cache; then echo $(shell pwd)/cache; fi)
 
+REPO ?= /srv/rhcos/repo
+INSTALLER_TREE_URL ?= https://dl.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os
+OSTREE_INSTALL_URL ?= ${REPO}
+REF ?= openshift/3.10/x86_64/os
+
 all: rpmostree-compose
 
 .PHONY: syntax-check
@@ -22,10 +27,52 @@ init-ostree-repo:
 	ostree --repo=build-repo init --mode=bare-user
 	ostree --repo=repo init --mode=archive
 
+.PHONY: check-tools
+check-tools:
+	which ostree
+	which rpm-ostree
+	which rpmdistro-gitoverlay
+	which imagefactory
+	which qemu-img
+	which mock
+
+
+# Runs all targets needed to create an image
+.PHONY: build-rhcos
+build-rhcos: check-tools clean repo-refresh rdgo rpmostree-compose os-image
+
+# Cleans up
+.PHONY: clean
+clean:
+	rm -rf build build-0 build-1 src snapshot ${REPO}
+
+# Pulls sources, builds packages, and creates a dnf repo for use
+.PHONY: rdgo
+rdgo:
+	rpmdistro-gitoverlay init
+	rpmdistro-gitoverlay resolve --fetch-all
+	rpmdistro-gitoverlay build
+
+
+# Composes an ostree
 .PHONY: rpmostree-compose
 rpmostree-compose: ${ROOT_DIR}/openshift.repo init-ostree-repo
+	mkdir -p ${REPO}
+	ostree init --repo=${REPO} --mode=archive
 	if test -d cache; then cachedir='--cachedir $(shell pwd)/cache'; fi && \
 	  cd ${ROOT_DIR} && set -x && \
 	  rpm-ostree compose tree $(COMPOSEFLAGS) $${cachedir:-} --repo=$(shell pwd)/build-repo host.yaml
 	ostree --repo=repo pull-local build-repo
 	ostree --repo=repo summary -u
+
+# Makes an image using the ostree
+.PHONY: os-image
+os-image:
+	ostree --repo=repo remote add rhcos --no-gpg-verify ${OSTREE_INSTALL_URL}
+	ostree --repo=repo pull --mirror --commit-metadata-only rhcos
+	sed -i 's,\(<url>\).*\(<\/url\),\1${INSTALLER_TREE_URL}\2,' rhcos.tdl
+	sed -i 's,@@OSTREE_INSTALL_URL@@,${OSTREE_INSTALL_URL},' cloud.ks
+	sed -i 's,@@OSTREE_INSTALL_REF@@,${REF},' cloud.ks
+	imagefactory --debug base_image --file-parameter install_script cloud.ks --parameter offline_icicle True rhcos.tdl
+	export IMAGE=`ls /var/lib/imagefactory/storage/*.body`
+	qemu-img convert -f raw -O qcow2 ${IMAGE} rhcos-devel.qcow2
--- a/README-development.md
+++ b/README-development.md
@@ -1,3 +1,30 @@
+# Operating System
+
+## Prerequisites
+
+### Usually packaged by operating systems
+- git
+- ostree
+- rpm-ostree
+- qemu-img
+- fedpkg
+- mock
+
+### May need to be built from source
+- [coreos-assembler](https://github.com/cgwalters/coreos-assembler/): [container](http://quay.io/cgwalters/coreos-assembler)
+- [rpmdistro-gitoverlay](https://github.com/projectatomic/rpmdistro-gitoverlay)
+- [imagefactory-plugins-TinMan](https://github.com/redhat-imaging/imagefactory)
+
+## Building
+
+- Clone ``openshift/os``
+- Move into the cloned repo
+- Build packages from source repos: ``make rdgo``
+- Make the ostree: ``make rpmostree-compose``
+- Make the qcow2: ``make os-image OSTREE_INSTALL_URL=<OSTREE_INSTALL_URL_HERE>``
+
+# Container Image
+
 This repository uses [https://docs.docker.com/develop/develop-images/multistage-build/](multi-stage) builds.
 If you're using Project Atomic/RHEL Docker, your best bet is to build [OpenShift imagebuilder](https://github.com/openshift/imagebuilder)
 docker.
--- a/README-development.md
+++ b/README-development.md
@@ -11,11 +11,12 @@
 - mock
 
 ### May need to be built from source
-- [coreos-assembler](https://github.com/cgwalters/coreos-assembler/): [container](http://quay.io/cgwalters/coreos-assembler)
 - [rpmdistro-gitoverlay](https://github.com/projectatomic/rpmdistro-gitoverlay)
 - [imagefactory-plugins-TinMan](https://github.com/redhat-imaging/imagefactory)
 
 ## Building
+
+Choose a local mirror to use for OSTREE_INSTALL_URL from [the list](https://admin.fedoraproject.org/mirrormanager/mirrors/Fedora/28/x86_64)
 
 - Clone ``openshift/os``
 - Move into the cloned repo


README-development.md
```diff
@@ -1,3 +1,39 @@
...
+
+## Prerequisites
+
+- git
+- coreos-assembler
+- rpmdistro-gitoverlay
+- ostree
+- rpm-ostree 
+- imagefactory-plugins-TinMan
+- qemu-img
```
jlebon: Hmm, might be worth inlining these requires into the building steps. That way if one just wants to know how to compose an OSTree, they don't need to install e.g. imgfac (which pulls in libvirt + systemd).

README-development.md
```diff
@@ -1,3 +1,39 @@
...
+- Make the ostree
+ - ``export REPO=/tmp/repo``
+ - ``ostree init --repo=$REPO --mode=archive --src $REPO/ --dest=$REPO``
+ - ``coreos-assembler --repo=$REPO host.yaml``
+ - ``ostree summary --repo=$REPO --update --dest $REPO/ --src=$REPO``
+- Make the qcow2
+ - ``ostree --repo=repo remote add rhcos --no-gpg-verify ${OSTREE_INSTALL_URL}``
+ - ``ostree --repo=repo pull --mirror --commit-metadata-only rhcos``
+ - ``imagefactory --debug base_image --file-parameter install_script cloud.ks --parameter offline_icicle True rhcos.tdl``
+ - ``export IMAGE=`ls /var/lib/imagefactory/storage/*.body```
```
jlebon: Oh neat! I didn't know that you could use double ticks on the outside to safely use single ticks `` ` `` inside them!

ashcrow: I'm not sure you can :laughing:

README-development.md
lucab: Offline suggestion: link the exotic tools to their repo (or, if they have it, to their "how to install" doc).

README-development.md
```diff
@@ -1,3 +1,39 @@
+# Operating System
```
cgwalters: I run basically all of this from a container, and so does the pipeline (except for Imagefactory for bad reasons).  I think we should recommend that as well; it's mentioned below for the rpm-ostree bits.  In fact we could just recommend pulling `quay.io/cgwalters/coreos-assembler` and running things in there right?

README-development.md
cgwalters: Most of this is `yum -y install`, with a few exceptions, those we should call out.

README-development.md
```diff
@@ -1,3 +1,39 @@
...
+
+- Clone ``openshift/os``
+- Move into the cloned repo
+- Build packages from source repos:
+ - ``rpmdistro-gitoverlay init``
+ - ``rpmdistro-gitoverlay resolve --fetch-all``
+ - ``rpmdistro-gitoverlay build``
+- Make the ostree
+ - ``export REPO=/tmp/repo``
+ - ``ostree init --repo=$REPO --mode=archive --src $REPO/ --dest=$REPO``
```
cgwalters: These `--src/--dest` args seem like a copy/paste error from rsync?

README-development.md
```diff
@@ -1,3 +1,39 @@
...
+ - ``coreos-assembler --repo=$REPO host.yaml``
+ - ``ostree summary --repo=$REPO --update --dest $REPO/ --src=$REPO``
```
cgwalters: Ditto.

README-development.md
```diff
@@ -1,3 +1,39 @@
...
+- Clone ``openshift/os``
+- Move into the cloned repo
+- Build packages from source repos:
+ - ``rpmdistro-gitoverlay init``
+ - ``rpmdistro-gitoverlay resolve --fetch-all``
+ - ``rpmdistro-gitoverlay build``
+- Make the ostree
+ - ``export REPO=/tmp/repo``
+ - ``ostree init --repo=$REPO --mode=archive --src $REPO/ --dest=$REPO``
+ - ``coreos-assembler --repo=$REPO host.yaml``
```
cgwalters: rpm-ostree from git master supports the yaml...but this is OK for now.

README-development.md
ashcrow: I actually wasn't 100% sure the right way to convert these lines from the Jenkinsfile without actually going through and testing out to see what happens.

README-development.md
```diff
@@ -1,3 +1,39 @@
...
+ - ``ostree summary --repo=$REPO --update --dest $REPO/ --src=$REPO``
+- Make the qcow2
+ - ``ostree --repo=repo remote add rhcos --no-gpg-verify ${OSTREE_INSTALL_URL}``
+ - ``ostree --repo=repo pull --mirror --commit-metadata-only rhcos``
+ - ``imagefactory --debug base_image --file-parameter install_script cloud.ks --parameter offline_icicle True rhcos.tdl``
```
jlebon: This is missing two `sed`s for subbing out `OSTREE_INSTALL_URL` and `OSTREE_INSTALL_REF`. Though we can probably hardcode the latter at least now. (It was originally written to be usable for both CentOS and RHEL-derived input).

README-development.md
```diff
@@ -1,3 +1,39 @@
+# Operating System
+
+## Prerequisites
+
+- git
+- coreos-assembler
+- rpmdistro-gitoverlay
```
lucab: This additionally requires `fedpkg` and `mock`.

ashcrow:
Updates based on feedback.

README-development.md
cgwalters: It doesn't look like `--src $REPO/ --dest=$REPO` was dropped, definitely should be.

ashcrow: It's now removed

README-development.md
ashcrow: Switched ``coreos-assembler`` to use ``podman``

README-development.md
```diff
@@ -1,3 +1,49 @@
...
+- Move into the cloned repo
+- Build packages from source repos:
+ - ``rpmdistro-gitoverlay init``
+ - ``rpmdistro-gitoverlay resolve --fetch-all``
+ - ``rpmdistro-gitoverlay build``
+- Make the ostree
+ - ``export REPO=/tmp/repo``
+ - ``ostree init --repo=$REPO --mode=archive``
+ - ``podman run --privileged --rm -v $REPO:/srv quay.io/cgwalters/coreos-assembler``
+ - ``cd /srv && coreos-assembler --repo=/srv --cachedir=cache host.yml``
```
miabbott: I used the `coreos-assembler` container to run all the commands up to here (including the rdgo ones).  

If you follow the steps verbatim, the `coreos-assembler` command will fall over because there is no `openshift.repo` file available in `openshift/os`

I was able to generate the `openshift.repo` with the `generate-openshift-repo` command in the `openshift/os` repo, but then we fall over due to missing the Red Hat GPG key.

So I propose that the additional steps of:

1) running `generate-openshift-repo`
2) copying the Red Hat GPG key into `/etc/pki/rpm-gpg/`

...then running the `coreos-assembler` command.

(It might also be worth noting separately that the `coreos-assembler` command is not going to work unless you have access to the Red Hat stage CDN, just to dissuade anyone who is curious)

lucab: I was also going through the steps, and hit the same. For the former there seems to be already a `make repo-refresh` (similarly for init). For the latter I'm confused because it seems to bail on a `gpgcheck=0` repo; is there a way to hook repo configs to relative paths from this git-repo?

README-development.md
```diff
@@ -1,3 +1,49 @@
...
+### Usually packaged by operating systems
+- git
+- ostree
+- rpm-ostree
+- qemu-img
+- fedpkg
+- mock
+
+### May need to be built from source
+- [coreos-assembler](https://github.com/cgwalters/coreos-assembler/): [container](http://quay.io/cgwalters/coreos-assembler)
```
cgwalters: Note https://github.com/openshift/os/pull/119 should make this unnecessary.

cgwalters:
I'm fine merging this as is, it doesn't need to be perfect, and will clearly change over time.

README-development.md
cgwalters: Yeah, we can.

ashcrow:
@cgwalters I agree it doesn't have to be perfect. However, I'm going to turn these commands into make targets, and update this doc. At that point I'd like to have it re-reviewed for merging.

README-development.md
ashcrow: Is `coreos-assembler` deprecated at this point as well?

ashcrow:
Updated with make targets. I don't think it's quite ready to merge but close. Questions:

- Is there a default for ``OSTREE_INSTALL_URL``?
- Is [this](https://github.com/openshift/os/pull/107/files#diff-b67911656ef5d18c4ae36cb6741b7965R59) needed?

@lucab @jlebon PTAL

Makefile
```diff
@@ -3,6 +3,10 @@ COMPOSEFLAGS ?=
 ROOT_DIR := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 CACHE_ARGS := $(shell if test -d cache; then echo $(shell pwd)/cache; fi)
 
+REPO ?= /tmp/repo
```
jlebon: How about something like `/srv/rhcos/repo`?

README-development.md
```diff
@@ -11,35 +11,16 @@
...
- - ``ostree --repo=repo pull --mirror --commit-metadata-only rhcos``
- - ``sed -i 's,\(<url>\).*\(<\/url\),\1${INSTALLER_TREE_URL}\2,' rhcos.tdl``
- - ``sed -i 's,@@OSTREE_INSTALL_URL@@,${OSTREE_INSTALL_URL},' cloud.ks``
- - ``export REF="openshift/3.10/x86_64/os"``
- - ``sed -i 's,@@OSTREE_INSTALL_REF@@,${REF},' cloud.ks``
- - ``imagefactory --debug base_image --file-parameter install_script cloud.ks --parameter offline_icicle True rhcos.tdl``
- - ``export IMAGE=`ls /var/lib/imagefactory/storage/*.body```
- - ``qemu-img convert -f raw -O qcow2 $IMAGE rhcos-devel.qcow2``
+- Build packages from source repos: ``make rdgo``
+- Make the ostree: ``OSTREE_INSTALL_URL=<OSTREE_INSTALL_URL_HERE> make rpmostree-compose``
```
jlebon: Minor: I think the more standard way of setting vars when calling make is `make <target> VAR1=val1 VAR2=val2...`.

miabbott:
@ashcrow `OSTREE_INSTALL_URL` and `INSTALLER_TREE_URL` are sourced from the `pipeline-utils.groovy` file.

See here:  https://github.com/openshift/os/blob/master/pipeline-utils.groovy#L52

jlebon:
`INSTALLER_TREE_URL` will use the Fedora installer tree by default, so it doesn't *need* to be subbed out (though it'd be much faster if pointed at a local mirror). `OSTREE_INSTALL_URL`... I guess we can hardcode for now the URL to the internal repo?

miabbott:
The `ostree init` on [line 59](https://github.com/openshift/os/pull/107/files#diff-b67911656ef5d18c4ae36cb6741b7965R59) will silently no-op if the repo is already initialized.

Makefile
ashcrow: Sure. Will update.

ashcrow:
@miabbott Are you thinking that we should add that directory to the ``clean`` target to ensure it does not exist?

miabbott:
@ashcrow It would probably make sense in terms of keeping things tidy.

ashcrow:
@miabbott @jlebon updated

Makefile
```diff
@@ -22,10 +27,52 @@ init-ostree-repo:
...
 	  cd ${ROOT_DIR} && set -x && \
 	  rpm-ostree compose tree $(COMPOSEFLAGS) $${cachedir:-} --repo=$(shell pwd)/build-repo host.yaml
 	ostree --repo=repo pull-local build-repo
 	ostree --repo=repo summary -u
+
+# Makes an image using the ostree
+.PHONY: os-image
+os-image:
+	ostree --repo=repo remote add rhcos --no-gpg-verify ${OSTREE_INSTALL_URL}
+	ostree --repo=repo pull --mirror --commit-metadata-only rhcos
```
cgwalters: Hmm...I think if you're doing local builds you want to be using the local ostree repo.  That's what I do.  

The annoying pain point is it requires setting up a webserver.  See https://github.com/projectatomic/rpm-ostree-toolbox/pull/68 and there's other PRs/issues around this too.

cgwalters:
A high level question here too is how much of this flow should live in https://github.com/cgwalters/coreos-assembler

I think having an opinionated container makes sense (versus delivering as an RPM as `rpm-ostree-toolbox` did).

Anyways...I'm OK to merge this as is too.  Trying to change the Jenkinsfiles is going to be a whole other world.

README-development.md
```diff
@@ -11,35 +11,18 @@
...
- - ``ostree --repo=repo pull --mirror --commit-metadata-only rhcos``
- - ``sed -i 's,\(<url>\).*\(<\/url\),\1${INSTALLER_TREE_URL}\2,' rhcos.tdl``
- - ``sed -i 's,@@OSTREE_INSTALL_URL@@,${OSTREE_INSTALL_URL},' cloud.ks``
- - ``export REF="openshift/3.10/x86_64/os"``
- - ``sed -i 's,@@OSTREE_INSTALL_REF@@,${REF},' cloud.ks``
- - ``imagefactory --debug base_image --file-parameter install_script cloud.ks --parameter offline_icicle True rhcos.tdl``
- - ``export IMAGE=`ls /var/lib/imagefactory/storage/*.body```
- - ``qemu-img convert -f raw -O qcow2 $IMAGE rhcos-devel.qcow2``
+- Build packages from source repos: ``make rdgo``
+- Make the ostree: ``make rpmostree-compose OSTREE_INSTALL_URL=<OSTREE_INSTALL_URL_HERE> ``
```
cgwalters: I think this `INSTALL_URL` should be for `make os-image` right?

cgwalters: To be clear I'm OK to `/lgtm` after this small tweak, but up to you how much else you want to change.

ashcrow: Updated.

ashcrow:
@lucab you are correct. There is work that needs to happen before someone who didn't actively work on the Jenkinsfiles and work on the underlying infrastructure can follow and recreate what's going on. There is also a side thing @cgwalters is doing in terms of changing the pipeline which may change some things.

Unfortunately I don't think we will be able to really finish #108 until the pipeline relaxes in flux once again.

Issue #96, opened 2018-06-11
No sudoers groups exist with NOPASSWD


arithx:
By default it doesn't seem like any groups exist in `/etc/sudoers`. On Container Linux both the `%sudo` group & the `core` user have `NOPASSWD`. 

The default `sudoers` for CL (from `/usr/share/baselayout/sudoers`):
```
## Default sudoers for CoreOS

## Pass LESSCHARSET through for systemd commands run through sudo that call less.
## See https://github.com/coreos/bugs/issues/365.
Defaults env_keep += "LESSCHARSET"

## enable root and wheel
root ALL=(ALL) ALL
%wheel ALL=(ALL) ALL

## enable passwordless access for sudo
%sudo ALL=(ALL) NOPASSWD: ALL

## the core user has no password by default
core ALL=(ALL) NOPASSWD: ALL

## Read drop-in files from /etc/sudoers.d
## (the '#' here does not indicate a comment)
#includedir /etc/sudoers.d
```

cgwalters:
This is definitely going to be one of numerous origCL/newCL transition points.  Now on Fedora/Atomic classically there are no predefined users for humans to log in other than `root`.  One can add them with kickstart or cloud-init.  And while cloud-init does have a hardcoded default user (`fedora` on fedora, `cloud-user` on RHEL) - if one e.g. *does* specify a user ([personal example](https://github.com/cgwalters/ansible-personal/blob/master/cloud-init/user-data.secure)) then that user won't be created.

I'd lean a bit towards doing the same in our built-in ignition schema - it feels aesthetically cleaner.  Having a truly reserved user might be problematic in network-login scenarios etc.

cgwalters:
Strawman: something like:

```
diff --git a/dracut/30ignition/ignition-quench.clc b/dracut/30ignition/ignition-quench.clc
index 7fd4e82..ed57380 100644
--- a/dracut/30ignition/ignition-quench.clc
+++ b/dracut/30ignition/ignition-quench.clc
@@ -1,11 +1,15 @@
-# This file contains the source for ignition-quench.ign. This file becomes the
-# base system Ignition config, which makes Ignition install a oneshot systemd
-# unit that only runs on first boot to remove `coreos.first_boot` from the
-# kernel command line in /boot/grub2/grub.cfg.
+# This file contains the source for the base system Ignition config.  It
+# does a few things such as adding the `core` user if no users were specified,
+# and also adds oneshot systemd unit to mark the system as having been booted.
 #
[...]

PR #109, opened 2018-06-15
cloud: add in 'core' user for now

dustymabe:
workaround for https://github.com/openshift/os/issues/96


dustymabe:
i'm testing this now - will report back

arithx:
I have a WIP commit for mantle which will automatically inject the rules into an AWS machine's Ignition config. I've done some limited testing so far but it is working atm: https://github.com/arithx/mantle/commit/cddefc147eaaab0c1e0738c9227c2da6dd61a6ec

dustymabe:
so that just injects it in the ignition config when someone uses kola?

dustymabe:
testing worked :100:

arithx:
@dustymabe : Yeah, when the distribution is set to `rhcos` and it's running on AWS it injects into the Ignition config to create a file (the call itself takes `filesystem, path, contents string`).

arithx:
If we're going to use this can we also add the `core` user to the `wheel`, `adm`, & `systemd-journal` groups?

cgwalters:
I still feel like doing this in Ignition defaults will work best, but I have no problem doing this short term hack out of pragmatism.

PR #110, opened 2018-06-15
cloud.ks: Add core user to some groups

jlebon:
See https://github.com/openshift/os/pull/109#issuecomment-397651647.

(Untested).

### DIFF
--- a/cloud.ks
+++ b/cloud.ks
@@ -7,7 +7,7 @@ selinux --enforcing
 rootpw --lock --iscrypted locked
 # create core user for now
 # https://github.com/openshift/os/issues/96
-user --name=core
+user --name=core --groups='wheel,adm,systemd-journal'
 
 # Explicitly disable firewall since cloud providers generally provide
 # higher level firewall constructs (i.e. security groups).
--- a/host.yaml
+++ b/host.yaml
@@ -20,6 +20,8 @@ mutate-os-release: "7"
 postprocess-script: "compose-post.sh"
 etc-group-members:
   - wheel
+  - systemd-journal
+  - adm
   - docker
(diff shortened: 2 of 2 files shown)


jlebon:
/hold
Sanity checking...

cgwalters:
Ah sorry, had missed the hold.

cgwalters:
We likely need e.g.:

```
diff --git a/host.yaml b/host.yaml
index 5b290d7..9603ac5 100644
--- a/host.yaml
+++ b/host.yaml
@@ -20,6 +20,8 @@ mutate-os-release: "7"
 postprocess-script: "compose-post.sh"
 etc-group-members:
   - wheel
+  - systemd-journal
+  - adm
   - docker
 ignore-removed-users:
   - root
```

jlebon:
OK, I've confirmed this works! Though it requires us to update the `coreos-assembler` container to use the latest rpm-ostree for https://github.com/projectatomic/rpm-ostree/pull/1417.

ashcrow:
Since this is waiting for a merge and rebuild of another PR should this go back on hold?

ashcrow:
The other PR merged. Is it OK to merge now?

jlebon:
Hmm, let's leave it open until https://github.com/cgwalters/coreos-assembler/pull/5 is merged and the container rebuilt. If we merge this now, we'll be going from a `core` user without any groups to no `core` user at all (at least that's how my tests went locally before https://github.com/projectatomic/rpm-ostree/pull/1417 was merged).

jlebon:
`coreos-assembler` was rebuilt with the fix, so this should be good to go now!
/hold cancel

PR #113, opened 2018-06-18
cloud: Fixes for ec2 stamping

cgwalters:
We were seeing errors doing the ec2 stamping; things seem
reliable for me when I do it from a qcow2, then convert that to vmdk.
I didn't entirely pin down the problem, but I suspect it's a bug
in the qemu VMDK layer - we're running a long unsupported and insecure
Fedora 25 OS for the cloud bits.  Will look at updating our stack
before debugging farther.

In the meantime, this seems to work for me, although I didn't
test the resulting AMI.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -103,7 +103,7 @@ node(NODE) {
         """
     }
 
-    def dirname, image, commit, version, dirpath, qcow, vmdk, vagrant_libvirt
+    def dirname, image, commit, version, dirpath, qcow, vmdk, ec2, vagrant_libvirt
     stage("Postprocessing") {
         image = utils.sh_capture("ls /var/lib/imagefactory/storage/*.body")
         // just introspect after the fact to avoid race conditions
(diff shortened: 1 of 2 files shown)


Jenkinsfile.cloud
```diff
@@ -180,15 +193,15 @@ node(NODE) {
...
                     amijson=${dirpath}/aws-${AWS_REGION}.json
                     ore aws upload --region ${AWS_REGION} \
                         --ami-name 'rhcos_dev_${commit[0..6]}' \
                         --ami-description 'Red Hat CoreOS ${version} (${commit})' \
                         --bucket 's3://${S3_PRIVATE_BUCKET}/rhcos/cloud' \
-                        --file rhcos-aws-${commit}.vmdk \
+                        --file ${ec2} \
                         --name "rhcos_dev_${commit[0..6]}" \
                         --delete-object | tee \${amijson}
+                    rm -f ${WORKSPACE}/rhcos-aws-${commit}.vmdk
```
jlebon: Can this just be `rm -f ${ec2}`?

Jenkinsfile.cloud
```diff
@@ -131,21 +131,34 @@ node(NODE) {
...
         // We use direct as we hit SELinux issues otherwise
         sh "env LIBGUESTFS_BACKEND=direct ${WORKSPACE}/qcow2-to-vagrant/qcow2-to-vagrant ${qcow} ${vagrant_libvirt}"
     } }
+    // Execute parallel group
     parallel par_stages; par_stages = [:]
 
-    stage("Finalizing, generate metadata") {
+    // Convert the qcow2 to have an openstack stamp, and gzip
+    stage("OpenStack qcow2/ec2") {
         sh """env LIBGUESTFS_BACKEND=direct ${WORKSPACE}/scripts/coreos-oemid ${qcow} ${qcow}.openstack openstack
```
jlebon: Can't we make this part of the parallel group as well? Or does it fail then?

Jenkinsfile.cloud
```diff
@@ -131,21 +131,34 @@ node(NODE) {
...
     par_stages["vagrant-libvirt"] = { -> stage("Generate vagrant-libvirt") {
         // We use direct as we hit SELinux issues otherwise
         sh "env LIBGUESTFS_BACKEND=direct ${WORKSPACE}/qcow2-to-vagrant/qcow2-to-vagrant ${qcow} ${vagrant_libvirt}"
     } }
+    // Execute parallel group
     parallel par_stages; par_stages = [:]
 
-    stage("Finalizing, generate metadata") {
+    // Convert the qcow2 to have an openstack stamp, and gzip
+    stage("OpenStack qcow2/ec2") {
```
jlebon: Even though we're not actually running `qemu-img convert`, let's just call this `Generate openstack` to be consistent?

cgwalters:
OK, reworked :arrow_up:

PR #115, opened 2018-06-18
host: drop kexec-tools

miabbott:
The `kdump` service was failing to start because there was no
`crashkernel` option passed to the kernel.  @cgwalters suggested we
drop it, so let's just remove all of the `kexec-tools` package.
Nothing depends on the package, so I don't believe this should break
anything.

### DIFF
--- a/host.yaml
+++ b/host.yaml
@@ -68,7 +68,7 @@ packages:
  # Time sync
  - chrony
  # Extra runtime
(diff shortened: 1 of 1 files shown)


miabbott:
This also drops out `snappy` as a side-effect

ashcrow:
Waiting for CI to pass.

Issue #117, opened 2018-06-18
network.service is enabled


cgwalters:
Shouldn't be.  We should being using NM.  Offhand...I suspect this is `rd.neednet=1` which we're using for Ignition - dracut may have some rules somewhere which is telling the target to enable `network.service`?  If so I'm not finding it.

PR #118, opened 2018-06-19
compose-post: Remove network.service

cgwalters:
I still didn't debug why this is getting enabled, but there's
no reason to ship it, so let's delete it, then it can't be enabled.

Closes: https://github.com/openshift/os/issues/117


jlebon:
Let's try it out!
/lgtm

PR #119, opened 2018-06-19
Makefile: s/coreos-assembler/rpm-ostree compose tree/

cgwalters:
Now that https://github.com/cgwalters/coreos-assembler/pull/5 has
landed, there's no need to use the wrapper command.

That said, we may revisit this more later, but for now it makes
things more painful when working in e.g. an existing pet container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ FROM quay.io/cgwalters/coreos-assembler AS build
 COPY RPM-GPG-KEY-* /etc/pki/rpm-gpg/
 COPY . /srv/build/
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Ugh, so this blocks on using rpm-ostree git master too.

jlebon:
Wait, how come it worked the second time? (Actually, the better question is why did it even *fail* the first time?) Did you do some magic in the background?

cgwalters:
What I was trying to debug is whether `imagebuilder` did a pull of images at build time or not.  I suspect it doesn't.  We probably got scheduled on another node.

jlebon:
Ahh, gotcha.
/lgtm

PR #122, opened 2018-06-20
Add sudo group

cgwalters:
This increases compatibility with the default CL configs.

Closes: https://github.com/openshift/os/issues/96


compose-post.sh
```diff
@@ -30,6 +30,13 @@ set -x
 cp -f /usr/lib/locale/locale-archive /usr/lib/locale/locale-archive.tmpl
 build-locale-archive
 
+# https://github.com/openshift/os/issues/96
+# sudo group https://github.com/openshift/os/issues/96
+echo '%sudo        ALL=(ALL)       NOPASSWD: ALL' > /etc/sudoers.d/coreos-sudo-group
+# Also set up sudo access for core user for now...but
+# we may require users to explicitly ask for this.
+echo 'core ALL=(ALL) NOPASSWD: ALL' > /etc/sudoers.d/coreos-core-user
```
jlebon: Since right now, our hack for adding the core user is in `cloud.ks`, we might as well keep this part there as well, e.g.:

```diff
-user --name=core --groups='wheel,adm,systemd-journal'
+user --name=core --groups='wheel,adm,systemd-journal,sudo'
```

?

dustymabe:
what does the sudo group give us over wheel?

arithx:
@dustymabe: nothing, this is more intended as a compat between existing CL setups.

compose-post.sh
cgwalters: Good idea!  Updated :arrow_down:

cgwalters:
> what does the sudo group give us over wheel?

It defaults to `NOPASSWD`.

Basically, I tried to use the [CLC from the CL users page](https://coreos.com/os/docs/latest/adding-users.html) (though I knew to drop `docker`) and hit the fact that `sudo` didn't work.

I think we're going to have a fuzzy line here for what we merge versus what we require people to change.

For example, my strawman proposal is that we update that page to drop the password (why would you have a password *and* a key?)  and also drop the `docker` group.

bgilbert:
> For example, my strawman proposal is that we update that page to drop the password (why would you have a password and a key?) and also drop the docker group.

You wouldn't have both; it's meant to show how to use CLC features.

In general, we shouldn't be addressing incompatibilities with CL by changing the CL documentation; we should have a separate migration guide (and tooling etc.).  For one thing, users will have existing configs that embed the same assumptions the documentation does.

compose-post.sh
```diff
@@ -30,6 +30,10 @@ set -x
 cp -f /usr/lib/locale/locale-archive /usr/lib/locale/locale-archive.tmpl
 build-locale-archive
 
+# https://github.com/openshift/os/issues/96
+# sudo group https://github.com/openshift/os/issues/96
```
jlebon: Not worth a respin, though looks like duplicate comments.

cgwalters:
> In general, we shouldn't be addressing incompatibilities with CL by changing the CL documentation; 

Agree, and this is not the right place to discuss it, but just briefly: First, I think we need a page with a "recommended CLC" for trying out a single node instance, which (IMO) is "ssh key".  Things with passwords I think should just be in a "reference" section.

I find the https://coreos.com/os/docs/latest/provisioning.html page very confusing as it starts out talking about `etcd`.

I guess all of this gets into a deep question around whether we want to have something like the [existing CL CF templates](https://coreos.com/os/docs/latest/booting-on-ec2.html) which make a "cluster", or leave that to software which makes a Kube cluster.

PR #130, opened 2018-06-22
Jenkinsfile.cloud: start producing QEMU image

arithx:
Not sure if there's a better way, open to suggestions

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -150,13 +150,14 @@ node(NODE) {
     } }
     par_stages["openstack"] = { -> stage("Generate OpenStack") {
         sh """env LIBGUESTFS_BACKEND=direct ${WORKSPACE}/scripts/coreos-oemid ${qcow} ${qcow}.openstack openstack
-              gzip < ${qcow}.openstack > ${qcow}.gz
+              gzip < ${qcow}.openstack > ${qcow}.openstack.gz
               rm ${qcow}.openstack"""
     } }
     // Execute parallel group
     parallel par_stages; par_stages = [:]
 
     stage("Finalizing, generate metadata") {
+        sh "gzip < ${qcow} > ${qcow}.qemu.gz"
         // Everything above in parallel worked on qcow, we're done with it now
         sh "rm ${qcow}"
         sh "ln -sfn ${dirname} ${images}/cloud/latest"


jlebon:
Hmm, the unstamped qcow2 we generate in the "Postprocessing" stage *should* work directly with qemu using the firmware config approach. We have a bit in bootengine that detects it: https://github.com/dustymabe/bootengine/blob/bf3b454db89bcff82d01b472786821bd458d3593/dracut/30ignition/ignition-generator#L105.

So maybe let's first try not deleting that qcow2 and just publishing it too? Right now we're deleting it here: https://github.com/openshift/os/blob/99d04773609d81867c4a33500023befd966a9c2f/Jenkinsfile.cloud#L161

arithx:
Updated to drop the `qcow` deletion instead.

cgwalters:
Hmm...won't ignition not know to look for the config drive if there's no oem?  This feels like it's more likely something going wrong with RHCOS + configdrive?

(Drive by from reading the ignition code, looks like it's not using a mount namespace, and it's not umounting AFAICS, so it'll leak the mount)

jlebon:
OK, I just tested that this works!

> Hmm...won't ignition not know to look for the config drive if there's no oem?

If there's no oem, we'll default to `qemu`, which will make ignition use the QEMU firmware config.

I was trying to also make it work using the openstack stamped qcow2 and attaching an ISO labeled `config-2`, though Ignition wasn't seeing it. We might need a `udevadm settle` somewhere.

I'll have to look more into this next week; it'd be nice if we could get that working rather than outputting two qcow2s, though I'm cool if we want to get something like this in in the meantime. (An easy workaround is on first boot to delete the `coreos.oem.id=` kernel arg and once fully booted from the grub cfg).

Jenkinsfile.cloud
```diff
@@ -158,7 +158,6 @@ node(NODE) {
 
     stage("Finalizing, generate metadata") {
         // Everything above in parallel worked on qcow, we're done with it now
-        sh "rm ${qcow}"
```
jlebon: Hmm, we should probably gzip this one too. Maybe something like:

```
gzip < ${qcow} > ${qcow}.local.gz
```

?

(Or alternatively rename the openstack one to `${qcow}.openstack.gz` and this one just `${qcow}.gz`).

cgwalters:
Wow, I'd never heard of that.  Interesting.  See also [this](https://serverfault.com/questions/856741/load-qemu-firmware-configuration-device-in-virt-manager) - the libvirt trying to hide qemu is actually pretty annoying.

Anyways I'll let @jlebon do the lgtm.

Jenkinsfile.cloud
arithx: Hmm, how about `${qcow}.openstack.gz` & `${qcow}.qemu.gz`

jlebon: Sure, that works. The `qemu` is a bit misleading since it's not actually explicitly stamped with `coreos.oem.id=qemu`. Though I suppose that's the target use case.

Jenkinsfile.cloud
```diff
@@ -158,7 +158,6 @@ node(NODE) {
 
     stage("Finalizing, generate metadata") {
         // Everything above in parallel worked on qcow, we're done with it now
```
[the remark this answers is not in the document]
arithx: Going to leave this comment (and restore the `rm`) placing a `sh` line above to handle `gzip`ing the `${qcow}.qemu.gz`

arithx:
:arrow_up: changes pushed

PR #131, opened 2018-06-22
cloud: Share AMI with new OpenShift AWS CI account

jlebon:
This will allow the new AWS CI account to make use of our uploaded AMIs.


cgwalters:
This looks sane to me!  Will give a few mins for any other reviewers.

arithx:
LGTM

Issue #133, opened 2018-06-22
Fails to extend disk on AWS with NVMe


cgwalters:
We dropped cloud-init which has disk resizing code, and it turns out Ignition doesn't do this, see https://github.com/coreos/init/blob/master/scripts/extend-filesystems

Now the next problem we have is that script doesn't handle LVM...in practice what we did with Fedora Atomic was that it was actually [container-storage-setup](https://github.com/projectatomic/container-storage-setup) that does this, I need to figure out why that's not working.

cgwalters:
OK, `container-storage-setup.service` isn't running because...`docker.service` isn't?  And we're missing `growpart`.  Fixing.

cgwalters:
Now tested, this fixed it.

cgwalters:
Fixed by https://github.com/openshift/os/pull/134

cgwalters:
The next problem here is:

```
Jun 25 20:43:30 localhost.localdomain container-storage-setup[916]: Physical volume "/dev/nvme0n1p2" changed
Jun 25 20:43:30 localhost.localdomain container-storage-setup[916]: FAILED: /dev/nvme: does not exist
```

i.e. c-s-s falls over for NVMe.  This was supposed to be fixed in https://github.com/projectatomic/container-storage-setup/pull/138#issuecomment-283870272 but AFAICS that doesn't help.

PR #134, opened 2018-06-22
host: Explicitly include container-storage-setup and growpart

cgwalters:
We lost `cloud-utils-growpart` when we dropped `cloud-init`.
Should help with https://github.com/openshift/os/issues/133
although I haven't had time to test yet, but it's what we need to do now.


cgwalters:
To be clear, we are including `container-storage-setup` today via `docker`, but it's worth calling out.  However, it doesn't have a hard dependency on `cloud-utils-growpart` since it's not needed for physical installs, it's just cloud images that we enable it.

PR #135, opened 2018-06-23
Fix ConditionFirstBoot= by removing /etc/machine-id

cgwalters:
Requires: https://github.com/projectatomic/rpm-ostree/pull/1425

We want systemd's ConditionFirstBoot to fire. The
primary rationale here is that we're adopting Ignition for Fedora CoreOS,
and having ConditionFirstBoot= function will help a lot, as the idea
is it only runs once.  Further, this ensures that any presets generated
in the initramfs work.

### DIFF
--- a/compose-post.sh
+++ b/compose-post.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 
-set -e
+set -xe
+
+# See machineid-compat in host.yaml.
+# Since that makes presets run on boot, we need to have our defaults in /usr
+ln -sfr /usr/lib/systemd/system/{multi-user,default}.target
 
 # The loops below are too spammy otherwise...
 set +x
--- a/host.yaml
+++ b/host.yaml
@@ -17,6 +17,8 @@ initramfs-args:
   - "iscsi"
 automatic_version_prefix: "3.10-7.5"
 mutate-os-release: "7"
+# https://github.com/projectatomic/rpm-ostree/pull/1425
+machineid-compat: false
 postprocess-script: "compose-post.sh"
 etc-group-members:
   - wheel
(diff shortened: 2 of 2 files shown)


cgwalters:
/hold

Since it requires the rpm-ostree PR.

ashcrow:
@cgwalters does this change what @Rahuls0720 has been working on?

cgwalters:
Yep, see https://github.com/dustymabe/bootengine/pull/11

cgwalters:
Ok container is built, this should be good to go :train:.

/hold cancel

compose-post.sh
```diff
@@ -1,6 +1,15 @@
 #!/usr/bin/env bash
 
-set -e
+set -xe
+
+# See machineid-compat in host.yaml.
+# Since that makes presets run on boot, we need to have our defaults in /usr
+ln -sfr /usr/lib/systemd/system/{multi-user,default}.target
+# This should be in our -release package, but let's do this here for now
```
dustymabe: how do we track removing this workaround and putting it in the proper place?

ashcrow: I'd say make an issue and tag it into Jira so it's tech debt that doesn't get lost.

cgwalters: https://github.com/openshift/redhat-release-coreos/pull/3

cgwalters:
Yep, I wrote the same in comments.  I filed as https://github.com/projectatomic/rpm-ostree/issues/1426 but I'd like to get out of the weeds on this one, we can address that later.

jlebon:
(LGTM BTW, not sure if you wanted to first merge https://github.com/openshift/redhat-release-coreos/pull/3 and drop the presets from here right away).

PR #136, opened 2018-06-25
cloud: Fix sha256sum generation

jlebon:
Minor regression from #130. Also now start doing `sha256sum` for the new
qcow image.


cgwalters:
Seems sane to me, will let arithx do the lgtm.

PR #138, opened 2018-06-25
jenkins: Move container to registry.svc.ci.openshift.org/rhcos/os:latest

cgwalters:
See https://github.com/openshift/release/issues/972

This is external, but auth-protected, so it can work for people outside
of RHT if they have the secret, which aids testing.

### DIFF
--- a/Jenkinsfile.oscontainer
+++ b/Jenkinsfile.oscontainer
@@ -22,12 +22,17 @@ node(NODE) {
     }
 
     withCredentials([
-        string(credentialsId: params.REGISTRY, variable: 'REGISTRY'),
+        usernameColonPassword(credentialsId: params.REGISTRY_CREDENTIALS, variable: 'CREDS'),
     ]) {
-        docker.withRegistry("${REGISTRY}", params.REGISTRY_CREDENTIALS) {
+        def (username, password) = "${CREDS}".split(':')
+        sh """set +x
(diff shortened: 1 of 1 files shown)


Jenkinsfile.oscontainer
```diff
@@ -22,12 +22,17 @@ node(NODE) {
...
 
     withCredentials([
-        string(credentialsId: params.REGISTRY, variable: 'REGISTRY'),
+        usernameColonPassword(credentialsId: params.REGISTRY_CREDENTIALS, variable: 'CREDS'),
     ]) {
-        docker.withRegistry("${REGISTRY}", params.REGISTRY_CREDENTIALS) {
+        def (username, password) = "${CREDS}".split(':')
+        sh """set +x
+              echo docker login -e 'unused@example.com' -u 'username' -p password registry.svc.ci.openshift.org
+              docker login -e 'unused@example.com' -u "${username}" -p "${password}" registry.svc.ci.openshift.org
```
ashcrow: Just verifying, does the version of docker in use still support ``-e``? I believe it was dropped.

cgwalters: Sadly because we don't apply updates to our CI nodes, it's actually required there right now.  It's a bit tempting to rebase them on CL or something, but let's not do that right now...

PR #140, opened 2018-06-26
compose-post: Remove readahead

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1594984

### DIFF
--- a/compose-post.sh
+++ b/compose-post.sh
@@ -42,3 +42,7 @@ echo '%sudo        ALL=(ALL)       NOPASSWD: ALL' > /etc/sudoers.d/coreos-sudo-g
 # https://github.com/openshift/os/issues/117
 rm /etc/rc.d/init.d/network
 rm /etc/rc.d/rc*.d/*network
+
+# And readahead https://bugzilla.redhat.com/show_bug.cgi?id=1594984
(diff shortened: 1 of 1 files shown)


ashcrow:
Does this fix the upstream bug, or keep RHCOS from inheriting it (or both)?

cgwalters:
> Does this fix the upstream bug, or keep RHCOS from inheriting it (or both)?

Ah...upstream being...RHEL?  RHELAH?  systemd?  I personally have been thinking of RHCOS as upstream of RHELAH, i.e. we do things first here.  As far as systemd it's long since dead there.

jlebon:
Interesting, I hadn't heard of it before. It seems like something more geared towards desktops than servers. Given that it's already nuked upstream, I agree we should just drop it here as well.

/lgtm

ashcrow:
> Ah...upstream being...RHEL? RHELAH? systemd? 

Apologies, I meant does this PR fix the bugzilla bug linked, or is it just a reference to a problem that we are fixing in RHCOS.

cgwalters:
>  I meant does this PR fix the bugzilla bug linked,

Yep!  Next I'll "backport" to RHELAH.

Issue #78, opened 2018-06-05
Pull in (or rewrite) toolbox


cgwalters:
I think the *idea* of https://github.com/coreos/toolbox is really critical - even if the implementation is very basic. 

My strawman: Write a new tool that uses `podman` via the varlink API.  Have some opinionated defaults, like mounting `/srv:/srv` for passthrough, and also e.g. `-v /etc:/host/etc:ro -v /usr:/host/usr:ro` say.   But also just pass through some basic options like `--pid=host`.

In my dev containers I ended up doing `-v /home:/srv/container-home` so that I get persistent history.  There's a whole big decision point about using `--rm` by default here (which forces storing state elsewhere, IMO a good idea) or not.

cgwalters:
https://pagure.io/atomic-wg/issue/505 is related here as far as passing through `/srv` by default.

dustymabe:
should we migrate this to??  https://github.com/coreos/fedora-coreos-tracker/issues

dustymabe:
or maybe https://pagure.io/ContainerSIG/container-sig/issues

ashcrow:
IMHO it depends on ownership: If the CC&T wants to own it or if the ContainerSIG wants to own it.

ashcrow:
@yuqi-zhang is working on toolbox updates in https://github.com/yuqi-zhang/toolbox/tree/rhcos-toolbox

ashcrow:
PR: https://github.com/coreos/toolbox/pull/44

ashcrow:
Is there more work we need to do before closing this issue @yuqi-zhang @cgwalters?

yuqi-zhang:
@ashcrow see the discussions here: https://github.com/debarshiray/fedora-toolbox/issues/8

At some point I think we'd like to consolidate the two scripts. I'm testing it out and seeing what would be a good way to preserve existing workflow for *COS systems.

cgwalters:
I feel like we have the basic functionality here now, we can track further work in other issues.

PR #141, opened 2018-06-26
RFC: Add strace to the host

jlebon:
I was surprised to find that `strace` wasn't already included. For comparison, both Fedora Atomic Host and Container Linux ship with `strace` today. It's one of those tools that have a really nice value/size ratio. We're looking at 1.6M on disk for `/usr/bin/strace` + `libunwind` libs. This is related to #78, though in my opinion it's useful enough to be directly in the tree.

Thoughts?


cgwalters:
Sure, though I would really like to do the toolbox thing soon too...

/lgtm

PR #149, opened 2018-06-29
cloud.ks: Add $coreos_firstboot karg, create /boot/coreos-firstboot

cgwalters:
This is part of https://github.com/dustymabe/ignition-dracut/pull/12

### DIFF
--- a/cloud.ks
+++ b/cloud.ks
@@ -23,8 +23,8 @@ clearpart --initlabel --all
 #  - ip=dhcp           # how to get network
 #  - rd.neednet=1      # tell dracut we need network
 #  - enforcing=0       # ignition + selinux doesn't work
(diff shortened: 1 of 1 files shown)


jlebon:
This (and the two related PRs) look sane to me. It should be safe to merge these out of order, though let's just wait until the `ignition-dracut` one goes in. (OOC, I was looking at https://www.gnu.org/software/grub/manual/grub/grub.html#Shell_002dlike-scripting, and it's not actually specified whether the equivalent of `set -u` is in effect re. unset variables, though it'd be pretty dramatic to fail a boot because of that!).

ashcrow:
Previous item has merged. Marking :+1: per @jlebon's review.

PR #153, opened 2018-07-03
Dockerfile: Fix nginx root

cgwalters:
This broke when reworking things earlier in ffbc3285b757ebc10e71b7dced04172cbd964ab1

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ $ pivot -r $REGISTRY/os:latest
 OR
 
 ```
-$ docker run --network host -d -w /srv/tree/repo $REGISTRY/os:latest
+$ docker run --network host -d -w /srv/repo $REGISTRY/os:latest
 $ ostree remote add --no-gpg-verify oscontainer http://localhost:8080/repo
 $ rpm-ostree rebase -r oscontainer:openshift/3.10/x86_64/os
 ```
--- a/Dockerfile.rollup.in
+++ b/Dockerfile.rollup.in
@@ -5,6 +5,6 @@ ADD rollup.sh /root
 RUN env OSTREE_REPO_URL=@OSTREE_REPO_URL@ /root/rollup.sh && rm -f /root/rollup.sh
 # Keep this in sync with Dockerfile
 COPY nginx.conf /etc/nginx/nginx.conf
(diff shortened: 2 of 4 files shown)


nginx.conf
```diff
@@ -36,7 +36,7 @@ http {
         listen       8080 default_server;
         listen       [::]:8080 default_server;
         server_name  _;
-        root         /srv/repo;
+        root         /srv/tree/repo;
```
ashcrow: The old location looks correct to me reading https://github.com/openshift/os/commit/ffbc3285b757ebc10e71b7dced04172cbd964ab1#diff-3254677a7917c6c01f55212f86c57fbfR14

cgwalters: I didn't test this locally actually but...see the problem, we need to keep both container builds in sync.  It's tempting to neuter the base one until such time as we switch to a pure-container solution.

cgwalters:
OK, updated :arrow_up:

cgwalters:
(And tested locally too)

README.md
```diff
@@ -29,7 +29,7 @@ $ pivot -r $REGISTRY/os:latest
 OR
 
 ```
-$ docker run --network host -d -w /srv/tree/repo $REGISTRY/os:latest
+$ docker run --network host -d -w /srv/repo $REGISTRY/os:latest
 $ ostree remote add --no-gpg-verify oscontainer http://localhost:8080/repo
```
peterbaouoft: Hi, this line seems to have some inconsistency from the change in index.html?
`ostree remote add --no-gpg-verify os-svc http://os-content.namespace.svc:8080/`

The following(change in `index.html`) works for me while the other doesn't:
```
[root@localhost ~]# ostree remote add --no-gpg-verify os-test http://localhost:8080 
[root@localhost ~]# rpm-ostree rebase -r os-test:openshift/3.10/x86_64/os
1774 metadata, 12350 content objects fetched; 650605 KiB transferred in 217 seconds 
```

Did I miss something? Thanks.

**P.S**: I came across it while testing for https://github.com/openshift/os/issues/157.

jlebon: Yes, you're correct! This should just be `http://localhost:8080` now since the root dir for nginx is already in the repo.

PR #155, opened 2018-07-05
pipeline-utils: Group secret and not-secret values

cgwalters:
This helps keep clear what values are actually secrets, which
was helpful for me to understand things.


pipeline-utils.groovy
```diff
@@ -39,27 +33,35 @@ def define_properties(timer) {
                     required: true),
         credentials(name: 'AWS_CI_ACCOUNT',
                     credentialType: 'org.jenkinsci.plugins.plaincredentials.impl.StringCredentialsImpl',
-                    description: "OpenShift AWS CI root account number.",
+                    description: "OpenShift AWS bI root account number.",
```
dm0-: Was this intentional?

jlebon:
The division between secret and non-secret looks correct to me. LGTM with typo fix!

pipeline-utils.groovy
cgwalters: Fixed :arrow_down:

PR #158, opened 2018-07-06
cloud.ks: Add tty0 console output for ignition logging

Rahuls0720:
Many platforms, like QEMU, use tty0 for console output. This PR adds tty0 console output to allow ignition logging/debugging.

Jira card: https://projects.engineering.redhat.com/browse/COREOS-202

### DIFF
--- a/cloud.ks
+++ b/cloud.ks
@@ -24,7 +24,7 @@ clearpart --initlabel --all
 #  - rd.neednet=1      # tell dracut we need network
 #  - enforcing=0       # ignition + selinux doesn't work
 #  - $coreos_firstboot # This is actually a GRUB variable
-bootloader --timeout=1 --append="no_timer_check console=tty1 console=ttyS0,115200n8 net.ifnames=0 biosdevname=0 ip=dhcp rd.neednet=1 enforcing=0 $coreos_firstboot"
+bootloader --timeout=1 --append="no_timer_check console=ttyS0,115200n8 console=tty0 net.ifnames=0 biosdevname=0 ip=dhcp rd.neednet=1 enforcing=0 $coreos_firstboot"
 
 part /boot --size=300 --fstype="xfs"
 part pv.01 --grow


jlebon:
Have you confirmed this fixes the issue? My memory is a bit fuzzy on this, but I remember looking into this some time ago and found https://github.com/systemd/systemd/issues/3403. Basically, the order in which the `console=` arguments are given matters. I wonder if we also need `systemd.show_status=1` for this?

ashcrow:
@jlebon that's a good find. I believe @Rahuls0720 is attempting to build a local version of RHCOS with these changes. We chatted a bit about it yesterday but I wasn't able to help as much as I would have liked due to meetings.

@Rahuls0720 were you able to successfully create a local version of RHCOS with your change to test?

Rahuls0720:
I was able to run it locally using vagrant, but Ignition logs output with both tty0 and tty1. So the card doesn't apply to vagrant. I'm trying to get a copy of rhcos.qcow2 to run qemu but I don't have access to the build servers. 

https://coreos.slack.com/archives/C043AL8FR/p1531237535000018

crawford:
@Rahuls0720 were you looking at the VGA output or serial output when you tested vagrant? Which virtualization backend were you using with vagrant?

Rahuls0720:
VGA output. I used virsh to view the output.

ashcrow:
I spoke with @crawford today and it sounds like we should go with setting to ttys0, tty0, and tty1.

@crawford Agree?

Rahuls0720:
Does anyone know how to log the tty0 console output for either qemu or vagrant? I'm looking for something like -nographic or -serial file:FILENAME qemu flags that allows me to view and log the tty0 output (to ensure ignition ran).

@crawford mentioned putting an invalid url in the ignition config, but the boot up halts at an emergency shell
screenshot from 2018-07-18 14-04-26

crawford:
@Rahuls0720 Use an Ignition config that references something from an address that is unreachable. Something like this should be interesting:

```json
{
  "ignition": {
    "version": "2.2.0",
    "config": {
      "append": [{
        "source": "http://localhost"
      }]
    }
  }
}
```

I haven't validated that or tested it.

arithx:
In kola we use `-chardev file,id=log,path=<path_to_log>` to capture the serial console on QEMU (https://github.com/coreos/mantle/blob/master/platform/machine/qemu/cluster.go#L189) but I'm not sure about specifically capturing the `tty0`

Rahuls0720:
It boots fine for me when I added ^ to my config. I validated it also. My config looks like:

{"ignition":{"config":{
"append": [{
        "source": "http://localhost"
      }]
},"security":{"tls":{}},"timeouts":{},"version":"2.2.0"},"networkd":{},"passwd":{"users":[{"groups":["sudo"],"name":"core","sshAuthorizedKeys":["..."]}]},"storage":{},"systemd":{}}

Rahuls0720:
@jlebon Do you know the reason tty1 is being used?

Also, the last console set in the kernel command line arg is the one systemd logs everything to (this console has the C flag in /proc/consoles). The other consoles only get the kernel logs. If tty1 is needed,  console=tty1 console=ttyS0,115200n8 console=tty0 should do the trick. Otherwise console=ttyS0,115200n8 console=tty0.

jlebon:
@Rahuls0720 It's just inherited from the Fedora kickstart ours is based on.

So, reading https://github.com/systemd/systemd/issues/3403 again, it sounds like there's no way right now to get systemd messages to go to all consoles without Plymouth? If that's the case, then I'd vote for `ttyS0` rather than `tty0` being the last argument. E.g. cloud providers make it easy to access those logs and locally we can use `virsh console` (which I've found super useful exactly wrt. debugging ignition).

Rahuls0720:
I'm a bit confused on Plymouth. Plymouth is part of the initramfs and RHCOS doesn't include Plymouth in its initramfs. So why do boot up messages still appear on both ttyS0 and tty0?

jlebon:
Right, `virsh console` connects to the serial console. Since it's already set as the last `console=` argument, systemd messages show up there. I thought @crawford wanted to make them appear on both of them. Though reading the original Jira ticket now, it looks like the goal was to see journal output such as ignition, which definitely works today with `virsh console`.

I also tried running QEMU directly and got ignition log messages using e.g. `-chardev file,id=log,path=/tmp/log.txt -serial chardev:log` like kola does. @crawford Does that not work for you?

ashcrow:
Related https://github.com/coreos/bugs/issues/2136

Rahuls0720:
@crawford @ashcrow @jlebon I believe this should solve the problem. 

Just to ensure everyone is on the same page, I replaced tty1 with tty0 and moved tty0 as the last console in the kernel args. Tty0 should look identical to ttyS0 with the addition to ignition logs (on tty0). The journald.conf option 'forwardtoconsole' is part of RHCOS, but doesn't work with ignition. Please see  https://github.com/coreos/bugs/issues/2136 for details about this ignition bug.

crawford:
@jlebon I was watching the VGA console (tty0).

@Rahuls0720 We need to use ForwardToKMsg so it shows up on both.

Rahuls0720:
lgtm

cgwalters:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/OI7BOWZSXDJM5W2KVDK3BBEN6S7ASLZC/

PR #159, opened 2018-07-09
repo: use local rpm-gpg assets

lucab:
This avoids updating and sourcing rpm-gpg assets from the global PKI
path, using local resources shipped in the repository instead.


lucab:
Ah, this PR was likely a bit too quick. Metadata fetching works fine, but then it fails after downloads with:
```
error: package bash-completion-1:2.1-6.el7.noarch cannot be verified and repo rhel-7.5-server is GPG enabled: failed to lookup digest in keyring for /proc/self/fd/10/cache/repomd/rhel-7.5-server-rpmostree-unset-releasever-x86_64/packages/bash-completion-2.1-6.el7.noarch.rpm
```

cgwalters:
Yeah...unfortunately I think that's only going to work today if `rpm-ostree compose tree` is running with the repo checkout as its current dir.  It'd definitely be nice to support `.repo`-file relative paths though.  

https://github.com/rpm-software-management/libdnf/issues/43 is related.

lucab:
I'm inclined to close this for the moment. I think medium-term we can either:
 * standardize the build-container, and rely on absolute paths in there
 * standardize the interactive/local build process, and rely on specific relative arrangements of paths
 * both of those (this is what CL does with the SDK and scripts library)

For the short term, I'll just add a note that GPG keys needs to be propagated at specific absolute paths in the build environment.

miabbott:
Since were using the `distribution-gpg-keys` from #245, I don't think this PR is valid anymore.

Can we close it?

lucab:
@miabbott yep, done.

Issue #147, opened 2018-06-28
re-organize this repo


dustymabe:
There are a ton of files in the top level and they are for various purposes. It would be nice if we could group things together in subfolders and have a structure that was more user friendly.

Thoughts?

ashcrow:
I agree with this. Not sure if now is literally the right time to do it, but it does need to happen. We may even want to split out into a few repos.

dustymabe:
> We may even want to split out into a few repos.

monolith for now - it's good to know you've got most of the pieces in this single bag (for now)

cgwalters:
This is `2` in https://github.com/openshift/os/issues/143 btw

cgwalters:
Calling this one fixed.

PR #160, opened 2018-07-09
Move rdgo bits to rdgo/ subdirectory

cgwalters:
Part of: https://github.com/openshift/os/issues/147


cgwalters:
Tide is a bit broken right now, merging manually.

PR #161, opened 2018-07-09
Makefile: Use download.fp.org for mirrors

jlebon:
This will automatically redirect to a close mirror.


jlebon:
/hold
I'm just gonna tweak the docs as well in this one.

cgwalters:
The downside is going back to mirror roulette :game_die: though.  But it's fine by me.

jlebon:
OK, dumped some more info in here!
/hold cancel

cgwalters:
Merging until we're out of low tide, so to speak.

PR #162, opened 2018-07-09
fedora-rhcos: Initial location to help kickstart further work

ashcrow:
Meant to help kick start RHCOS using Fedora content POC.

Items that can easily be reused are soft-linked in to the new sub-directory.


fedora-rhcos/overlay.yml
```diff
@@ -0,0 +1,43 @@
...
+#       patches: drop
+#       branch: master
+
+  # Required as a cri-o dependency
+  - src: https://github.com/containernetworking/plugins.git
+    distgit:
+        name: containernetworking-cni
+        patches: drop
+
+  - src: github:openshift/redhat-release-coreos
```
cgwalters: I think we'll need to fork this one.

fedora-rhcos/rhcos.tdl
```diff
@@ -0,0 +1,16 @@
+<template>
```
cgwalters: This one can likely be a symlink too.

fedora-rhcos/host.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - root
+ignore-removed-groups:
+  - root
+check-passwd:
+  type: "file"
+  filename: "passwd"
+check-groups:
+  type: "file"
+  filename: "group"
+packages:
```
cgwalters: Let's add a big comment here like:
```
# The canonical version of this is in host.yaml - edit that first!
```

ashcrow:
Updating shortly ... thanks!

fedora-rhcos/host.yaml
```diff
@@ -0,0 +1,95 @@
...
+documentation: false
+# Doesn't quite work out of the box yet in RHEL7,
+# but see https://pagure.io/fedora-atomic/pull-request/62/edit
+tmp-is-dir: false
+initramfs-args:
+  - "--no-hostonly"
+  - "--add"
+  - "iscsi"
+automatic_version_prefix: "3.10-7.5"
+mutate-os-release: "7"
```
jlebon: I guess these should be something like...

```yaml
automatic_version_prefix: "3.10-28"
mutate-os-release: "28"
```

?

jlebon:
> I'm going back and forth on whether this should be a new repo or not...leaning a bit towards the latter. If we do that, one idea is to have rhcos inherit it as a git submodule or something? The main idea would be that if FCOS adds a package we should inherit that by default say.

I might be misunderstanding here. This PR seems to be about making RHCOS based on Fedora content, right? Or is this actually to start FCOS? I reviewed it under the assumption it was the former. If the latter, I definitely agree we should make it a separate repo!

Once we have FCOS going though, I do like the idea of having RHCOS just doing `include: fcos/host.yaml`, where `fcos/` is a submodule.

fedora-rhcos/host.yaml
```diff
@@ -0,0 +1,95 @@
...
+  type: "file"
+  filename: "group"
+packages:
+ # Kernel + systemd
+ - kernel irqbalance microcode_ctl systemd
+ # Release package
+ - redhat-release-coreos
+ # rpm-ostree
+ - rpm-ostree nss-altfiles
+ # Subscriptions
```
miabbott: Probably could drop these `subscription-manager` packages for FCOS

fedora-rhcos/host.yaml
```diff
@@ -0,0 +1,95 @@
...
+  type: "file"
+  filename: "passwd"
+check-groups:
+  type: "file"
+  filename: "group"
+packages:
+ # Kernel + systemd
+ - kernel irqbalance microcode_ctl systemd
+ # Release package
+ - redhat-release-coreos
```
miabbott: We'll likely need a `fedora-release-coreos`

fedora-rhcos/host.yaml
```diff
@@ -0,0 +1,95 @@
+ref: "openshift/3.10/x86_64/os"
```
miabbott: Do we want to differentiate the `ref` between RHCOS and FCOS?

ashcrow:
> I might be misunderstanding here. This PR seems to be about making RHCOS based on Fedora content, right?

Correct.

> Or is this actually to start FCOS? I reviewed it under the assumption it was the former. If the latter, I definitely agree we should make it a separate repo!

It's not to start FCOS, though I'm sure there is overlap.

> Once we have FCOS going though, I do like the idea of having RHCOS just doing include: fcos/host.yaml, where fcos/ is a submodule.

That makes sense to me as well. However, at the time of this writing, we've been asked to have a Fedora version of RHCOS as part of this milestone.

fedora-rhcos/host.yaml
ashcrow: Keep in mind this is still RHCOS.

fedora-rhcos/host.yaml
ashcrow: Added

fedora-rhcos/rhcos.tdl
ashcrow: Done

fedora-rhcos/host.yaml
miabbott: Oops...that invalidates most of my comments  ¯\_(ツ)_/¯

cgwalters:
We're going to need naming conventions...I propose `rhcos-el7` for the current stream, and `rhcos-f28` for what's being discussed here.  I think this is better than `fedora-rhcos` because the "rhcos" aspect basically has higher "priority".  

One thing that's related to this is I think we should consider *requiring* the pivot...i.e. we only ship BootstrapOS.  If we go that approach, then a natural thing to do is to start with having `rhcos-f28` just be BootstrapOS for now.  (Shorthand: `rhbootcoreos-f28`?)

cgwalters:
I'd like to avoid duplicating the package lists; we should be able to do that using the treefile `include:` feature.

ashcrow:
To follow up a little here, Dusty and I spoke with Ben and we are pushing this goal out in favor of other goals. We still want to do this but having two pipelines for RHCOS isn't an activy priority right now.

PR #163, opened 2018-07-11
jenkins/treecompose: Maintain content in container

cgwalters:
Rather than having a "traditional" OSTree repository stored externally,
maintain it canonically inside a container image.  The "oscontainer/rollup"
task then goes away.

The advantages of this are that we now only have the OSTree repository
in exactly one place, and can e.g. consistently mirror it or promote it.

This PR prototypes out doing "podman in privileged containers"; I struggled
a bit with trying to use the Jenkins/docker APIs, and I think this ends
up being better.


cgwalters:
WIP since the cloud images need to be changed to pull down the container and run the webserver locally, then pull content from that.

cgwalters:
Or alternatively, we do two things at once, and change the cloud images to be "bootstrapOS".

cgwalters:
[Example run](https://url.corp.redhat.com/453fe29)

jlebon:
OK, so this is about taking it one step further and using the container as the canonical source even internally for our builds. Interesting idea! So then, I suppose doing a promotion would consist of pushing to a different tag?

> Or alternatively, we do two things at once, and change the cloud images to be "bootstrapOS".

It sounds like this is still up for discussions. Maybe in the meantime we can just keep using the externally hosted OSTree too, and just think of it as a secondary artifact? IOW, we no longer do an "Rsync In", but still do an "Rsync Out".

It's also useful to have it hosted on a traditional server internally to easily e.g. `pull --subpath` or pull the commit object and `rpm-ostree db list`.

cgwalters:
> So then, I suppose doing a promotion would consist of pushing to a different tag?

Yep, exactly.

> IOW, we no longer do an "Rsync In", but still do an "Rsync Out".

I think we need to do that short term.  Long term though...I don't think we're gaining the benefit unless we only do it *one* way (this is the same argument I make with rojig).

cgwalters:
OK I updated this :arrow_up:

 - Readded rsync out
 - Uncommented "for testing" hacks like dropping the timer and commenting out triggers

Jenkinsfile.treecompose
```diff
@@ -73,24 +88,45 @@ node(NODE) {
...
-                    --rsync-opt=--stats --rsync-opt=-e \
-                    --rsync-opt='ssh -i ${KEY_FILE} -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no'
-                """
-            }
-        }
+      // This takes OSTree repo and copies it out of the original
+      // container into our working directory, which then gets put
+      // into a new container image.  This is slow, but it's hard to do better
+      // with Docker/OCI images.  Ideally there'd be a way to "bind mount" content
+      // into the build environment rather than copying.
```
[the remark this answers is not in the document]
cgwalters: SELinux policy denies it.  Simplest would be adding `buildah bud --privileged` or so, I may look at that.

cgwalters:
OK, reworked for comments :arrow_up:

Jenkinsfile.treecompose
```diff
@@ -57,13 +70,19 @@ node(NODE) {
...
+            ostree --repo=${repo} rev-parse ${ref} > commit.txt
+        """
             def commit = readFile('commit.txt').trim();
             version = utils.get_rev_version("${repo}", commit)
             currentBuild.description = "🆕 ${version} (commit ${commit})";
-            sh "rpm-ostree --repo=${repo} db diff ${commit}^ ${commit}"
+            if (previous_commit.length() > 0) {
+                sh "rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit}"
+            }
+        sh "ostree summary --repo=${repo} --update"
```
jlebon: Minor: indentation

Jenkinsfile.treecompose
```diff
@@ -57,13 +70,19 @@ node(NODE) {
             return
         }
 
-        stage("Compose Tree") {
-            sh "coreos-assembler --repo=${repo} ${manifest}"
-            sh "ostree --repo=${repo} rev-parse ${ref} > commit.txt"
+        // Note that we don't keep history in the ostree repo
```
jlebon: Don't we still need to `ostree prune` though to drop out old objects from the previous commit?

cgwalters: Yep, fixed thanks :arrow_down:

PR #167, opened 2018-07-12
jenkins/treecompose: Fix prune ordering

cgwalters:
I added it in review but didn't test it.  It needs to be after
the diff, otherwise we'll have pruned what we try to diff.
(Perhaps we should go all the way here and deploy a service
 that maintains the diffs/history of the oscontainer images)


jlebon:
Ahh of course!
/lgtm

PR #169, opened 2018-07-12
compose-post.sh: Enable crio by default

ashcrow:
See https://github.com/coreos/mantle/pull/885#issuecomment-404637101

This seemed like a safer place to do an `enable` compared to `cloud.ks` but let me know if it should move.


jlebon:
This probably belongs more in https://github.com/openshift/redhat-release-coreos/blob/master/42-coreos.preset?

ashcrow:
@jlebon updating

PR #171, opened 2018-07-13
jenkins/treecompose: Fix rsync destination

cgwalters:
Should ensure the cloud images are still picking up changes.


miabbott:
/lgtm

Should fix the `rsync` failure

PR #172, opened 2018-07-13
jenkins/cloud,rdgo: Sync cloud images and rdgo into rhcos bucket

cgwalters:
Move content into the `rhcos` bucket I made which should be separate
from the CI one.  This ensures we have backups in a better way.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -224,12 +224,24 @@ node(NODE) {
         }
     }
 
-    stage("Sync Out") {
+    par_stages["rsync-out"] = { -> stage("rsync out") {
         withCredentials([
             string(credentialsId: params.ARTIFACT_SERVER, variable: 'ARTIFACT_SERVER'),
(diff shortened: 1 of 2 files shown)


Jenkinsfile.cloud
```diff
@@ -224,12 +224,24 @@ node(NODE) {
...
+    par_stages["rsync-out"] = { -> stage("rsync out") {
         withCredentials([
             string(credentialsId: params.ARTIFACT_SERVER, variable: 'ARTIFACT_SERVER'),
             sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
         ]) {
             utils.rsync_dir_out(ARTIFACT_SERVER, KEY_FILE, images)
         }
-    }
+    } }
+    par_stages("s3-out") = -> { stage("s3 upload") {
```
miabbott: Should this be `par_stages["s3-out"]`?

cgwalters: It's not required but yeah, let's keep them in sync.  (It's sort of annoying that `parallel` wants a map and not just an array/set), and we have to just make up a key.

PR #176, opened 2018-07-13
host: Install all languages

cgwalters:
I was going to do some treefile changes, ended up doing
https://github.com/projectatomic/rpm-ostree/pull/1459
which turned up the bug in this repo.

Let's actually fix it by just installing all languages.
I'd actually like to do man pages too, but that will come separately
after discussion.

### DIFF
--- a/host.yaml
+++ b/host.yaml
@@ -8,7 +8,7 @@ repos:
   - rhel-7.5-server-extras
   - rhel-7.5-atomic
   - dustymabe-ignition
-install_langs:
+install-langs:
  - en_US
 documentation: false
 # Doesn't quite work out of the box yet in RHEL7,


ashcrow:
This PR makes sense to me. Though if we are not expecting admins to have to ssh in to the system why add manpages?

lucab:
What's the reasoning/requirement behind this PR? I think that following CL setup here would be good. For reference:
```
$ locale -a

C
C.utf8
POSIX
```

(unless I completely missed the scope of this PR, in which case please just dismiss)

cgwalters:
Updated :new: :arrow_up: to keep the status quo (I would like to address the docs issue in a nicer way, which is related to languages, but that's clearly a separate thing).

PR #184, opened 2018-07-23
ostree: explicitly set a version metadata string on the commit

miabbott:
We weren't incrementing the version number each time we made a commit, which could be confusing to a user who wasn't tracking commit IDs.

Additionally, I believe not changing the version was causing us to fail [this version string check](https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L76) in the cloud pipeline.  (So at the time of the writing, our latest AMI (`ami-02323e873420d8eab`) is from July 13)

I chose to set the version to the previous version prefix plus the build number from Jenkins.

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -12,6 +12,7 @@ def artifact_repo = "/srv/rhcos/output/repo"
 
 def manifest = "host.yaml"
 def ref = "openshift/3.10/x86_64/os";
+def version_prefix = "3.10-7.5"
 
 node(NODE) {
     checkout scm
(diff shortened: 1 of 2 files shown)


cgwalters:
Ah heh, I was just debugging this now too and you beat me to it!  Nice.  So...I think my vote is to entirely purge references to the ostree version number; it's redundant with the container image versioning.  Instead of a version comparison in the cloud build check, we should use the ostree checksum.

miabbott:
I think there is some value in retaining a human-friendly version number in the metadata.  Those (hopefully) infrequent times when folks need to collaboratively debug, it is easier to say (and remember) "I'm on version X.Y.Z"   

But I agree we should use commit IDs in the pipeline for comparison operations.

cgwalters:
> I think there is some value in retaining a human-friendly version number in the metadata. 

Definitely!  To be clear...I'm more arguing that version number should live in the container image (or perhaps be in sync).

Anyways let's not the let the perfect be the enemy of the good....

/lgtm

PR #185, opened 2018-07-23
cloud: stop using version strings for comparison

miabbott:
This changes some of the comparison logic to use commit IDs instead of
version strings; more precision == better!


miabbott:
Note:  not tested in the pipeline

cgwalters:
Nice!  Let's try it.

/lgtm

PR #186, opened 2018-07-23
treecompose: define the ostree_version variable

miabbott:
Need to define during assignment or bad things happen.


miabbott:
I really need to setup a test pipeline to make sure I stop breaking things...

cgwalters:
This needed at least https://github.com/openshift/os/commit/4a194b0c06a9a8b509f19d5f99f0700b82ede47c

(Don't hesitate to go to a less-heavyweight process of pastebin when we're in this iterate-on-fixes process)

PR #187, opened 2018-07-24
oscontainer: workaround podman label bug

miabbott:
When building a container image via `podman` (which actually uses
`buildah`), key=value pairs split across multiple lines under a
single LABEL instruction are getting mangled.  This misbehavior
is captured in projectatomic/buildah#879.

Until that is fixed, use a single LABEL per key=value pair.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,8 +12,10 @@ RUN rpm -q rpm-ostree && rpm-ostree --version && \
 FROM registry.centos.org/centos/centos:7
 ARG OS_VERSION="3.10-7.5"
 ARG OS_COMMIT="null"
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,8 +12,10 @@ RUN rpm -q rpm-ostree && rpm-ostree --version && \
 FROM registry.centos.org/centos/centos:7
 ARG OS_VERSION="3.10-7.5"
 ARG OS_COMMIT="null"
-LABEL io.openshift.os-version = "$OS_VERSION" \
-      io.openshift.os-commit = "$OS_COMMIT"
+# Use separate LABEL instructions until podman/buildah is fixed
```
ashcrow: Thank you for noting why!

PR #189, opened 2018-07-25
oscontainer: remove spaces from image labels

miabbott:
The Dockerfile reference states that LABELs need to have explicit
'key=value' pairs.  Previously, we had 'key = value' and the extra
spaces were causing the labels to be incorrect.

With this change, we are also able to use a single LABEL instruction,
with the 'key=value' pairs split across multiple lines.

Whitespace matters, folks!


miabbott:
Hopefully this is the last commit around these `LABELs` for a while

PR #190, opened 2018-07-26
jenkins/cloud: Rework to use coreos-assembler's virt-install

cgwalters:
Requires: https://github.com/cgwalters/coreos-assembler/pull/9

There are a lot of things going on here.  The *primary* goal though
is to drop the ImageFactory dependency because it's annoying to
containerize.  It fundamentally clashes with the way we were driving
it via Jenkins; our job would just find the first image in
`/var/lib/imagefactory`, meaning we couldn't do concurrent builds.

I spend my day-to-day development in a dev container, and while I
do leak in the libvirt socket, ImageFactory still has other problems
with that.

There were some unexpected hurdles to getting this to work:

 - When testing interactively, it went fine, but then I had to
   rework our console to log to a file for Jenkins with no tty
 - In Jenkins there were races/bugs due to libvirt (session)
   relying on the qemu process being reaped to determine shutdown.
   I tried to use `docker run --init` but ended up using bwrap
   in the container.
 - Kernel panics on the jenkins worker running Fedora 25/28 - I
   provisioned a RHEL7 worker and things went fine.

Overall though I think this is a lot better.


jlebon:
Let's make sure the coreos-assembler PR gets in first.
/hold

Jenkinsfile.cloud
```diff
@@ -83,27 +72,33 @@ node(NODE) {
...
+    // Note this runs unprivileged to avoid the virt stack trying to use privileges
+    // We run in bwrap which sets up a pid namespace +init to ensure qemu is reaped,
+    // which libvirt relies on.
+    try {
+    withCredentials([
+        string(credentialsId: params.INSTALLER_TREE_URL, variable: 'INSTALLER_TREE_URL')]) {
+    stage("Running coreos-virt-install") { sh """
+        chmod a+rw /dev/kvm
+        setfacl -m u:builder:rwX ${WORKSPACE}
+        bwrap --bind / / --dev-bind /dev /dev --proc /proc --unshare-pid runuser -u builder -- coreos-virt-install --dest /tmp/rhcos.qcow2 --create-disk \
```
jlebon: For consistency, let's define `image = "/tmp/rhcos.qcow2"` in this step rather than in the next one and pass it to `coreos-virt-install`?

Jenkinsfile.cloud
```diff
@@ -163,15 +156,13 @@ node(NODE) {
         // we're stuck with an older host because imagefactory produces kernel
         // panics on newer hosts, so let's use a container to have access to a
         // newer rpm-ostree that understands the new pkglist metadata
```
jlebon: Let's also nuke that outdated comment?

cloud.ks
```diff
@@ -33,8 +35,7 @@ logvol / --size=3000 --fstype="xfs" --name=root --vgname=coreos
 
 ostreesetup --nogpg --osname=rhcos --remote=rhcos --url=@@OSTREE_INSTALL_URL@@ --ref=@@OSTREE_INSTALL_REF@@
 
-
-reboot
+poweroff
```
jlebon: Minor: this is fine, though we shouldn't actually need this if we have `--noreboot`, right?

Jenkinsfile.cloud
```diff
@@ -1,7 +1,7 @@
 def TIMER = "H H/3 * * *"
-def NODE = "rhcos-imgfac-jslave"
-def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler"
-def DOCKER_ARGS = "--net=host -v /srv:/srv --privileged"
+def NODE = "rhcos-jenkins"
+def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler:testing"
```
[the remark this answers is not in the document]
cgwalters: Yes, we need to change that before merging, but that's where I was pushing the in-progress PRs.

cloud.ks
cgwalters: I added this while debugging `virt-install`.  It polls the domain being active, which I thought was being racy, but the way it actually works is that it tells libvirt to shutdown the domain on reboot, so it doesn't matter what we do here.

Jenkinsfile.cloud
```diff
@@ -83,27 +72,33 @@ node(NODE) {
...
-    stage("Running imagefactory") {
-        sh """
-            imagefactory --debug base_image \
-                --file-parameter install_script cloud.ks \
-                --parameter offline_icicle True \
-                rhcos.tdl
+    // Note this runs unprivileged to avoid the virt stack trying to use privileges
+    // We run in bwrap which sets up a pid namespace +init to ensure qemu is reaped,
+    // which libvirt relies on.
+    image = "/tmp/rhcos.qcow2"
```
miabbott: Does this need a `def image` before `stage("Prepare Configs")`?  It looks like it is used in the `Postprocessing` stage, too

miabbott: Oh whoops...I see this is outside of the `Prepare Configs` stage.  Still  needs to be `def`'ed here, though?

Jenkinsfile.cloud
```diff
@@ -83,27 +72,33 @@ node(NODE) {
...
         }
     }
 
-    stage("Running imagefactory") {
-        sh """
-            imagefactory --debug base_image \
-                --file-parameter install_script cloud.ks \
-                --parameter offline_icicle True \
-                rhcos.tdl
+    // Note this runs unprivileged to avoid the virt stack trying to use privileges
```
ashcrow: Thank you for the explanation in comments!

ashcrow:
> Overall though I think this is a lot better.

Reading through the flow I tend to agree. Thank you for doing this @cgwalters!

jlebon:
Let's see if we can do this in a single comment...

/hold cancel
/lgtm

PR #194, opened 2018-08-02
Set version to 4

cgwalters:
The 7.5 is not very useful today.  I'm tentatively thinking that
we represent a new base with a different image stream entirely; probably
rename our current image too.  Something like:

`openshift/os-maipo:4.1249`
and
`openshift/os-2018:4.1249`
or so.


jlebon:
Should we match the OCP version more tightly? E.g. `4.0.$BUILD_ID` or `4.0-$BUILD_ID`?

cgwalters:
> Should we match the OCP version more tightly? E.g. 4.0.$BUILD_ID

Sure, done :arrow_up:

PR #197, opened 2018-08-02
pipelines: skip the repo checkout

miabbott:
Jenkins will checkout the `openshift/os` repo into its workspace
before any of the stages are started.  The workspace is then mounted
into the container to be used by the stages.  We can skip checking out
the repo a second time, as it is automatically made available inside
the container.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -18,7 +18,6 @@ node(NODE) {
        step([$class: 'WsCleanup'])
     }
 
-    checkout scm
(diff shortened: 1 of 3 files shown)


cgwalters:
And here me having the submitter go back and make bigger changes that are hard to test caused fallout - in the cloud job we were using the workspace cleanup, and that was cleaning up the scm.

I have a fix for this in https://github.com/openshift/os/pull/203 that I'll split out.

cgwalters:
PR in https://github.com/openshift/os/pull/204

PR #200, opened 2018-08-03
Make /tmp be tmpfs by default

cgwalters:
Drops cruft from the cloud kickstart, hence unifying across
metal and cloud.  Also, this is a key prerequisite for having
`/sysroot` be read-only by default.

I finally figured out the problem I had with this before; systemd
in RHEL7 is missing the `tmp.mount` unit being enabled by default.


jlebon:
I just fixed up the commit to fix the Pagure link.
/lgtm

PR #202, opened 2018-08-03
jenkins/treecompose: Rename current oscontainer to `os-maipo`

cgwalters:
We know in the future we're going to introduce a fundamentally
much newer version of the OS.  Let's prepare for that by renaming
our existing container to have an "OS identifier".

The goal in avoiding an explicit OS version number here is to make clear
that we may have different content than an existing OS; but
at the same time we want to retain *some* linkage to that OS.
Hence, I chose to use the "codename".


cgwalters:
This required a `skopeo copy docker://registry.svc.ci.openshift.org/rhcos/os:latest docker://registry.svc.ci.openshift.org/rhcos/os-maipo:latest`.

PR #205, opened 2018-08-06
postprocess: Unify /usr/sbin and /usr/bin

cgwalters:
It will help kola.  But it also follows on from the general
`/usr` unification model - basically because we can.  Arch Linux
at least has this today.

### DIFF
--- a/compose-post.sh
+++ b/compose-post.sh
@@ -2,6 +2,44 @@
 
 set -xe
 
+# bin+sbin unification; this is the case on e.g. Arch
+# today, and it helps the kola tool which does direct
+# SSH; don't need to worry about "is /sbin in root's path".
+#
+# First, handle any sbin -> bin symlinks, including
+# ones which cross names, like /usr/sbin/ping6 -> /usr/bin/ping
+for sbin in /usr/sbin/*; do
+    if ! test -L ${sbin}; then
+        continue
+    fi
+    bn=$(basename ${sbin})
+    bin=/usr/bin/${bn}
+    if ! [ -e "${bin}" ]; then
+        continue
+    fi
+    sbin_real=$(realpath ${sbin})
+    bin_real=$(realpath ${bin})
+    if [ "${bin_real}" = "${sbin_real}" ]; then
+        echo "sbin -> bin: ${bin_real} ${sbin_real}"
+        rm -f ${sbin}
+    fi
+done
+# Now walk over all files in sbin and move them,
+# this time handling any bin -> sbin links
+for sbin in /usr/sbin/*; do
+    bn=$(basename ${sbin})
+    bin=/usr/bin/${bn}
+    if test -L ${bin}; then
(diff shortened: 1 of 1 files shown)


ashcrow:
I'm not sure how I feel about unifying `/usr/bin/` and `/usr/sbin/`. I've seen unification efforts recently (such as doing the same thing with `/bin/` and `/usr/bin/`). Other than the `$PATH` issue hit by `kola` what do we get from the unification?

cgwalters:
The [UsrMerge](https://www.freedesktop.org/wiki/Software/systemd/TheCaseForTheUsrMerge/) was already done in RHEL7 which is over 4 years ago now - ostree was designed around it (although it's not strictly required).

kola isn't the only one to get hit by the randomness of where different distributions put binaries in `/sbin` vs `/bin` - e.g. just recently I wrote [this script](https://github.com/cgwalters/playground/blob/master/coreos-run-qemu) which I based on [this Arch wiki entry](https://wiki.archlinux.org/index.php/getty#Automatic_login_to_virtual_console) and hit on the fact that it's `/usr/bin/getty` (or `/usr/sbin/getty` on Arch), but on Fedora-derived systems it must be `/usr/sbin/getty`.

In general - as far as I'm aware the only reason we didn't do it in Fedora is no one has yet fully killed off `usermode`, though it's been close to dead for years.

ashcrow:
OK, that's fair. I'm OK with this. Thanks for the background @cgwalters!

dustymabe:
A few questions:

- how does this affect `rpm -qV`? does it still work?
- What happens if someone package layers an rpm that puts contents in either `/usr/sbin` or `/usr/bin` or both? 
- Does this trip up `rpm -qF`?

cgwalters:
> how does this affect rpm -qV? does it still work?

Remember, you generally shouldn't be using that; instead use `ostree fsck`.  (Though they aren't quite the same)

> What happens if someone package layers an rpm that puts contents in either /usr/sbin or /usr/bin or both?

Cases like `usermode` which tries to put the same binary in both will fail; otherwise it'll work.

> Does this trip up rpm -qF?

Yep, it breaks it.

dustymabe:
> Remember, you generally shouldn't be using that; instead use ostree fsck. (Though they aren't quite the same)

I typically do both. It's a warm fuzzy from the old world and a warm fuzzy from the new world.

cgwalters:
>    Does this trip up rpm -qF?

Actually it seems to follow the symlink, in some quick testing.

jlebon:
This is a pretty significant change.

> In general - as far as I'm aware the only reason we didn't do it in Fedora is no one has yet fully killed off `usermode`, though it's been close to dead for years.

Can we confirm this before doing merging this? Is this something Fedora is planning to do eventually? Are there any other potential issues holding this back in Fedora that might be relevant to us?

ashcrow:
This work is no longer relevant for this repo.

PR #207, opened 2018-08-06
Fix rdgo location

cgwalters:
We have to neuter the Dockerfile as we're not syncing that repo
out yet, and I don't remember where the code is for that.
This inside-and-outside problem is really awful.

### DIFF
--- a/rhcos-continuous.repo
+++ b/rhcos-continuous.repo
@@ -1,5 +1,5 @@
 [rhcos-continuous]
-name=Bleeding edge of packages relevant to RHCOS
-baseurl=https://s3.amazonaws.com/aos-ci/rhcos/rdgo
+name=Packages built by this pipeline
+baseurl=http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/rdgo/build/
 enabled=1
 gpgcheck=0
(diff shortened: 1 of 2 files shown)


rhcos-continuous.repo
```diff
@@ -1,5 +1,5 @@
 [rhcos-continuous]
-name=Bleeding edge of packages relevant to RHCOS
-baseurl=https://s3.amazonaws.com/aos-ci/rhcos/rdgo
+name=Packages built by this pipeline
+baseurl=http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/rdgo/build/
```
Rahuls0720: @cgwalters How/where is baseurl used?  I'm trying to wrap my head around the build pipeline :)

miabbott: This `.repo` file is used as part of the treecompose pipeline.  It provides the RPMs built by the `rdgo` pipeline.

It was changed to this internal location because the S3 bucket where the RPMs were originally getting synced to was private.  There is no good way (or any way?) to provide AWS credentials to `rpm-ostree compose`, so we had to fall back to the internal repo.

Rahuls0720: Thanks!

PR #208, opened 2018-08-07
jenkins: Define a common prelude

cgwalters:
This effectively reverts https://github.com/openshift/os/pull/197
and goes with a common prelude where we clean the workspace and
use the SCM defined by the pipeline.


cgwalters:
> and use the SCM defined by the pipeline.

(Otherwise it breaks to test the pipeline)

PR #209, opened 2018-08-07
jenkins: Add concept of "host workspace"

cgwalters:
For treecompose right now, but I was planning to change the cloud
task to pull the container too, so let's add an API.

We need a "host workspace" because $WORKSPACE may be on overlayfs,
and one can't do overlayfs on overlayfs.

This solves a few problems.  First, we could easily "leak" content
across task invocations; say that a job created a file one time and
then stopped.  Cleaning is a good idea for the same reason workspace
cleanup is.

Second, hardcoded paths on the host clash badly with testing
the pipeline; we don't want my `cgwalters-test-pipeline` to potentially
stomp on data.

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -5,11 +5,6 @@ def API_CI_REGISTRY = "registry.svc.ci.openshift.org"
 def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-maipo:latest"
 def DOCKER_ARGS = "--net=host -v /srv:/srv -v /run/docker.sock:/run/docker.sock --privileged"
 
-def treecompose_workdir = "/srv/rhcos/treecompose"
-def repo = "${treecompose_workdir}/repo"
-// We write to this one for now
-def artifact_repo = "/srv/rhcos/output/repo"
-
 def manifest = "host.yaml"
 // TODO - stop using the ref in favor of oscontainer:// pivot
 def ref = "openshift/3.10/x86_64/os";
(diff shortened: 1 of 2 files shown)


Jenkinsfile.treecompose
```diff
@@ -42,15 +40,13 @@ node(NODE) {
...
             rm -rf /var/lib/containers && mkdir /var/lib/containers
-            mkdir -p ${treecompose_workdir}/containers
-            mount --bind ${treecompose_workdir}/containers /var/lib/containers
+            mkdir -p ${host_workspace}/containers
+            mount --bind ${host_workspace}/containers /var/lib/containers
             skopeo inspect docker://${OSCONTAINER_IMG}
             podman pull ${OSCONTAINER_IMG}
             cid=\$(podman run --net=host -d --name oscontainer --entrypoint sleep ${OSCONTAINER_IMG} infinity)
-            ln -sf \$(podman mount \${cid})/srv/repo ${treecompose_workdir}/repo
+            ln -sf \$(podman mount \${cid})/srv/repo ${host_workspace}/repo
```
miabbott: `repo` is defined on line 33 as `${host_workspace}/repo`...is there a reason for not using `${repo}` everywhere there is `${host_workspace}/repo`?

cgwalters: Yeah that'd probably be saner but...hold, I am testing another PR that's simpler.

Jenkinsfile.treecompose
```diff
@@ -112,18 +108,18 @@ node(NODE) {
       // with Docker/OCI images.  Ideally there'd be a way to "bind mount" content
       // into the build environment rather than copying.
       stage("Prepare repo copy") { sh """
-          cp -a --reflink=auto * ${treecompose_workdir}/
-          repo=\$(readlink ${treecompose_workdir}/repo)
-          rm ${treecompose_workdir}/repo
-          cp -a --reflink=auto \${repo} ${treecompose_workdir}/repo
+          cp -a --reflink=auto * ${host_workspace}/
+          repo=\$(readlink ${host_workspace}/repo)
```
miabbott: Could `repo` here get renamed to something like `original_repo` to disambiguate things?

Issue #201, opened 2018-08-03
oscontainer tags


cgwalters:
We should make use of tags; today we just push `:latest`, and I manually pushed a `:alpha` tag.

It'd be great to have someone look at automatically tagging :alpha after it passes tests - and also ensuring that once this is done we have corresponding images generated (and hmm....we should probably put the kickstart in the oscontainer too!)

Also, to make things more complex: I think we should defer uploading AMIs until after basic qemu sanity checking. Right now the AMI upload is 20 minutes; we could have run tons of basic qemu tests in that time.

So maybe the flow is e.g.:

treecompose -> `:latest`
basic sanity tests -> `:alpha-sanity` -> upload to EC2
Once it's in EC2 we do a lot more extensive testing, maybe of a cluster -> `:alpha-cluster-sanity`

Promotion from `:alpha-cluster-sanity` to e.g. `:beta` would be manual?

Yet another tricky thing in all of this is that we will almost certainly need to (in practice) decouple OS-basics testing from kube-and-above testing.

cgwalters:
So one thing I learned is that there's no Docker registry API to list untagged images.  However, OpenShift "ImageStreams" do maintain this history.  So we could base some of our infrastructure on that:

```
apiVersion: image.openshift.io/v1
kind: ImageStream
metadata:
  creationTimestamp: 2018-07-23T19:21:43Z
  generation: 1
  name: os
  namespace: rhcos
  resourceVersion: "38797714"
  selfLink: /apis/image.openshift.io/v1/namespaces/rhcos/imagestreams/os
  uid: a25d7043-8ead-11e8-b23a-42010a8e0004
spec:
  lookupPolicy:
    local: false
status:
  dockerImageRepository: docker-registry.default.svc:5000/rhcos/os
  publicDockerImageRepository: registry.svc.ci.openshift.org/rhcos/os
  tags:
  - items:
    - created: 2018-07-23T21:02:05Z
      dockerImageReference: docker-registry.default.svc:5000/rhcos/os@sha256:8d1781ccb9dd9296321857b0b17b9331efddbdc9b64d13b49cf06b90a897796b
      generation: 1
      image: sha256:8d1781ccb9dd9296321857b0b17b9331efddbdc9b64d13b49cf06b90a897796b
    tag: alpha
  - items:
    - created: 2018-08-03T18:21:23Z
[...]

PR #210, opened 2018-08-07
jenkins: Redirect pkg diff to file

Rahuls0720:
Archives pkg_diff.txt and syncs a copy of the file to the artifact server.

Jira card: https://projects.engineering.redhat.com/browse/COREOS-275

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -5,6 +5,8 @@ def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-maipo:latest"
 
 // We write to this one for now
 def artifact_repo = "/srv/rhcos/output/repo"
+// We write pkg_diff.txt here
+def images = "/srv/rhcos/output/images"
 
 node(NODE) {
     def manifest = "host.yaml"
@@ -81,21 +83,27 @@ node(NODE) {
         // delete the ref head (so rpm-ostree won't add a parent) and
         // then we do a repo prune after.
         def version, commit
-        stage("Compose Tree") { sh """
-            rm ${repo}/refs/heads/* -rf
-            env G_DEBUG=fatal-warnings rpm-ostree compose tree --repo=${repo} \
-                                                        --add-metadata-string=version=${version_prefix}.${env.BUILD_NUMBER} \
-                                                        ${manifest}
-            ostree --repo=${repo} rev-parse ${ref} > commit.txt
-        """
-            commit = readFile('commit.txt').trim();
-            version = utils.get_rev_version("${repo}", commit)
-            currentBuild.description = "🆕 ${version} (commit ${commit})";
-            if (previous_commit.length() > 0) {
-                sh "rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit}"
+        try {
+            stage("Compose Tree") { sh """
+                rm ${repo}/refs/heads/* -rf
+                env G_DEBUG=fatal-warnings rpm-ostree compose tree --repo=${repo} \
+                                                            --add-metadata-string=version=${version_prefix}.${env.BUILD_NUMBER} \
+                                                            ${manifest}
+                ostree --repo=${repo} rev-parse ${ref} > commit.txt
+            """
+                commit = readFile('commit.txt').trim();
+                version = utils.get_rev_version("${repo}", commit)
+                currentBuild.description = "🆕 ${version} (commit ${commit})";
+                if (previous_commit.length() > 0) { sh """
+                    rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit} > ${WORKSPACE}/pkg_diff.txt
+                    mkdir -p ${images}
+                    cp ${WORKSPACE}/pkg_diff.txt ${images}/pkg_diff.txt"
+                """ }
+                sh "ostree --repo=${repo} prune --refs-only --depth=0"
+                sh "ostree summary --repo=${repo} --update"
             }
-            sh "ostree --repo=${repo} prune --refs-only --depth=0"
-            sh "ostree summary --repo=${repo} --update"
+        } finally {
+            archiveArtifacts artifacts: "pkg_diff.txt", allowEmptyArchive: true
         }
 
         // This takes OSTree repo and copies it out of the original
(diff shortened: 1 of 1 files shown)


Rahuls0720:
@cgwalters How would you test that the pkgdiff.txt file appears at http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/images/cloud/latest/?

cgwalters:
Yeah, the treecompose job doesn't currently sync there.  We do sync out to http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/repo/ - what's `${repo}` in this job, and so we could just dump the text file in there?

But that would mean we'd only get diffs of the latest.

This quickly gets into https://github.com/openshift/os/issues/201 as well as https://github.com/openshift/os/pull/212

Rahuls0720:
`${repo}` refers to `/srv/rhcos/treecompose/repo` (I didn't update repo's definition in the commit).

I noticed treecompose isn't synced to http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/images/cloud/latest/ so I added `dirpath` (not sure if it will work, I was hoping to test it using the PR). Another method could be passing the pkgdiff.txt file as a parameter to jenkinsfile.cloud but I'm not sure how wait: false might affect this (https://github.com/openshift/os/blob/master/Jenkinsfile.treecompose#L154).

Yes, this is only posting diffs of the latests. When starting work on this card, I was confused why this takes priority over #201. But, if I can test that this works, updating it to pkg diff tags should be straight forward.

Rahuls0720:
@cgwalters Do you know the login/password for the jenkins account or how to set one up? I'm not able to get in using my normal RH login.

I guess I could sync the files to http://aos-ostree.rhev-ci-vms.eng.rdu2.redhat.com/rhcos/repo/. I feel that it would be most useful to have the pkg diff live in the same directory as the tagged image. Something like having the latest directory contain a pkg_diff.txt between the previous commit and the current. In the alpha directory, having a pkg_diff.txt between latest and alpha, and so on.

cgwalters:
> Do you know the login/password for the jenkins account or how to set one up? 

It's still a huge pain to test jobs, particularly in a way that doesn't end up overwriting prod data.  That said I granted you access (I think).

Jenkinsfile.treecompose
```diff
@@ -92,7 +94,9 @@ node(NODE) {
             version = utils.get_rev_version("${repo}", commit)
             currentBuild.description = "🆕 ${version} (commit ${commit})";
             if (previous_commit.length() > 0) {
-                sh "rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit}"
+                sh "mkdir -p ${images}"
+                utils.rsync_dir_in(ARTIFACT_SERVER, KEY_FILE, images)
```
Rahuls0720: If `images` is the best place for this file to live, should I create a `sync in` stage?

Rahuls0720:
@miabbott When you get a chance, can you look over this.

Jenkinsfile.treecompose
miabbott: You don't need to sync anything in for this operation.  Just making the directory and writing out the pkg diff should be sufficient.

Jenkinsfile.treecompose
```diff
@@ -92,7 +94,9 @@ node(NODE) {
...
+                sh "rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit} > /${images}/pkg_diff.txt"
```
miabbott: I think we should archive the `pkg_diff.txt` each time it is created, as it will be overwritten whenever there is a new commit to diff.

Perhaps you can write out the diff to `${WORKSPACE}/pkg_diff.txt` and then copy it to `${images}`.  Having the file in the workspace would allow us to use the `archiveAritifacts()` function like we do in the cloud pipeline:

https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L94-L96

Jenkinsfile.treecompose
Rahuls0720: I noticed the `cloud` pipeline sync it here: https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L36. If it isn't sync, wouldn't files/directories get overwritten during treecompose's `rsync-out` stage?

Rahuls0720: ignore ^, `archiveAritifacts` solves this

Jenkinsfile.treecompose
Rahuls0720: Updated. If we archive pkg_diff.txt, doesn't it still get overwritten?

miabbott: My understanding of `archiveArtirfacts` is that it keeps the files around on the Jenkins host until they are pruned.  For example, in the `cloud` pipeline, we archive the `rhcos-qcow2-install.txt`:

https://continuous-infra-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/rhcos/job/coreos-rhcos-cloud/1561/artifact/rhcos-qcow2-install.txt

https://continuous-infra-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/rhcos/job/coreos-rhcos-cloud/1558/artifact/rhcos-qcow2-install.txt

So we are able to keep files per build, but it is overwritten on the artifact server in the `sync out` stage.

Jenkinsfile.treecompose
miabbott: As I understand it, in the `cloud` pipeline, that sync in is used so that the previous `commit.txt` artifact can be referenced in the `Check for changes` stage:

https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L60

Since writing out the pkg diff doesn't require the knowledge of any previous artifacts in that `images` directory, I believe the `sync-in` can be skipped.

Rahuls0720:
@miabbott Can you look over this PR once again before you leave for the conference (and my internship ends)?

Jenkinsfile.treecompose
```diff
@@ -81,21 +83,27 @@ node(NODE) {
...
-                                                        --add-metadata-string=version=${version_prefix}.${env.BUILD_NUMBER} \
-                                                        ${manifest}
-            ostree --repo=${repo} rev-parse ${ref} > commit.txt
-        """
-            commit = readFile('commit.txt').trim();
-            version = utils.get_rev_version("${repo}", commit)
-            currentBuild.description = "🆕 ${version} (commit ${commit})";
-            if (previous_commit.length() > 0) {
-                sh "rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit}"
+        try {
```
miabbott: I don't think it's necessary to wrap this whole stage in a `try{}`, but it shouldn't cause any problems leaving it in.

miabbott:
@Rahuls0720 I had one minor comment that doesn't need to be resolved.

I think this is ready to get merged.  :tada: 

First we'll test it, though

miabbott:
Let's roll with it!

/lgtm

Jenkinsfile.treecompose
Rahuls0720: I could be totally wrong since jenkins is (seems to be) unpredictable, but it failed saying it doesn't understand the command `archiveArtifacts` when I had the `finally` statement here-ish: https://github.com/Rahuls0720/rhcos/blob/pkg_diff-jenkinsTESTING/Jenkinsfile#L98

Rahuls0720: I'll try testing it without the try again

Rahuls0720: LOL nvm ^...already merged :)

PR #211, opened 2018-08-07
 jenkins/cloud: Stop binding /srv, use workspace

cgwalters:
Just use $WORKSPACE for the images - it'll be naturally namespaced
by our job and lifecycled with it.

Note I tried *really* really hard to also drop `--privileged` but
the problem is the usage of bwrap drags us into unprivileged
nested containers.

One downside is we'll be rsyncing in the images on each job
invocation, but I'll work on a fix for that later.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -1,14 +1,13 @@
 def TIMER = "H H/3 * * *"
 def NODE = "rhcos-jenkins"
 def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler"
-def DOCKER_ARGS = "-v /srv:/srv --privileged --device /dev/kvm"
+// See https://github.com/projectatomic/bubblewrap/issues/284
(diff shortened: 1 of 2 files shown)


cgwalters:
See bwrap issue 284 for where I got on the unprivileged bits.

cgwalters:
Rebased :surfing_man:

jlebon:
> One downside is we'll be rsyncing in the images on each job
invocation, but I'll work on a fix for that later.

Yeah, that will mean a *massive* amount of data to rsync on each run. May I suggest until we fix that, we drop the workspace cleanup here?

cgwalters:
One approach we could take would be to do pruning (and s3 sync) as a separate task.  So we drop the `--delete`, and this job only writes a new directory and swaps the `latest` symlink, which wouldn't require syncing in anything.

jlebon:
That makes sense, though... is it possible to do the symlink swap without having to `rsync` the parent directory? I guess we could revert to `scp` for that.

PR #212, opened 2018-08-08
WIP: promote 'os-maipo' image to 'alpha' after successful sanity test

miabbott:
Now that we have a basic `kola` test in the pipeline, we can promote our images to 'alpha' status after a successful test.

To do this, we create an additional tag using the commit ID for the `os-maipo` image after its created.  Then, we reference this tag when doing the promotion to 'alpha'

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -2,6 +2,8 @@ def TIMER = "H H/3 * * *"
 def NODE = "rhcos-jenkins"
 def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler"
 def DOCKER_ARGS = "-v /srv:/srv --privileged --device /dev/kvm"
+def API_CI_REGISTRY = "registry.svc.ci.openshift.org"
+def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-maipo"
 
 // Are there really other regions?  We're sending explorers to find out
 def AWS_REGION = "us-east-1"
@@ -21,8 +23,23 @@ node(NODE) {
     utils = load("pipeline-utils.groovy")
     utils.define_properties(TIMER)
 
+    // Split the credentials up
+    def username, password;
+    withCredentials([
+        usernameColonPassword(credentialsId: params.REGISTRY_CREDENTIALS, variable: 'CREDS'),
+    ]) {
+        (username, password) = "${CREDS}".split(':')
+    }
+
     docker.image(DOCKER_IMG).pull()
     docker.image(DOCKER_IMG).inside(DOCKER_ARGS) {
+    stage("Login") { sh """
+        set +x
+        echo podman login -u ${username} -p '<password>' ${API_CI_REGISTRY}
+        podman login -u "${username}" -p "${password}" ${API_CI_REGISTRY}
+        echo "login done"
+    """ }
+
     stage("Clone qcow2-to-vagrant") {
         // I tried to use git() but not sure where that went, it isn't the workspace
         sh('git clone https://github.com/cgwalters/qcow2-to-vagrant')
@@ -104,6 +121,17 @@ node(NODE) {
     """ }
     } finally {
         archiveArtifacts artifacts: "rhcos-basic.tap", allowEmptyArchive: true
+    }
+
+    // We passed some sanity checks, so pull the image again, retag it, and push
+    stage("Push an 'alpha' tag") {
+        if (params.DRY_RUN) {
+            echo "DRY_RUN set, skipping alpha tag"
+        } else if (!params.DRY_RUN) { sh """
+            podman pull ${OSCONTAINER_IMG}:${latest_commit}
+            podman tag ${OSCONTAINER_IMG}:${latest_commit} ${OSCONTAINER_IMG}:alpha
+            podman push ${OSCONTAINER_IMG}:alpha
+        """ }
     }
 
     def dirname, img_prefix, commit, version, dirpath, qcow, vmdk, ec2, vagrant_libvirt
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -2,7 +2,7 @@ def TIMER = "H/30 * * * *"
 def NODE = "atomic-jslave-autobrew"
 def DOCKER_IMG = "quay.io/cgwalters/coreos-assembler"
 def API_CI_REGISTRY = "registry.svc.ci.openshift.org"
-def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-maipo:latest"
+def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-maipo"
 def DOCKER_ARGS = "--net=host -v /srv:/srv -v /run/docker.sock:/run/docker.sock --privileged"
 
 def treecompose_workdir = "/srv/rhcos/treecompose"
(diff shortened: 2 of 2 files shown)


miabbott:
I labeled this as WIP because I'm looking for feedback about:

1.  tagging the original image with `:latest`, `:commit ID` and `:version`
2.  doing the promotion to 'alpha' by using the `:commit ID` tag as the source

cgwalters:
Is the idea here that the test system says "this ostree commit hash passed", and we need to backreference from the ostree-hash to the container-hash?

Soo...yes this is a problem.  Part of where I was going is to de-emphasize the ostree hash in favor of the container-hash.  See [this PR](https://github.com/projectatomic/rpm-ostree/pull/1406).  But what isn't fully plumbed through here is having our cloud images *build* using a pivot.  I.e. when you boot them and run `rpm-ostree status` you'd see `oscontainer://image@sha256:5891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03` or so.

Once we do that though, we basically stop thinking about the ostree-hash and that would solve the problem here, right?

It's not going to be trivial though to change to do oscontainer-based builds...I was thinking about it and it'd really make the most sense to do in Anaconda (or drop Anaconda, which is its own ball of work).

So, short term, I think rather than doing 3 tags, let's just maintain a simple JSON file in the releng space that is a bijection between the ostree-hash and the container-hash?

cgwalters:
Although, we have a mapping from container-hash → ostree-hash thanks to what you did [with labels](https://github.com/openshift/os/pull/181).  So one thing we could do is - in the cloud image builds, walk the imagestream and find the oscontainer that has a label with the ostree-hash.

cgwalters:
>  the version of cloud image

Though another problem here we haven't formalized quite is that "cloud image version" versus "ostree version" but let's ignore that for now.

> Had to look up bijection to be certain 

TBH I looked it up too :smile: But I wanted to be a bit formal about it.

> I'll close this PR and open a new one for that work.

Are you able to log into the api.ci namespace?  I can try to quickly script up a bit to generate the mapping.

miabbott:
> Though another problem here we haven't formalized quite is that "cloud image version" versus "ostree version" but let's ignore that for now.

We haven't?  My understanding of the cloud pipeline was that we were building the QEMU image from the content in the ostree repo.  Wouldn't that tie the image version to the ostree version?  (I guess it is possible that the repo was updated before the image is built, but I don't think we have scaled up enough to hit that race yet.)

cgwalters:
> Wouldn't that tie the image version to the ostree version?

It's true that today the pipeline will only build an image if the ostree version changes; I think we can probably just rely on that, but my feeling was that that logic was added as an *optimization*.  Today things like the partitioning can change without the ostree changing, and the only way to generate a new image would be to bump the ostree.

cgwalters:
OK I wrote this: https://github.com/cgwalters/playground/blob/9b4573c2a7fbd9b2e5ad756ed0b4ec45f682d510/rust/imagestream-ostree-hash/src/imagestream-ostree-hash.rs

Output:

```
$ cargo run os-maipo
   Compiling imagestream-ostree-hash v0.1.0 (file:///var/srv/walters/src/github/cgwalters/playground/rust/imagestream-ostree-hash)
warning: unused variable: `oscontainer_to_ostree`
  --> src/imagestream-ostree-hash.rs:54:9
   |
54 |     let oscontainer_to_ostree : BTreeMap<String, String> = BTreeMap::new();
   |         ^^^^^^^^^^^^^^^^^^^^^ help: consider using `_oscontainer_to_ostree` instead
   |
   = note: #[warn(unused_variables)] on by default

    Finished dev [unoptimized + debuginfo] target(s) in 1.65s
     Running `target/debug/imagestream-ostree-hash os-maipo`
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:524bce97199ecba000408ff40031ea4003a20ae8a09ce5cc1a876985d688f37e Some("a6d9144d664945c4e6cd517a05d117497caf761ec242faf2e5ad7a3eb492db18")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:ba318f7d02b588277093dd98c75289feedf7aafedf56599da8ac0b77352980b5 Some("d13796fa8cda7d4bb73586bc6d10115703490c7c0595c51ccfcfcd21fa5d0634")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:c42bfd57eb8fd0ef5cdc6c53454a7dac702aa883c2c22cd3eafa4288f3f2bb9e Some("3375564c10dd3379cbb643ffe6360bf0298111a69a9ec2b6c2c3067fc06fb07b")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:0afa392b91e5c36397648f1ca8c878a96abe948688f9f9460146ce794dace32f Some("b8bda959bd9d37de7e7e183e9043ba93f0ef20ba417a86220dfb49c84dce321e")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:9f0bce4be66563b50f1f22da14d330305b6ec6affbac9ecfd8b5dad0e57d56da Some("6912c0742e4b90c3929e6aa4a75c25258b29fe7801db4ed88f70f8f97f59ea59")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:8a3607f9ff501dc23241a5c478c8dc4850d40fc7351d3a2a1ef768be506b163c Some("f05f4ed2fec6892e03dc6645fcf059234306e4fee370f0808a1a6758ec012ac8")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:3f387e40acbf8b6f84f395487a99e5f474b069ac9fd40c8917e6197e2acee324 Some("919fddc0da916ab8b33677a376e69b660a0bc74cf2d13e236662e637c47a79a3")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:161220ce6e59675c38f18a2fe6c58d44c7b706c7e63f57b9c7086a0cf677e868 Some("3555cb2f0f673c7285038e492e72a5738977c6cb2fa684c103c1568a3da9e110")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:dbf3af3ca8dafe1f6d3f3d0ac8b5753f3e546e6b46aa0f5a7cfc6dc4680d1ef3 Some("5af3c99522bd47304e401d51a54851215a54b789d098f997145072016bf87b07")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:a82c21a0b00ce77473f32ea6827a657e8fe01f6058299bb8e36f9c703479c47f Some("82cf883e1deca68ffb51222133da62411935c10a263d9e6e833086bd15e6932a")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:11c1b9097dcdcf177d5be93f9eb0bbe0378ed312661cd6249ed641a5d0ebc30e Some("511e3ce2cb3953f83e5fd1d0eb73b103a2e922e679728a6b7f71e29d7b8877bb")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:6e075206cea13c5f2942095f54568be7b05cb4204386ab241077125979f07052 Some("60569af7b58ee63a58f92dad10945b247d0d0a9f5a3838f1d92df140dd9dd724")
docker://registry.svc.ci.openshift.org/rhcos/os-maipo@sha256:85b4be7a3919b2f24d86162f8a644c5e952f3a45b070211c4668bf0fd19d658f Some("6897f6dd8517281433b513d9f00c3c717ed6cf6c1fe93283ca23e3fe2b5d32be")

```

cgwalters:
Though that was fun...deploying it is going to be hacky.  The more I think about this we really need to pull the band aid off and have the cloud job pull the os container.  Which is what I was working on; the prep task https://github.com/openshift/os/pull/209 was part of this.

miabbott:
Based on these conversations and ongoing changes to the pipeline, it seems like this attempt was a little early.  I'm going to close this and revisit later.

PR #213, opened 2018-08-08
jenkins/treecompose: Just use ${WORKSPACE} rather than -v /srv

cgwalters:
It turns out that ${WORKSPACE} *is* on the host, under `/home/jenkins`.
So let's just use that.  This is part of making our jobs more fully
containerized.


miabbott:
/lgmt

Let's go for it

miabbott:
/lgtm

i guess edits don't trigger the bot...

jlebon:
Sorry, I should've jumped in earlier on that. Just didn't have the full context loaded in. That's correct, the `$WORKSPACE` is under the bind mount we pass through to the swarm container. This is required for e.g. PAPR too since we need to be able to `docker run` from inside -- see e.g. https://github.com/projectatomic/papr/blob/044ff232a5a3a4414d683284800cde33e72ed76a/Dockerfile#L20.

PR #214, opened 2018-08-08
jenkins: Read ref and version from manifest file

cgwalters:
Avoid duplicating things in the pipeline in prep for making the pipeline
more generic.


jlebon:
`readYaml`... neat!
/lgtm

jlebon:
Follow up to this in https://github.com/openshift/redhat-release-coreos/pull/10.

PR #216, opened 2018-08-08
treecompose: don't copy a dir into itself

miabbott:
Before we used `$WORKSPACE` as the base for `$treecompose_workdir`,
we were copying all the contents of `$WORKSPACE` into what was
`/srv/rhcos/treecompose`.  Trying to do that same copy operation now
gives us:

`cp: cannot copy a directory, 'treecompose', into itself, '/home/jenkins/workspace/coreos-rhcos-treecompose/treecompose/treecompose'`

So let's stop doing that.


miabbott:
I think this is safe to do?  And with that statement, I have admitted that this was not tested.

PR #219, opened 2018-08-09
treecompose: correct location of repo in Dockerfile

miabbott:
The container build process expects to be able to copy the ostree repo
into the container.  Since the restructuring of the workspace and use
of `$treecompose_workdir`, this means the location of the repo needs
to be updated in the `Dockerfile`.


miabbott:
WIP until I can test the pipeline again

miabbott:
WIP lifted

Turns out that the `COPY` directive is relative to the build context; you can't use absolute paths

PR #220, opened 2018-08-09
Do backups asynchronously

cgwalters:
This takes us to a single point of truth for data; the other is
clearly secondary.

### DIFF
--- a/Jenkinsfile.backup
+++ b/Jenkinsfile.backup
@@ -1,0 +1,45 @@
+def TIMER = "H/30 * * * *"
+def NODE = "rhcos-jenkins"
+
+def dirs_to_backup = ["rdgo", "images" ];
+
+node(NODE) {
+    def par_stages = [:]
+    stage("Clean workspace") {
+       step([$class: 'WsCleanup'])
+    }
+    checkout scm
+    utils = load("pipeline-utils.groovy")
+    utils.define_properties(TIMER)
+
+    for (dir in dirs_to_backup) {
+        // Avoid iteration+closure trap
+        def basetarget = dir;
+        par_stages["backup-" + basetarget] = { -> stage("Backup " + basetarget) {
+    utils.inside_assembler_container("") {
+        def remote = "/srv/rhcos/output/${basetarget}";
+        def local = "${WORKSPACE}/${basetarget}";
+        stage("Sync In") {
+            withCredentials([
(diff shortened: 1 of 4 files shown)


miabbott:
Since this is run on a timer, do we run the risk of backing up the results of a partially finished `rdgo` or `cloud` pipeline?  Or phrased differently, do we care if we back up a partially finished pipeline?

Jenkinsfile.backup
```diff
@@ -0,0 +1,43 @@
...
+        // Expand, but also avoid iteration+closure trap
+        def target = "/srv/rhcos/output/${dir}";
+        par_stages["backup-" + target] = { -> stage("Backup " + target) {
+    utils.inside_assembler_container("") {
+        stage("Sync In") {
+            withCredentials([
+                string(credentialsId: params.ARTIFACT_SERVER, variable: 'ARTIFACT_SERVER'),
+                sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
+            ]) {
+                sh "mkdir -p ${target}"
```
jlebon: How about keeping this under `$WORKSPACE` and use the `rsync_dir_in_dest()` from #211? I really like the move away from `/srv`.

Jenkinsfile.backup
```diff
@@ -0,0 +1,43 @@
...
+       step([$class: 'WsCleanup'])
+    }
+    checkout scm
+    utils = load("pipeline-utils.groovy")
+    utils.define_properties(TIMER)
+
+    for (dir in dirs_to_backup) {
+        // Expand, but also avoid iteration+closure trap
+        def target = "/srv/rhcos/output/${dir}";
+        par_stages["backup-" + target] = { -> stage("Backup " + target) {
```
jlebon: Let's use (the equivalent of) `$(basename $target)` for the stage name?

cgwalters:
> Or phrased differently, do we care if we back up a partially finished pipeline?

Yeah, we could add exclusions but...eh, doesn't seem worth it.  At least for rdgo though we need to teach it to make datestamped-dirs that will sync more naturally.

jlebon:
It might be as simple as wrapping the "Sync in" here and the "Sync out" in the cloud/rdgo jobs with `lock()`: https://jenkins.io/blog/2016/10/16/stage-lock-milestone/#lock. I can take a look into that after this merges.

PR #221, opened 2018-08-09
cloud: read host.yaml after SCM checkout

miabbott:
When using a freshly configured pipeline (i.e. new devel pipeline),
reading the `host.yaml` will fail.  But moving it after the SCM
checkout is successful.


ashcrow:
The bot hates me.

cgwalters:
Hmm, it should work for you too.  I think there's some ongoing github/prow issues though.  Let's see...

/lgtm

ashcrow:
Yeah, I think you're right @cgwalters.

Issue #116, opened 2018-06-18
Make SELinux enforcing


cgwalters:
AFAIK we have `enforcing=0` as a hack to bringup Ignition.  We need to ship in enforcing mode.

sdemos:
Seems like an important part of this is making ignition properly set selinux labels for files it writes - https://github.com/coreos/bugs/issues/2417

cgwalters:
Work ongoing in https://github.com/coreos/ignition/pull/569

PR #224, opened 2018-08-13
cloud.ks: stop disabling SELinux

jlebon:
Ignition + SELinux works now, so stop disabling it.
Ref: coreos/ignition#569

Closes: #116


jlebon:
Working! :heavy_check_mark: 
/hold cancel

PR #226, opened 2018-08-13
cloud: archive freshly built qcow2

miabbott:
If our sanity tests fail, we want to be able to grab the failing image
and debug locally.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -99,12 +99,16 @@ node(NODE) {
     try {
     stage("Kola rhcos.basic") { sh """
         kola version
-        kola -b rhcos --tapfile rhcos-basic.tap --qemu-image ${image} run rhcos.basic
+        if ! kola -b rhcos --tapfile rhcos-basic.tap --qemu-image ${image} run rhcos.basic; then
+          cp ${image} ${WORKSPACE}
+          exit 1
+        fi
     """ }
     } finally {
         sh 'tar -cJf _kola_temp.tar.xz _kola_temp'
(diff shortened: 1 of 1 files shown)


miabbott:
We got bit by this last week, so let's make it easier on all of us.

Note, this wasn't tested in a devel pipeline....

Jenkinsfile.cloud
```diff
@@ -89,10 +89,12 @@ node(NODE) {
         bwrap --bind / / --dev-bind /dev /dev --proc /proc --unshare-pid runuser -u builder -- coreos-virt-install --dest ${image} --create-disk \
                 --console-log-file ${WORKSPACE}/rhcos-qcow2-install.txt \
                 --wait=25 --kickstart cloud.ks --location "${INSTALLER_TREE_URL}"
+        cp ${image} ${WORKSPACE}
         """
     } }
     } finally {
         archiveArtifacts artifacts: "rhcos-qcow2-install.txt", allowEmptyArchive: true
+        archiveArtifacts artifacts: "rhcos.qcow2", allowEmptyArchive: true
```
jlebon: How about something like this in the "Kola rhcos.basic" step:

```
stage ... {
  kola version
  if ! kola -b rhcos ...; then
    cp ${image} ${WORKSPACE}
    exit 1
  fi
} finally {
  ...
  archiveArtifacts artifacts: "rhcos.qcow2", allowEmptyArchive: true
}
```

?

That way we only copy it and archive it if the test failed. If the test passed, then we'll already have access it to it at the usual place.

miabbott: This is smart.  I will explore this.  Thanks!

PR #228, opened 2018-08-14
Jenkinsfile.cloud: +8G the qemu qcow2

ashcrow:
The installer requires the image to start off with an extra
+8G developer usage. @eparis noted it would be helpful to have
this happen in the pipeline rather than each developer doing it
on their own.

In the (near) future it would be preferable for the installer
to handle this.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -166,6 +166,8 @@ node(NODE) {
     parallel par_stages; par_stages = [:]
 
     stage("Finalizing, generate metadata") {
+        // See https://github.com/openshift/installer/blob/master/Documentation/dev/libvirt-howto.md#13a-rhcos
+        sh "qemu-img resize ${qcow} +8G"
         sh "gzip < ${qcow} > ${img_prefix}-qemu.qcow2.gz"
         sh "ln -sr ${img_prefix}-qemu.qcow2.gz ${dirpath}/rhcos-qemu.qcow2.gz"
         // Everything above in parallel worked on qcow, we're done with it now


dustymabe:
/me wonders why we are resizing rather than just making the disk bigger when we create it?

dustymabe:
> /me wonders why we are resizing rather than just making the disk bigger when we create it?

is it because we only want the qemu image to be the one that is bigger?

Jenkinsfile.cloud
```diff
@@ -166,6 +166,8 @@ node(NODE) {
     parallel par_stages; par_stages = [:]
 
     stage("Finalizing, generate metadata") {
+        // See https://github.com/openshift/installer/blob/master/Documentation/dev/libvirt-howto.md#13a-rhcos
+        sh "qemu-img resize ${img_prefix}-qemu.qcow2 +8G"
```
arithx: The gzip command below is still taking in `${qcow}` which is `${img_prefix}.qcow2`

ashcrow:
> is it because we only want the qemu image to be the one that is bigger?

That's the only one that I'm aware of that needs to be larger. We may want to bump up the vagrant ones as well.

cgwalters:
The default size of the default disk is going to be influenced by the base size of all of the containers in the default install right?  It feels like we'll end up needing to track that somewhere; probably the installer, but I also don't see a problem with just making all of our images unconditionally larger.

Doing the latter would be just a matter of tweaking the magic comment at the top of `cloud.ks`.

ashcrow:
> Doing the latter would be just a matter of tweaking the magic comment at the top of cloud.ks.

In that case should I close this in favor of changing `cloud.ks` to up the size? Is it safe to assume the images will be growable rather than literally adding an extra 8G per artifact?

dustymabe:
> The default size of the default disk is going to be influenced by the base size of all of the containers in the default install right?

right. To me the best place for this long term is to have the installer request an appropriate size of the image (just like it would do for public cloud) and have that "size" live inside the installer. This PR is a temporary solution IMHO and we have noted that in our previous conversation we had about it.

ashcrow:
> This PR is a temporary solution [..]

Agreed.

ashcrow:
I opened https://github.com/openshift/installer/issues/126 so it doesn't get forgotten.

cgwalters:
It looks like AWS supports OVA upload, and hopefully at least that supports being sparse; something to look into later.

/lgtm

jlebon:
IIRC, the reason we upload VMDK today is because that's what `ore` supports (along with raw). Though it shouldn't be very hard to add support for a different type supported by the AWS API.

PR #231, opened 2018-08-17
remove extra quote

Rahuls0720:
Big shout out to @ashcrow for finding this syntax error causing jenkins to fail.


Rahuls0720:
@dm0- can you also add a lgtm label?

PR #232, opened 2018-08-20
Jenkinsfile.cloud: fix image archive

arithx:
The image being archived doesn't exist. Switches to the variable rather
than a direct name.

Not tested.


cgwalters:
FWIW I just direct pushed https://github.com/openshift/os/commit/25ce4a6467262cbdfedbc0bad25d841e3f03d7ba

PR #236, opened 2018-08-23
WIP: ootpa oscontainer images

cgwalters:
The treecompose job works.  Obviously sharing code between
the jobs would be nice.  I still need to push some WIP I have
for `coreos-virt-install` to run `ksflatten`.

There are some SELinux denials and failing services (it looks
like preset is enabling sssd socket units somehow?)

Anyways submitting as WIP for now.


cgwalters:
This now depends on https://github.com/openshift/os/pull/252 - and includes a bit of work in there starting on image builds: splitting up the kickstarts.  (maipo requires the `auth` verb and hence the `authselect` package, ootpa fails with `auth` enabled due to [this issue](https://github.com/pbrezina/authselect/issues/48))

cgwalters:
To clarify, this WIP is now about generating the disk images, which is also going to require some more pipeline refactoring.

cgwalters:
I'm going to redo this one entirely to use coreos-assembler.

PR #237, opened 2018-08-24
Split host into base + maipo

cgwalters:
Prep for adding ootpa.  Split off a `host-base.yaml` that
contains things common to both, and rename the manifest to
`host-maipo.yaml`.


cgwalters:
I verified that `rpm-ostree db diff` for this is a no-op.

PR #241, opened 2018-08-24
jenkins: Automatically make jobs not starting with rhcos- devel

cgwalters:
I keep having the timer set on my test job, which ends up overriding
DRY_RUN and pushing to prod, which I obviously don't want.

### DIFF
--- a/pipeline-utils.groovy
+++ b/pipeline-utils.groovy
@@ -7,11 +7,13 @@
 
 def define_properties(timer) {
 
-    // Set this to TRUE to disable the timer, and set DRY_RUN=true by default
-    def developmentPipeline = false;
+    // Name your test jobs something not starting with rhcos-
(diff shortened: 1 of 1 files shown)


pipeline-utils.groovy
```diff
@@ -7,12 +7,14 @@
 
 def define_properties(timer) {
 
-    // Set this to TRUE to disable the timer, and set DRY_RUN=true by default
-    def developmentPipeline = false;
+    // Name your test jobs something not starting with rhcos-
+    def developmentPipeline = !"${env.JOB_NAME}".startsWith("rhcos-");
```
jlebon: `!env.JOB_NAME.startsWith()` should work, no?

cgwalters: Probably, but now I haven't tested it.

PR #244, opened 2018-08-28
jenkins: Refactor treecompose oscontainer pull to script

cgwalters:
Prep for adding ootpa; we can more easily reuse it in multiple
pipelines.

### DIFF
--- a/scripts/pull-mount-oscontainer
+++ b/scripts/pull-mount-oscontainer
@@ -1,0 +1,28 @@
+#!/usr/bin/bash
+set -xeuo pipefail
+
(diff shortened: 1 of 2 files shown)


scripts/pull-mount-oscontainer
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/bash
+set -xeuo pipefail
+
+# First, login.  We expect a CREDS environment variable.
+registry=$1; shift
+set +x
+username=${CREDS%:*}
+password=${CREDS##*:}
```
miabbott: bash...your syntax is so crazy

Issue #243, opened 2018-08-27
systemd presets for units from ignition don't get evaluated


yuqi-zhang:
Doing -b rhcos with ami ami-0fc9356f07902b569 shows the following failure:
```
 --- FAIL: coreos.ignition.v2.once (123.83s)
        cluster.go:117: "test ! -e /etc/ignition-ran" failed: output , status Process exited with status 1
```
The original test does: `sudo rm /etc/ignition-ran` and reboots the machine. It seems that the reboot has re-created the file.

Raising issue here as suggested by @arithx. Original mantle issue: https://github.com/coreos/mantle/issues/905

dustymabe:
what does `sudo journalctl -b 0 -u ignition-files.service` show ?

yuqi-zhang:
After reboot: https://paste.fedoraproject.org/paste/gl47Xr76MgTmFSQJAtadrw

dustymabe:
yeah it looks like ignition is running on every boot because the `/boot/coreos-firstboot` is not getting deleted. It looks like `coreos-firstboot-complete.service` from https://github.com/dustymabe/ignition-dracut/pull/12 is not getting run.

dustymabe:
looks like an issue with systemd presets not getting evaluated as mentioned here https://github.com/coreos/ignition/issues/624#issuecomment-416381319 - maybe we need a new issue to track that.

yuqi-zhang:
@dustymabe Where would be a good place to track this issue?

dustymabe:
we could probably just rename this issue to "systemd presets for units from ignition don't get evaluated" and track it here.

rphillips:
I booted the latest Fedora Atomic Host which contains systemd version 238. The rhcos image has systemd 219 (from ~2015).

dustymabe:
> I booted the latest Fedora Atomic Host which contains systemd version 238. The rhcos image has systemd 219 (from ~2015).

I don't think that's relevant. We are and have been using an EL7 base. This was working and broke recently.

cgwalters:
PR in https://github.com/openshift/os/pull/247

PR #247, opened 2018-08-28
manifest: Explicitly set machineid-compat: false at toplevel

cgwalters:
It looks like the include logic in rpm-ostree isn't handling
this correctly.

Closes: https://github.com/openshift/os/issues/243


dustymabe:
> It looks like the include logic in rpm-ostree isn't handling
> this correctly.

can we open a bug and include it in the comment?

cgwalters:
done :arrow_up:

PR #249, opened 2018-08-28
jenkins: Lower more treecompose logic to shell

cgwalters:
Prep for sharing with ootpa and moving away from groovy.

I like this idea of "scripts/tools generate JSON" - then it's
easy for Jenkins to parse.

### DIFF
--- a/Jenkinsfile.treecompose
+++ b/Jenkinsfile.treecompose
@@ -63,23 +63,12 @@ node(NODE) {
         // Note that we don't keep history in the ostree repo, so we
         // delete the ref head (so rpm-ostree won't add a parent) and
         // then we do a repo prune after.
-        def version, commit
+        def composeMeta;
         try {
-            stage("Compose Tree") { sh """
-                rm ${repo}/refs/heads/* -rf
-                env G_DEBUG=fatal-warnings rpm-ostree compose tree --repo=${repo} \
-                                                            --add-metadata-string=version=${version_prefix}.${env.BUILD_NUMBER} \
-                                                            ${manifest}
-                ostree --repo=${repo} rev-parse ${ref} > commit.txt
-            """
-                commit = readFile('commit.txt').trim();
-                version = utils.get_rev_version("${repo}", commit)
-                currentBuild.description = "🆕 ${version} (commit ${commit})";
-                if (previous_commit.length() > 0) { sh """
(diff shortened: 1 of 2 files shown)


scripts/run-treecompose
```diff
@@ -0,0 +1,39 @@
...
+
+# We don't keep history in the tree.
+previous_commit=$(ostree --repo="${repo}" rev-parse "${ref}" || true)
+rm ${repo}/refs/heads/* -rf
+# This should be the default.  It'll help us trace down a
+# coredump we're seeing periodically.
+export G_DEBUG=fatal-warnings
+rpm-ostree compose tree --repo=${repo} "$@"
+commit=$(ostree --repo=${repo} rev-parse ${ref})
+version=$(ostree --repo=${repo} show --print-metadata-key=version ${commit} | sed -e "s,',,g")
```
miabbott: This works for us, since we add the metadata `version` string when the script is called, but if someone were to reuse this script, this could bomb out.

Probably not a deal-breaker, but thought I would mention it.

Jenkinsfile.treecompose
```diff
@@ -63,23 +63,12 @@ node(NODE) {
...
-                version = utils.get_rev_version("${repo}", commit)
-                currentBuild.description = "🆕 ${version} (commit ${commit})";
-                if (previous_commit.length() > 0) { sh """
-                    rpm-ostree --repo=${repo} db diff ${previous_commit} ${commit} > ${WORKSPACE}/pkg-diff.txt
-                """ }
-                sh "ostree --repo=${repo} prune --refs-only --depth=0"
-                sh "ostree summary --repo=${repo} --update"
+            stage("Compose Tree") {
+                sh """./scripts/run-treecompose ${repo} ${ref} ${manifest} --add-metadata-string=version=${version_prefix}.${env.BUILD_NUMBER}"""
+                composeMeta = readJSON file: "compose.json"
```
miabbott: neat trick!

scripts/run-treecompose
cgwalters: True.  My thoughts on reusability are that this commit is stage 2 - "stage 3" of reusability would be having it in coreos-assembler or so, and at that point we'd hopefully be using a not-bash language and be more flexible.

PR #250, opened 2018-08-28
Add RHCOS kola testing on aws

yuqi-zhang:
Add a new job coreos-rhcos-aws-test to take an ami and run kola tests
on it. If tests pass, the job gives launch permission to Openshift CI
for that ami. Modify cloud job to no longer give permission, but
instead to trigger the new aws-test job. Split the generated aws json
into smoketested and awstested, which signifies successful basic tests
and successful aws kola tests, respectively.

Signed off by: Yu Qi Zhang <jerzhang@redhat.com>

### DIFF
--- a/Jenkinsfile.aws-test
+++ b/Jenkinsfile.aws-test
@@ -1,0 +1,78 @@
+def NODE = "rhcos-jenkins"
+def AWS_REGION = "us-east-1"
+
+// this var conveniently refers to a location on the server as well as the
+// local dir we sync to/from
+def images = "/srv/rhcos/output/images"
+
+node(NODE) {
+    stage("Clean workspace") {
+       step([$class: 'WsCleanup'])
+    }
+    checkout scm
+    utils = load("pipeline-utils.groovy")
+
+    // This job should only be triggered by Jenkinsfile.cloud
+    utils.define_properties(null)
+
+    if (params.DRY_RUN) {
+        echo "DRY_RUN set, skipping aws tests and launch permissions"
+        currentBuild.result = 'NOT_BUILT'
+        currentBuild.description = '(dry run)'
+        return
+    }
+
+    utils.inside_assembler_container("-v /srv:/srv") {
+    stage("Sync In") {
+        withCredentials([
+            string(credentialsId: params.ARTIFACT_SERVER, variable: 'ARTIFACT_SERVER'),
+            sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
+        ]) {
+            sh "mkdir -p ${dirpath}"
+            utils.rsync_file_in(ARTIFACT_SERVER, KEY_FILE, "${dirpath}/aws-${AWS_REGION}-smoketested.json")
+        }
+    }
+
+    try {
+    stage("Run Kola tests on intermediate aws image") {
+            withCredentials([
+                [$class: 'AmazonWebServicesCredentialsBinding', credentialsId: params.AWS_CREDENTIALS],
+                string(credentialsId: params.S3_PRIVATE_BUCKET, variable: 'S3_PRIVATE_BUCKET'),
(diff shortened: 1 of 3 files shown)


yuqi-zhang:
Not sure if this is the best workflow, would require the addition of coreos-rhcos-aws-test jenkins job. Sample jobs can be found at: https://continuous-infra-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/rhcos/job/jerzhang-test-coreos-rhcos-cloud/ and https://continuous-infra-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/rhcos/job/jerzhang-test-coreos-rhcos-aws-test-and-launch/ (tests are passing with https://github.com/openshift/os/pull/247).

yuqi-zhang:
@cgwalters @ashcrow @arithx PTAL

Jenkinsfile.aws-test
```diff
@@ -0,0 +1,73 @@
...
+            string(credentialsId: params.ARTIFACT_SERVER, variable: 'ARTIFACT_SERVER'),
+            sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
+        ]) {
+            sh "mkdir -p ${images}"
+            utils.rsync_dir_in(ARTIFACT_SERVER, KEY_FILE, images)
+        }
+    }
+
+    try {
+    stage("Run Kola tests on intermediate aws image") {
```
miabbott: It would be helpful to set the `currentBuild.description` to the AMI ID near the beginning of this stage.  Kind of like this:

https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L125

yuqi-zhang: Added below!

Jenkinsfile.aws-test
```diff
@@ -0,0 +1,74 @@
...
+                currentBuild.description = "ami: ${ami_intermediate}"
+                sh """
+                    # Do testing with intermediate aws image passed in by cloud job
+                    if ! kola -b rhcos -p aws --aws-type ${aws_type} --tapfile rhcos-aws.tap --aws-ami ${ami_intermediate} --aws-region ${AWS_REGION} run; then
+                        exit 1
+                    fi
+
+                    # Tests pass, tag the json in the artifact server
+                    # and give launch permissions to OpenShift CI
+                    cp ${dirpath}/aws-${AWS_REGION}-smoketested.json ${dirpath}/aws-${AWS_REGION}-awstested.json
```
cgwalters: The problem is that `-awstested.json` will go away fairly quickly as the cloud images churn.

So anyone who wants to find out "what is the latest AMI that passed basic testing" is going to be fighting `404s`.

You know I keep framing every problem we have as "we need tagging" :smile: 

Anyways...a tactical fix here might be to stick `-awstested.json` *outside* of the versioned image directory, i.e. just sync to `${images}/aws-${AWS_REGION}-tested.json`.

arithx: :+1: on keeping the latest successfully tested image around until it's replaced.

yuqi-zhang:
I changed it to just sync the files, but I'm not sure if adding 3 (somewhat) duplicated util functions was the best way to do so.

cgwalters:
> I changed it to just sync the files, but I'm not sure if adding 3 (somewhat) duplicated util functions was the best way to do so.

That seems fine to me, but it doesn't look like you addressed my comment about the destination?
Basically, there's no stable URL for "latest tested AMI json".  If we moved it *outside* of the `images/` directory then there'd be no concurrency issues and it'd have a stable URL.

yuqi-zhang:
Apologies if I'm misunderstanding, but in the other comment you suggested `${images}/aws-${AWS_REGION}-tested.json` as the location of the file, which is where I changed it to. Would you like it to be one level higher? (i.e. `/srv/rhcos/output/aws-${AWS_REGION}-tested.json`). Or did you mean to put `xxx-smoketested.json` there alongside the `xxx-tested.json` file?

cgwalters:
Oops, you're right you did fix it!  I got confused between `${dirpath}` and `${images}`.

PR #252, opened 2018-08-29
Add a job to generate ootpa oscontainer

cgwalters:
The treecompose job works, and one can pivot to it.  I have
more work to do to share code between the now 2 treecompose jobs.
And we're obviously missing cloud images (further patches to come
for that).

There are some SELinux denials and failing services (it looks
like preset is enabling sssd socket units somehow?)

But this is useful for us to add to the matrix and make available.

### DIFF
--- a/Jenkinsfile.treecompose-ootpa
+++ b/Jenkinsfile.treecompose-ootpa
@@ -1,14 +1,9 @@
 def TIMER = "H/30 * * * *"
 def NODE = "rhcos-jenkins"
 def API_CI_REGISTRY = "registry.svc.ci.openshift.org"
-def OS_NAME = "maipo";
+def OS_NAME = "ootpa";
 def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-${OS_NAME}:latest"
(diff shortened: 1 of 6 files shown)


dustymabe:
> But this is useful for us to add to the matrix and make available.

not sure if I agree. I really don't want to have two versions of the pipeline running (if that's not what this is then forgive my mistake) especially in this early stage where everything is so fragile. maybe we can chat about this tomorrow in chat.

dustymabe:
> it only pushes an oscontainer.

if it only pushes an oscontainer and there are no image artifacts that are created (i.e. we don't ever have to ask someone "which RHCOS are you using?") then I'm ok with the change.

cgwalters:
> (i.e. we don't ever have to ask someone "which RHCOS are you using?")

That's right, we won't yet.  But we also need to raise awareness that the need to answer that question is coming - everything built on top should be thinking about how to consume multiple streams.  At least for a while; hopefully we can do a hard cutover, and *ideally* after that move into a much more regular cadence of updates.

ashcrow:
Pulling the trigger

/lgtm

PR #255, opened 2018-08-29
treecompose: Drop some dead code

cgwalters:
- We clean the workspace now
 - The `ls` is clearly just a debug leftover
 - The previous commit is set consistently in the composeMeta json
   generated by the later code (nothing was reading `previous_commit`)


cgwalters:
Ah sorry, yep I updated this with one more bit to drop `commit.txt`

Issue #258, opened 2018-08-30
create a single "source of truth" JSON artifact that holds build metadata


miabbott:
Currently we are generating a `commit.txt` that has the `ostree` commit ID for a build, but we don't have `version.txt` that has the version number.

Additionally, we have a `pkglist.txt` that has a list of packages and a `pkg-diff.txt` that has the package diff between the previous `ostree` commit and the current one.

Finally, there is a JSON file that has the AMI data (ID, S3 object, Snapshot ID).

It looks to me that we could combine some (or all?) of this data into a single JSON that we could then publish to the public as part of the pipeline.

miabbott:
This is one of the items that was discussed as part of https://github.com/openshift/os/issues/150#issuecomment-417094115

cgwalters:
PR in https://github.com/openshift/os/pull/260

cgwalters:
> Finally, there is a JSON file that has the AMI data (ID, S3 object, Snapshot ID).

Mmm...I'm not merging that (yet), but arguably the meta.json should have a reference to it or so?

PR #260, opened 2018-08-30
jenkins/cloud: Change version schema, add+use `meta.json`

cgwalters:
This drops the date stamp, and switches to a version number
scheme `${ostree_version}-${genver}` where `genver` is like
`Release` in RPM terms.

The general principle here is that we make the ostree version
a unifying identifier across the oscontainer, vmimages etc.

Also, we now have more easily parsable JSON in the images directory.

Closes: https://github.com/openshift/os/issues/258

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -50,17 +50,21 @@ node(NODE) {
 
     // Check if there's a new version out.
     // Really, we should be checksumming e.g. the ks and tdl too.
-    def latest_commit, last_build_commit, latest_version, force_nocache
+    def latest_commit, last_build_meta, latest_version, force_nocache
     stage("Check for Changes") {
         latest_commit = utils.sh_capture("ostree --repo=repo rev-parse ${ref}")
         latest_version = utils.get_rev_version("repo", "${latest_commit}")
         if (fileExists('force-nocache-build')) {
             force_nocache = readFile('force-nocache-build').trim();
         }
-        last_build_commit = utils.sh_capture("cat ${images}/cloud/latest/commit.txt || :")
+        def previous_meta_path = "${images}/cloud/latest/meta.json";
+        if (fileExists(previous_meta_path)) {
+            last_build_meta = readJSON file: previous_meta_path;
+        }
     }
 
-    if (!params.DRY_RUN && (latest_commit == last_build_commit && latest_commit != force_nocache)) {
+    def previously_built_commit = (last_build_meta != null && latest_commit == last_build_meta["ostree-commit"]);
+    if (!params.DRY_RUN && previously_built_commit && latest_commit != force_nocache) {
         echo "Last built ${latest_version} (${latest_commit}) - no changes"
         currentBuild.result = 'SUCCESS'
         currentBuild.description = '(No changes)'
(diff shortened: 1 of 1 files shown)


cgwalters:
I did a dry run test of this - I didn't test syncing out the content, and consequently I didn't test that we correctly bump the genver from `-1` to `-2` when doing a rebuild.

cgwalters:
> src=/srv/rhcos/output/images/cloud/null/rhcos-4.0.4883.qcow2

Hmm, nope.  Fixing.

cgwalters:
OK, now with 100% less `null` versions.

cgwalters:
One thing I struggle with in this is whether we really want rdgo/treecompose/cloud to be async.  It feels sort of like they should be async just for *buildmaster* - we want to be able to download the latest git master of some component if treecompose is failing, and we want to be able to quickly test an oscontainer update without blocking on AWS uploads.

But maybe what we should be doing for "smoketested" is gathering all of those things together.  So there'd be no `image-genver` for smoketested and below?

yuqi-zhang:
Trying on my test pipeline, so far so good.

I'm a little confused what you mean by gathering all of them together. Do you want to cloud job to effectively consolidate one version of rdgo/treecompose and bundle that as a "image-genver"? As in the image-genver is implying everything was successful up to smoketesting?

Jenkinsfile.cloud
```diff
@@ -50,17 +50,21 @@ node(NODE) {
...
         }
-        last_build_commit = utils.sh_capture("cat ${images}/cloud/latest/commit.txt || :")
+        def previous_meta_path = "${images}/cloud/latest/meta.json";
+        if (fileExists(previous_meta_path)) {
+            last_build_meta = readJSON file: previous_meta_path;
+        }
     }
 
-    if (!params.DRY_RUN && (latest_commit == last_build_commit && latest_commit != force_nocache)) {
+    def previously_built_commit = (last_build_meta != null && latest_commit == last_build_meta["commit"]);
```
yuqi-zhang: Shouldn't this be last_build_meta["ostree-commit"]

cgwalters: Yep good catch!  Fixed :arrow_down:

cgwalters:
> Would it be worthwhile to add a timestamp to the meta file for when the build was ran?

Definitely but...I'd like to tackle stuff like having consistent metadata and things as part of enhancing the assembler container flow.

arithx:
> > Would it be worthwhile to add a timestamp to the meta file for when the build was ran?
>
> Definitely but...I'd like to tackle stuff like having consistent metadata and things as part of enhancing the assembler container flow.

Fair enough, I don't think it's super necessary for this PR. Otherwise changes LGTM

arithx:
> (Another thing I started to add in was the output of git describe --tags --always --abbrev=42 but decided to defer to the assembler)

Isn't that present in `meta["git-commit"]`

cgwalters:
> Isn't that present in meta["git-commit"]

Oh hey apparently I didn't `git stash` that :smile:

yuqi-zhang:
@cgwalters yes, sorry, I thought you wanted to remove the git-commit.

Either way LGTM. Not sure if I have perms for the bot to merge?

cgwalters:
> Either way LGTM. Not sure if I have perms for the bot to merge?

You should, let's try!  If you don't we'll fix it.

PR #264, opened 2018-08-31
host: add coreos-metadata binary and service

lucab:
This adds coreos-metadata RPM to RHCOS treefile. Current package is for
latest upstream version (v3.0.0) and provides the binary and a
service unit (disabled). The ssh-keys service is not relevant to RHCOS
and not installed in the rpm.

This consumes a dedicated copr repo:
https://copr.fedorainfracloud.org/coprs/lucab/rhcos-coreos-metadata/

Source specfile is at https://github.com/lucab/rhcos-coreos-metadata

Ref: COREOS-406


dustymabe:
thanks @lucab !

lucab:
For the short term I can keep both repo under my namespace. For the long term, when we switch to newer content (RHEL8/Fedora) we can drop all those workarounds and (hopefully) move them it upstream.

PR #245, opened 2018-08-28
rhel.repo: Make use of distribution-gpg-keys

cgwalters:
See https://github.com/coreos/coreos-assembler/pull/23

PR #267, opened 2018-09-04
Drop tuned

cgwalters:
Helps move us towards not shipping python.  We also generally
want our configuration to be static - we don't want admins interactively
running tuned in general.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,6 @@ rdgo:
 # Composes an ostree
 .PHONY: rpmostree-compose
 rpmostree-compose: ${ROOT_DIR}/openshift.repo init-ostree-repo
-	cp -n ${ROOT_DIR}/RPM-GPG-* /etc/pki/rpm-gpg/
 	mkdir -p ${REPO}
 	ostree init --repo=${REPO} --mode=archive
 	if test -d cache; then cachedir='--cachedir $(shell pwd)/cache'; fi && \
--- a/host-maipo.yaml
+++ b/host-maipo.yaml
@@ -22,8 +22,6 @@ packages:
  # Docker legacy
  - docker docker-lvm-plugin docker-novolume-plugin
  - oci-umount
- # tuned (legacy IMO)
- - tuned tuned-profiles-atomic
  # SELinux
  - policycoreutils-python
  # Networking legacy


jeremyeder:
tuned in the base OS is being replaced in 4.0 by the node tuning operator, which is a  containerized tuned that runs as a daemonset.

/cc @jmencak for visibility (he is the developer of this new operator).

cgwalters:
> tuned in the base OS is being replaced in 4.0 by the node tuning operator, which is a containerized tuned that runs as a daemonset.

OK, sounds like a step forward - though we're going to have some issues reconciling dynamic privileged host mutation like that with the "immutable infrastructure" story promoted by the [machine config operator](https://github.com/openshift/machine-config-operator).

cgwalters:
For reference:

```
[root@vanadium rhcos]# rpm-ostree --repo=repo db diff openshift/3.10/x86_64/os{^,}
ostree diff commit old: openshift/3.10/x86_64/os^ (d128349ab95fa3dfd4e8f602b418a972d8dc9a37eec5f89eeb5c8305437ec17e)
ostree diff commit new: openshift/3.10/x86_64/os (fe94141df483a37b58dbfbed46a98f4669498c8c88967ac93c5e8a5cb25f74da)
Removed:
  dbus-glib-0.100-7.el7.x86_64
  dbus-python-1.1.1-9.el7.x86_64
  dmidecode-1:3.0-5.el7.x86_64
  ethtool-2:4.8-7.el7.x86_64
  gobject-introspection-1.50.0-1.el7.x86_64
  python-configobj-4.7.2-7.el7.noarch
  python-decorator-3.4.0-3.el7.noarch
  python-gobject-base-3.22.0-1.el7_4.1.x86_64
  python-linux-procfs-0.4.9-3.el7.noarch
  python-perf-3.10.0-862.2.3.el7.x86_64
  python-pyudev-0.15-9.el7.noarch
  python-schedutils-0.4-6.el7.x86_64
  tuned-2.9.0-1.el7.noarch
  tuned-profiles-atomic-2.9.0-1.el7.noarch
  virt-what-1.18-4.el7.x86_64
```

jeremyeder:
Thanks @cgwalters who do we need to discuss that with?

cgwalters:
> that python was going away

FWIW https://github.com/coreos/fedora-coreos-tracker/issues/32

Makefile
```diff
@@ -57,7 +57,6 @@ rdgo:
 # Composes an ostree
 .PHONY: rpmostree-compose
 rpmostree-compose: ${ROOT_DIR}/openshift.repo init-ostree-repo
-	cp -n ${ROOT_DIR}/RPM-GPG-* /etc/pki/rpm-gpg/
```
ashcrow: Dropped because we are using hosts keys during composes correct?

host-maipo.yaml
```diff
@@ -22,8 +22,6 @@ packages:
  # Docker legacy
  - docker docker-lvm-plugin docker-novolume-plugin
  - oci-umount
- # tuned (legacy IMO)
```
ashcrow: :+1:

Makefile
cgwalters: Yeah.

cgwalters: Since https://github.com/openshift/os/pull/245

cgwalters:
Hmm...how does `/approve` differ from `/lgtm`?

cgwalters:
Maybe part of the issue is this PR isn't yet rebased to include the `OWNERS` file?

ashcrow:
@cgwalters Could be that it's not in this PR. Will merge it :-)

PR #270, opened 2018-09-04
rhel.repo: Switch to 7Server

### LINKED ISSUE #269: podman is too old in our composes
@mheon noted that the version of `podman` in our composes is multiple months out of date. We should ship the latest release builds from the `podman` team.

ashcrow:
Switches repos from `7.5` :arrow_right: `7Server`.

Closed #269

### DIFF
--- a/rhel.repo
+++ b/rhel.repo
@@ -1,27 +1,27 @@
 [rhel-7.5-server]
 enabled=1
-baseurl=http://cdn.stage.redhat.com/content/dist/rhel/server/7/7.5/x86_64/os/
+baseurl=http://cdn.stage.redhat.com/content/dist/rhel/server/7/7Server/x86_64/os/
 metadata_expire=1m
 gpgcheck=1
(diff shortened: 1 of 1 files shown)


cgwalters:
It looks like `7Server` is a symlink/alias to one of `7.5`, `7.6` etc?  So we're effectively fixing this by switching to pre-release which we should at least call out here.  There's likely a big pkgdiff.

Hmm...as best I can tell on the CDN side we don't really offer versioned extras repos since only the latest is supported.  Which makes this messier.

miabbott:
FWIW, on a freshly subbed RHELAH host, we have use the following prod URLS:

https://cdn.redhat.com/content/dist/rhel/server/7/$releasever/$basearch/os

https://cdn.redhat.com/content/dist/rhel/server/7/7Server/$basearch/extras/os

https://cdn.redhat.com/content/dist/rhel/server/7/$releasever/$basearch/optional/os

https://cdn.redhat.com/content/dist/rhel/atomic/7/7Server/$basearch/os

PR #278, opened 2018-09-06
Rename maipo repos to use that name

cgwalters:
For consistency with ootpa.  Also having a version number in
the repo names now was deceptive as we now track "latest".


cgwalters:
Now with 100% more `git add`.

PR #280, opened 2018-09-06
ci: Also syntax check YAML files

jlebon:
Now that we mostly work in YAML files rather than JSON, let's adapt our
checks accordingly.

### DIFF
--- a/.prow.sh
+++ b/.prow.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -xeuo pipefail
 
-yum -y install jq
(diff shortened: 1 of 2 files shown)


cgwalters:
(Side note, any reason not to use the assembler container here?)

jlebon:
> (Side note, any reason not to use the assembler container here?)

Sure, that shouldn't be too hard:

https://github.com/openshift/release/pull/1381
https://github.com/openshift/os/pull/282

PR #283, opened 2018-09-11
Drop docker from compose

ashcrow:
Requires:
- [x] https://github.com/openshift/installer/pull/221 
- [x] https://github.com/openshift/installer/pull/207

### DIFF
--- a/host-maipo.yaml
+++ b/host-maipo.yaml
@@ -21,9 +21,6 @@ mutate-os-release: "4.0"
 packages:
  # auth legacy
  - authconfig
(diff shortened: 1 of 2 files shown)


ashcrow:
Required PRs have been merged. PTAL.

jlebon:
That's awesome!
To confirm, this will break `oc cluster up` for now, right?

ashcrow:
@jlebon from my understanding, yes. It also will break other systems which previously hard assumed `docker` was the only container runtime available.

PR #290, opened 2018-09-17
Revert revert: Remove docker

ashcrow:
Go back to no docker.

### DIFF
--- a/host-maipo.yaml
+++ b/host-maipo.yaml
@@ -22,9 +22,6 @@ mutate-os-release: "4.0"
 packages:
  # auth legacy
  - authconfig
- # Docker legacy
- - docker docker-lvm-plugin docker-novolume-plugin
- - oci-umount
  # SELinux
  - policycoreutils-python
  # Networking legacy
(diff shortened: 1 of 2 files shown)


cgwalters:
OK by me, though for things like this in the future it may be good to add an installer change first that removes it (`if [ -x /usr/bin/docker ]; then ostree admin unlock && rm /usr/bin/docker; fi` or so early in ignition) - then once that lands successfully we can confidently make a change like this.

/lgtm

wking:
> ... for things like this in the future it may be good to add an installer change first that removes it (`if [ -x /usr/bin/docker ]; then ostree admin unlock && rm /usr/bin/docker; fi` or so early in ignition) - then once that lands successfully we can confidently make a change like this.

This would have missed the resizing thingie that Docker was giving us (#293), wouldn't it?  Ideally there would be either CI here launching a cluster with the latest tested installer and the new RHCOS candidate.  Maybe with allowed failures for now, so we could break out of deadlocks when we wanted to push through breaking changes.  I don't know how hard that would be to set up, though.

Issue #293, opened 2018-09-17
Backport  "no lvm and custom growpart"


cgwalters:
It turns out `docker.rpm` bizarrely contains the systemd unit `docker-storage-setup.service`.  When we dropped docker we stopped getting disk resizing too.

Let's plow forwards and backport https://github.com/coreos/fedora-coreos-config/pull/8

ashcrow:
:+1: SGTM

PR #294, opened 2018-09-17
Drop LVM, use custom growpart wrapper

cgwalters:
Closes: https://github.com/openshift/os/issues/293


cgwalters:
Requires: https://github.com/coreos/coreos-assembler/pull/95

cgwalters:
Actually unless we backport https://github.com/coreos/fedora-coreos-config/commit/0fdb511dbcb35b0472d909959b6346c915ae517c the commits interlock - pushing one without the other will be broken.

compose-post.sh
```diff
@@ -9,6 +9,38 @@ ln -sfr /usr/lib/systemd/system/{multi-user,default}.target
...
+touch /var/lib/coreos-growpart.stamp
+EOF
+chmod a+x /usr/libexec/coreos-growpart
+cat > /usr/lib/systemd/system/coreos-growpart.service <<'EOF'
+[Unit]
+ConditionPathExists=!/var/lib/coreos-growpart.stamp
+[Service]
+ExecStart=/usr/libexec/coreos-growpart /
+RemainAfterExit=yes
+Before=sshd.service
```
dm0-: Isn't `Before=` required to be under the `[Unit]` section?

cgwalters: Apparently it's just ignored.

dm0-: Okay, I've submitted coreos/fedora-coreos-config#10.

cgwalters: https://github.com/coreos/fedora-coreos-config/pull/11

And updated below :arrow_down:

cgwalters:
> @cgwalters Can that be backported in this PR set or do we need to keep them separate?

Wouldn't hurt but wouldn't help either - the problem is once we build a new `coreos-assembler:alpha`.  But we can just wait to do that until this merges.

Issue #298, opened 2018-09-17
latest failing in ec2


cgwalters:
```
[    7.182589] ignition[2018]: INFO     : GET error: Get http://169.254.169.254/2009-04-04/user-data: dial tcp 169.254.169.254:80: connect: network is unreachable
```

until we time out.  The qcow2 boots fine for me with `coreos-assembler run -d /path/to/rhcos-4.0.5743-qemu.qcow2`, and I have networking post-boot.

Maybe something related to new Ignition?

cgwalters:
Some sort of patch for Ignition to print e.g. `ip addr && ip route` or something when it fails to finally fetch would help.

cgwalters:
One notable thing though is before, I see: `Sep 17 20:41:20 localhost dracut-initqueue[700]: dhcp: PREINIT eth0 up`.  But not with the new build.

yuqi-zhang:
Not sure if the same cause, but we don't make it past the basic tests:
```
--- FAIL: rhcos.basic (52.72s)
        harness.go:434: Cluster failed starting machines: machine "c3d40f67-42b7-492f-9e71 4464d8f502e9" failed basic checks: some systemd units failed:

crio.service loaded failed failed Open Container Initiative Daemon
```

cgwalters:
I went back and looked through the pkg diffs and there's nothing obvious that should/could cause this.  No dracut/kernel change; ignition did change but only after things started failing.

dustymabe:
I should have @eparis unblocked for now by telling him `4.0.5616` worked for me this weekend (this was pre-docker removal` and new ignition) we can fix this tomorrow I think.

cgwalters:
I have confirmed this is somehow related to LVM, if I apply:

```
-part / --size=3000 --fstype="xfs" --label=root --grow
+part pv.01 --grow
+volgroup coreos pv.01
+logvol / --size=3000 --fstype="xfs" --name=root --label=root --vgname=coreos
```

I again see the kernel doing DHCP.

Here's a theory; the `rd.lvm.lv=` kernel argument that Anaconda injects is somehow causing dracut to spend longer or go in a different codepath in the initramfs.

cgwalters:
Also of course worth noting that I did test this setup in Fedora 29, so this is presumably specific to maipo.

dustymabe:
[...]

PR #299, opened 2018-09-18
Switch back to LVM+container-storage-setup

cgwalters:
Dropping LVM caused race conditions in dracut startup - I *think*
it's the fact that the LVM bits basically process the whole udev
queue to assemble devices, there's a `udevadm settle` in that code.

Our dracut probably needs to learn to settle/wait for network
devices with `ip=dhcp`.

Closes: https://github.com/openshift/os/issues/298


cgwalters:
One thing I will note about this though is that the `GROWPART=true` now lives in the ostree, not the kickstart - doesn't matter in practice for us today but gets us closer to the FCOS model of "image installs".

cgwalters:
> /hold - i have a potential other solution - will open PR shortly to discuss

OK, though I did test this one.  I have to run an errand but will be back in a bit.

dustymabe:
see #300

cgwalters:
This is obsoleted.

PR #301, opened 2018-09-18
run-treecompose: stick pkg diff into JSON

miabbott:


miabbott:
Hmm...looking over at [coreos/coreos-assembler](https://github.com/coreos/coreos-assembler), maybe these kinds of changes should land over there.

ashcrow:
> Hmm...looking over at coreos/coreos-assembler, maybe these kinds of changes should land over there.
/cc @cgwalters

cgwalters:
> Hmm...looking over at coreos/coreos-assembler, maybe these kinds of changes should land over there.

Well yeah, we already did that in https://github.com/coreos/coreos-assembler/pull/59

miabbott:
> Well yeah, we already did that in coreos/coreos-assembler#59

Right...that's still missing the package diff which this is providing.  Anyways, I'll make that change over there.

cgwalters:
> Right...that's still missing the package diff which this is providing. 

That's in the rpm-ostree json right?

miabbott:
> That's in the rpm-ostree json right?

picard-facepalm

Yeah...

/me sighs

cgwalters:
To be fair it's going to be a little bit until we can deploy assembler...it can still make sense to do fixes to the existing one, but we have to measure the cost/benefit of changing it.

PR #304, opened 2018-09-18
jenkins: Make RHCOS AMI Public

kikisdeliveryservice:
This is my first PR, so I'm a bit unsure about the format of the PR, etc... If you have any feedback let me know! :)

### DIFF
--- a/Jenkinsfile.aws-test
+++ b/Jenkinsfile.aws-test
@@ -60,9 +60,6 @@ node(NODE) {
                     cp ${dirpath}/aws-${AWS_REGION}-smoketested.json ${images}/aws-${AWS_REGION}-tested.json
                     aws ec2 modify-image-attribute \
                         --image-id ${ami_intermediate} \
-                        --launch-permission '{"Add":[{"UserId":"${AWS_CI_ACCOUNT}"}]}'
-                    aws ec2 modify-image-attribute \
-                        --image-id ${ami_intermediate} \
                         --launch-permission '{"Add":[{"Group":"all"}]}'
                 """
             }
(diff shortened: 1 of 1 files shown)


miabbott:
You are on the right track, but I think we should be making the AMI public after it passes our `kola` tests. That happens over here - https://github.com/openshift/os/blob/master/Jenkinsfile.aws-test#L61-L63

kikisdeliveryservice:
Thanks for the feedback @miabbott ! Will update & repush.

ashcrow:
Good catch!

Jenkinsfile.aws-test
```diff
@@ -61,6 +61,9 @@ node(NODE) {
                     aws ec2 modify-image-attribute \
                         --image-id ${ami_intermediate} \
                         --launch-permission '{"Add":[{"UserId":"${AWS_CI_ACCOUNT}"}]}'
+                    aws ec2 modify-image-attribute \
+                        --image-id ${ami_intermediate} \
+                        --launch-permission "Add=[{Group=all}]
```
cgwalters: We can drop the previous invocation of `modify-image-attribute` since now anyone can access right?

cgwalters: (Or alternatively think of this as just changing the argument to `--launch-permission)

jlebon: Quoting looks a bit off -- I'm guessing it should be `'{"Add":[{"Group":"all"}]}'` ?

ashcrow: I agree. I think @jlebon has the right idea on the quoting.

```
--launch-permission "Add=[{Group=all}]
```

becomes

```
--launch-permission '{"Add":[{"Group":"all"}]}'
```

ashcrow: > We can drop the previous invocation of modify-image-attribute since now anyone can access right?

Yup! We shouldn't need to do that any longer. @kikisdeliveryservice If you don't mind, add another commit in this PR to remove that section.

kikisdeliveryservice: Thanks @ashcrow @jlebon @cgwalters ! I'll make those changes and repush.

PR #308, opened 2018-09-21
adopt `buildmaster` and `alpha` tags for oscontainer

miabbott:
This pile of commits attempts to implement some of the versioning scheme for the `oscontainer` discussed in #201.

Firstly, during the `treecompose` pipeline, the `oscontainer` is no longer tagged with `latest`, but with `buildmaster`.  Additionally, it is tagged with the `ostree` commit that was used to generate the container image.

This `ostree` commit is passed to the `aws-test` pipeline after a successful `cloud` pipeline run.  In the `aws-test` pipeline, the `oscontainer` is pulled by the `ostree` commit and if the AWS tests are successful, the `oscontainer` is tagged with `alpha` and pushed to the registry.

This is probably not the best way to do this kind of image promotion and will likely be obsoleted by `coreos-assembler`, but I figured I would still put this out there to see how it was received.

### DIFF
--- a/Jenkinsfile.aws-test
+++ b/Jenkinsfile.aws-test
@@ -1,5 +1,8 @@
 def NODE = "rhcos-jenkins"
 def AWS_REGION = "us-east-1"
+def API_CI_REGISTRY = "registry.svc.ci.openshift.org"
+def OS_NAME = "maipo";
+def OSCONTAINER_IMG = API_CI_REGISTRY + "/rhcos/os-${OS_NAME}"
 
 // location on the server we'll rsync to/from our $WORKSPACE
 def images = "/srv/rhcos/output/images"
@@ -23,7 +26,7 @@ node(NODE) {
 
     // We're only ever triggered by the cloud job, so we know the latest build is in latest/
     // We immediately resolve it back to the specific images/ dir
-    def version
+    def ostree_commit, version
     try {
     utils.inside_assembler_container("") {
     stage("Sync In") {
@@ -32,6 +35,7 @@ node(NODE) {
             sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
         ]) {
             utils.rsync_file_in_dest(ARTIFACT_SERVER, KEY_FILE, "${images}/cloud/latest/meta.json", "${WORKSPACE}/meta.json")
+            ostree_commit = utils.sh_capture("jq -r '.[\"ostree-commit\"]' ${WORKSPACE}/meta.json")
             version = utils.sh_capture("jq -r '.[\"ostree-version\"]' ${WORKSPACE}/meta.json")
             // resolve to original dir to avoid races in the next rsync in
             def imgv = utils.sh_capture("jq -r '.[\"image-version\"]' ${WORKSPACE}/meta.json")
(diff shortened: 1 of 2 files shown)


cgwalters:
At a high level...I didn't know you were working on this, I probably would have asked nicely for work to focus on the assembler :smile:   But - improving the current pipeline is useful too!  We just have to carefully weigh the risk/rewards.

Let's split out the two prep patches to the util API, they look clean and we can merge them now.

miabbott:
> The primary downside of this is that today, the api.ci cluster will be garbage collecting untagged oscontainers. If we start pushing ostree-commits-as-tags, we'll need to implement some GC policy ourselves.

I think we could implement this trivially by just using `skopeo delete` on the images tagged with an ostree commit that didn't pass AWS tests, no?

> At a high level...I didn't know you were working on this, I probably would have asked nicely for work to focus on the assembler :smile: But - improving the current pipeline is useful too! We just have to carefully weigh the risk/rewards.

Yeah, I was just tinkering in this area and wanted to measure the reception.  I could try to fit this into the assembler, but I guess I'm still not clear on everything the assembler would be responsible for and how to work in pieces like this.

I'm comfortable with sending the majority of this PR to `/dev/null`

> Let's split out the two prep patches to the util API, they look clean and we can merge them now.

Will do.

cgwalters:
> I think we could implement this trivially by just using skopeo delete on the images tagged with an ostree commit that didn't pass AWS tests, no?

Yeah that'd help a lot.

> Yeah, I was just tinkering in this area and wanted to measure the reception. I could try to fit this into the assembler, but I guess I'm still not clear on everything the assembler would be responsible for and how to work in pieces like this.

Right, that's a big topic.  I'm still trying to land the oscontainer basics there along with some work around how the images are sync'd.

miabbott:
> Right, that's a big topic. I'm still trying to land the oscontainer basics there along with some work around how the images are sync'd.

So let's hold off on this PR for now, with the expectation that we'll be able to do more intelligent tagging/promotion once `coreos-assembler` is more mature.  

It's not like there is any tooling here or on the OpenShift side that is waiting for a `buildmaster` or `alpha` tag.

miabbott:
> So let's hold off on this PR for now, with the expectation that we'll be able to do more intelligent tagging/promotion once coreos-assembler is more mature.

I chatted with @cgwalters and we decided to give this PR a shot, but first I need to try to implement the `skopeo delete` garbage collection mentioned above.

I've split out the utility functions in to a separate PR (#309)

miabbott:
The commits on this PR are kind of a mess right now because I rebased on top of #312 and force pushed.

However, I think the changes in #312 will simplify some of what I'm doing here, so I want to wait for that to land before trying to get this reviewed/merged again.

miabbott:
I'm pretty sure this should work; though kind of hard to test since it involves manipulating the production registry.

miabbott:
Rebased :arrow_up:

Jenkinsfile.treecompose
```diff
@@ -102,8 +101,10 @@ node(NODE) {
         }
 
         stage("Push container") { sh """
-            podman push ${OSCONTAINER_IMG}
-            skopeo inspect docker://${OSCONTAINER_IMG} | jq '.Digest' > imgid.txt
+            podman push ${OSCONTAINER_IMG}:buildmaster
+            podman push ${OSCONTAINER_IMG}:${composeMeta.commit}
+            skopeo inspect docker://${OSCONTAINER_IMG}:buildmaster | jq '.Digest' > imgid.txt
+            podman inspect --format='{{.Id}}' ${OSCONTAINER_IMG}:buildmaster > imgid.txt
```
cgwalters: Duplicate writes to `imgid.txt`?

Jenkinsfile.aws-test
```diff
@@ -32,6 +35,7 @@ node(NODE) {
             sshUserPrivateKey(credentialsId: params.ARTIFACT_SSH_CREDS_ID, keyFileVariable: 'KEY_FILE'),
         ]) {
             utils.rsync_file_in_dest(ARTIFACT_SERVER, KEY_FILE, "${images}/cloud/latest/meta.json", "${WORKSPACE}/meta.json")
+            ostree_commit = utils.sh_capture("jq -r '.[\"ostree-commit\"]' ${WORKSPACE}/meta.json")
```
cgwalters: This isn't new (you're building on code right below), but probably we should just use `readJSON` rather than calling out to `jq` - Groovy is a **real** programming language.

Jenkinsfile.aws-test
```diff
@@ -71,6 +84,12 @@ node(NODE) {
                     aws s3 cp --acl public-read \
                         ${WORKSPACE}/aws-${AWS_REGION}.json \
                         s3://${S3_PUBLIC_BUCKET}/aws-${AWS_REGION}-tested.json
+
+                    # Tag the container image to alpha, then GC the image tagged with the ostree commit
+                    podman pull ${OSCONTAINER_IMG}:${ostree_commit}
+                    podman tag ${OSCONTAINER_IMG}:${ostree_commit} ${OSCONTAINER_IMG}:alpha
+                    podman push ${OSCONTAINER_IMG}:alpha
```
cgwalters: We don't actually need to pull locally here, we can just `skopeo copy` right?

Jenkinsfile.treecompose
miabbott: Ugh, screwed up my rebase.

Jenkinsfile.aws-test
miabbott: Okay, I'll include another commit to use that in this area.

Jenkinsfile.aws-test
miabbott: Ooh, fancy!

miabbott:
New commits incorporating feedback from @cgwalters :arrow_up:

PR #309, opened 2018-09-21
pipeline-utils:  add `registry_login()` and `prep_container_storage()`

miabbott:
This adds two utility functions for logging into a container registry and setting up container storage inside a container.  The latter allows us to pull container images from inside the pipeline container.

Split out from #308

### DIFF
--- a/pipeline-utils.groovy
+++ b/pipeline-utils.groovy
@@ -148,16 +148,15 @@ def registry_login(username, password, registry) {
 // store containers, and bind mounts it to '/var/lib/containers`
 def prep_container_storage(dirFromHost) {
     sh """
-        container_storage="/var/lib/containers"
-        dest="${dirFromHost}/containers"
-        fstype=$(df -P ${dirFromHost} | awk 'END{print $6}' | xargs findmnt -n -o FSTYPE)
-        if [ "$fstype" == "overlay" ]; then
-            echo "Must supply non-overlay location"
+        container_storage=/var/lib/containers
+        fstype=\$(df -P ${dirFromHost} | awk 'END{print \$6}' | xargs findmnt -n -o FSTYPE)
+        if [ \$fstype == 'overlay' ]; then
+            echo 'Must supply non-overlay location'
             exit 1
         fi
-        rm -rf ${container_storage} && mkdir -p ${container_storage}
-        rm -rf ${dest} && mkdir -p ${dest}
-        mount --bind ${dest} ${container_storage}
+        rm -rf \${container_storage} && mkdir -p \${container_storage}
+        rm -rf ${dirFromHost} && mkdir -p ${dirFromHost}/containers
+        mount --bind ${dirFromHost}/containers \${container_storage}
     """
 }
 
--- a/pipeline-utils.groovy
+++ b/pipeline-utils.groovy
@@ -139,6 +139,28 @@ def sh_capture(cmd) {
     return sh(returnStdout: true, script: cmd).trim()
 }
 
+def registry_login(username, password, registry) {
+    sh "podman login -u '${username}' -p '${password}' ${registry}"
+}
+
+// re-implementation of some functionality from scripts/pull-mount-oscontainer
+// takes a directory mounted in from the host, creates a new location to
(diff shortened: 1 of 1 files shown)


pipeline-utils.groovy
```diff
@@ -139,6 +139,21 @@ def sh_capture(cmd) {
...
 }
 
+def registry_login(username, password, registry) {
+    sh "podman login -u '${username}' -p '${password}' ${registry}"
+}
+
+// re-implementation of some functionality from scripts/pull-mount-oscontainer
+// sets up local container storage so the pipeline can pull container images
+// from inside another container
+def prep_container_storage(hostStorage, localStorage) {
```
miabbott: Hmm...maybe I chose the wrong names here...maybe they should be `oldStorage` and `newStorage`?

ashcrow: How about using `hostStorage` and `containerStorage`? That may be a bit clearer.

miabbott: I think `hostStorage` was a mistake on my part.  

As far as I can tell, the host container storage isn't ever mounted into the assembler container (as of now), so we are really just deleting the "typical" location of the container storage in the container.  

For example,`scripts/pull-mount-oscontainer` deletes `/var/lib/containers` in its container, then bind mounts something like `$WORKSPACE/containers` to `/var/lib/containers`.  https://github.com/openshift/os/blob/master/scripts/pull-mount-oscontainer#L21-L23)

I think something more accurate would be `oldContainerStorage` and `newContainerStorage`

What do you think?

cgwalters: I actually discovered that podman/buildah support a `--root` argument and have been using that in https://github.com/coreos/coreos-assembler/pull/90

miabbott:
I'd like to improve this slightly to verify the proposed mount point isn't an overlay filesystem

pipeline-utils.groovy
```diff
@@ -139,6 +139,27 @@ def sh_capture(cmd) {
     return sh(returnStdout: true, script: cmd).trim()
 }
 
+def registry_login(username, password, registry) {
+    sh "podman login -u '${username}' -p '${password}' ${registry}"
```
ashcrow: Note: I believe this will populate `${XDG_RUNTIME_DIR}/containers/auth.json` (EG: `/run/user/1000/containers/auth.json`). Let's make sure this file doesn't get shown via the output or links.

pipeline-utils.groovy
```diff
@@ -139,6 +139,27 @@ def sh_capture(cmd) {
...
+// store containers, and bind mounts it to '/var/lib/containers`
+def prep_container_storage(dirFromHost) {
+    sh """
+        container_storage=/var/lib/containers
+        fstype=\$(df -P ${dirFromHost} | awk 'END{print \$6}' | xargs findmnt -n -o FSTYPE)
+        if [ \$fstype == 'overlay' ]; then
+            echo 'Must supply non-overlay location'
+            exit 1
+        fi
+        rm -rf \$container_storage && mkdir -p \$container_storage
```
yuqi-zhang: Minor nit, `\${container_storage}` is more consistent

miabbott: That will work for shell variables?

pipeline-utils.groovy
miabbott: That *should* be outside the `$WORKSPACE`, so the JSON shouldn't get archived anywhere.  And there's no output from the `podman login` except `Login Succeeded!`

For example, here - https://github.com/openshift/os/blob/1143fc94b8d9f70dcace9d3845053fc5557c57e0/scripts/pull-mount-oscontainer#L10

...looks like this from the pipeline logs

```
+ set +x
podman login -u unused -p <password> registry.svc.ci.openshift.org
Login Succeeded!
```

pipeline-utils.groovy
yuqi-zhang: Yes, e.g. there's a lot of uses of it here: https://github.com/openshift/os/blob/master/Jenkinsfile.cloud#L218-L239

miabbott:
@yuqi-zhang pushed a new commit with your suggested changes

PR #310, opened 2018-09-24
jenkins: List OS version in AMI test job

cgwalters:
Otherwise one has to back up to the cloud job to extract it when
browsing, or drop to the AWS API to list the tags.

### DIFF
--- a/Jenkinsfile.cloud
+++ b/Jenkinsfile.cloud
@@ -253,6 +253,7 @@ node(NODE) {
     def ami_intermediate = utils.sh_capture("jq -r .HVM ${dirpath}/aws-${AWS_REGION}-smoketested.json")
     build job: 'coreos-rhcos-aws-test', wait: false, parameters: [
         string(name: 'ami_intermediate', value: "${ami_intermediate}"),
+        string(name: 'version', value: "${version}"),
         string(name: 'dirpath', value: "${dirpath}"),
         string(name: 'aws_type', value: "t2.small")
     ]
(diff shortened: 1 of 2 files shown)


cgwalters:
(untested)

Jenkinsfile.cloud
```diff
@@ -253,6 +253,7 @@ node(NODE) {
     def ami_intermediate = utils.sh_capture("jq -r .HVM ${dirpath}/aws-${AWS_REGION}-smoketested.json")
     build job: 'coreos-rhcos-aws-test', wait: false, parameters: [
         string(name: 'ami_intermediate', value: "${ami_intermediate}"),
+        string(name: 'version', value: "${version}"),
```
dm0-: It looks like these parameters aren't defined in the AWS job's properties.  Is the server logging security warnings about them?

cgwalters: Offhand I don't see any, though I don't know the specific place to look for that.

dm0-: On the Container Linux Jenkins masters, they'd be in the `jenkins.service` journal output.

cgwalters: I don't see any warnings of that form.

jlebon: It looks like we're running with `-Dhudson.model.ParametersAction.keepUndefinedParameters=true`, which is... not good. We'd have to ask the other tenants of this Jenkins why that is; I distinctly remember having to deal with this in the past. Though I'd agree it'd be better to explicitly add them as parameters.

jlebon: So, I can turn this on in the next master restart. It looks like this is the only job that would need fixing, so we can adapt it first and then turn it on.

PR #315, opened 2018-09-25
maipo: Revert to released content

cgwalters:
Per discussion.


miabbott:
Aside:  looking at the most recent `treecompose`, we aren't pulling any packages from the `maipo-atomic` repo.  Could we ditch that repo?

PR #316, opened 2018-09-26
host: Switch to inline `postprocess` element

cgwalters:
The primary reason I'm doing this is that today rpm-ostree's
treefile inheritance doesn't quite do what one wants with external
files.  There's some major work on that in
https://github.com/projectatomic/rpm-ostree/pull/1574
But let's not block on/require it.

Furthermore, using the inline `postprocess` value allows us
to clearly demarcate the `maipo`-specific hacks in that file.


host-base.yaml
```diff
@@ -13,7 +13,89 @@ initramfs-args:
...
+     # Let's have a non-boring motd, just like CL (although theirs is more subdued
+     # nowadays compared to early versions with ASCII art).  One thing we do here
+     # is add --- as a "separator"; the idea is that any "dynamic" information should
+     # be below that.
+     cat > /etc/motd <<EOF
+     Red Hat CoreOS
+      Information: https://url.corp.redhat.com/redhat-coreos
+      Bugs: https://github.com/openshift/os
+     ---
+     EOF
```
dustymabe: in the old file there was an extra newline between the `---` and the `EOF`. Don't know if we want to keep that or not or if it's possible with inline yaml

dustymabe:
did a vimdiff between the old file and new ones. one small comment.  LGTM

host-base.yaml
cgwalters: Now with :new: newline :arrow_down:

jlebon: This is slight bikeshed but... why not keep the newline separators in the actual script itself too? Makes it much nicer to read through it. At least `PyYAML` definitely supports that. I'd be surprised if serde didn't.

PR #317, opened 2018-09-26
manifest: Drop duplicates from host-base now that rpm-ostree updated

cgwalters:
Since we've updated to `rpm-ostree-2018.8` in the `coreos-assembler:alpha`
container, we can drop duplicates from host-base.yaml.


dustymabe:
LGTM

PR #326, opened 2018-09-27
pipeline: don't send Slack update for dev pipelines

miabbott:


miabbott:
@dm0- Yeah, you are right.  I was too focused on the actual sending rather than the function as a whole.  Pushed another update :arrow_up:

PR #333, opened 2018-10-01
utils: don't nuke root dir in `prep_container_storage()`

miabbott:
The function should just nuke the `containers` dir, rather than the
root of the `containers` dir.

### DIFF
--- a/pipeline-utils.groovy
+++ b/pipeline-utils.groovy
@@ -168,7 +168,7 @@ def prep_container_storage(dirFromHost) {
             exit 1
         fi
         rm -rf \${container_storage} && mkdir -p \${container_storage}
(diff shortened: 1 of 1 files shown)


miabbott:
The `aws-test` job was failing because it couldn't find the JSON file it needed to copy to S3.  

```
+ aws s3 cp --acl public-read /home/jenkins/workspace/coreos-rhcos-aws-test/aws-us-east-1.json s3://****/aws-us-east-1-tested.json

The user-provided path /home/jenkins/workspace/coreos-rhcos-aws-test/aws-us-east-1.json does not exist.
```

The root cause looks like this line:

`+ rm -rf /home/jenkins/workspace/coreos-rhcos-aws-test`

...so this *should* fix that.

PR #337, opened 2018-10-01
treecompose: pass correct args to `pull-mount-oscontainer`

miabbott:
In #334, we changed to do the registry login outside of the
`pull-mount-oscontainer` script, so that we only require two arguments
to the script now.  So change the invocation accordingly.


cgwalters:
Thanks!  Was just looking at this too.  FTR I did test my PR but I think I only watched that my job did a successful login :cry: 

/lgtm

miabbott:
> Thanks! Was just looking at this too. FTR I did test my PR but I think I only watched that my job did a successful login cry

Well, I didn't test this, so we'll see how it goes  :man_shrugging:

PR #339, opened 2018-10-01
jenkins: Drop separate API_CI_REGISTRY constant

cgwalters:
Now that our login API parses the ref, there's no reason to have
it split out.

This clarifies that what we care about primarily is the oscontainer
reference, we're not currently interacting with that registry
in any other way.


miabbott:
Needs a rebase, but otherwise looks good

PR #340, opened 2018-10-01
treecompose: use 'skopeo copy' to push new tag to registry

miabbott:
The two `podman push` operations were taking a long time to complete
and could fail the whole pipeline if one failed.  Let's try using
`skopeo copy` to perform the additional tagging.


miabbott:
I tested this locally with some test images and quay.io....I hope that the OpenShift registry behaves the same way :crossed_fingers:

PR #341, opened 2018-10-02
scripts/ami-copy-regions: Try waiting on all AMIs

cgwalters:
- Unbuffer stdout so stderr isn't confusingly first
 - Try a --dry-run first so that we verify we have permissions
   (this may be the issue?)
 - Finally and perhaps most importantly, wait for every AMI at
   least once.  This way if the first one is slow, we'll move
   on to the next.

### DIFF
--- a/scripts/ami-copy-regions
+++ b/scripts/ami-copy-regions
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/python3 -u
 # Copy an AMI to multiple regions, generating an "AMI JSON"
 # file matching the Container Linux schema:
 # https://alpha.release.core-os.net/amd64-usr/current/coreos_production_ami_all.json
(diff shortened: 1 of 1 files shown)


cgwalters:
(Tested lightly locally)

cgwalters:
/hold cancel

Now tested better

scripts/ami-copy-regions
```diff
@@ -70,25 +72,52 @@ for region in args.regions:
...
+    else:
+        return subprocess.call(args) == 0
 
 print("Using modify-image-attribute to make AMIs public (may take a while)")
+amis_succeeded = set()
+# Try waiting for each AMI twice; if we were close to the timeout
+# this could take a long time, but in practice EC2 parallelizes
+# so this way we avoid failing if the first image or two happens
+# to take too long.
+for _ in range(1):
```
yuqi-zhang: Maybe this is a really dumb question but what's the point of this line?

cgwalters: The comment above was attempting to explain that.

cgwalters: This by the way is why we have code review :smile:

yuqi-zhang: Yeah, I was slightly confused and thought you were trying to do something magical with a single loop that I wasn't aware of :smile: 

The copying seems to have added ~14 minutes to the job run, which I guess is normal given that one upload in the `cloud` job takes about the same time.

PR #357, opened 2018-10-12
maipo: use correct, up-to-date repo

miabbott:
The previous repo configuration was pulling in packages that were
older than what we are currently shipping in RHELAH.  (The older
packages appear to be from 7.5 GA)

### DIFF
--- a/maipo.repo
+++ b/maipo.repo
@@ -1,13 +1,13 @@
 [maipo-server]
 enabled=1
-baseurl=http://pulp.dist.prod.ext.phx2.redhat.com/content/dist/rhel/server/7/7.5/x86_64/os/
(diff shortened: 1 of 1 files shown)


cgwalters:
I meant to reply earlier but won't this pull in 7.6 content again?

ashcrow:
It will. And if it's like the previous mirror we used it had a combination of 7.x->7.6 content differing for different packages unlike 7.5 which was 7.5 fully.

cgwalters:
OK actually looking at the content it looks like it's just 7.5; I think the root of my confusion here is that this baseurl mirrors the customer view, but the other one we were using before is development.

/hold cancel

miabbott:
I wasn't on top of the notifications for this PR, but you guys figured it out.  :+1:

PR #363, opened 2018-10-22
Remove rdgo bits here

cgwalters:
We're moving this to another repo.


cgwalters:
Yep!

ashcrow:
Side note: It's probably a good idea to update the pipeline docs after this merges.

PR #366, opened 2018-11-02
Remove v1 pipeline

cgwalters:
This has moved to an internal repository.


ashcrow:
/lgtm

This should help avoid confusion over where content is being built from :+1:

PR #385, opened 2019-10-10
FAQ.md: new file

cgwalters:
Let's start answering things in public and in git.
See also https://github.com/openshift/machine-config-operator/pull/1165

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,19 @@
+# Questions and answers
+
+The goal of this file is to have a place to easily commit answers to questions
+in a way that's easily searchable, and can make its way into official
+documentation later.
+
+## Q: How do I provide static IP addresses?
+
+As of OpenShift 4.2, by default the kernel command line arguments for networking
+are persisted.  See this PR: https://github.com/coreos/ignition-dracut/pull/89
+
+In cases where you want to have the first boot use DHCP, but subsequent boots
+use a different static configuration, you can write the traditional Red Hat Linux
+`/etc/sysconfig/network-scripts` files, or NetworkManager configuration files, and
+include them in Ignition.
+
+The MCO does not have good support for "per-node" configuration today, but
+in the future when it does, writing this as a MachineConfig fragment
+passed to the installer will make sense too.


FAQ.md
```diff
@@ -0,0 +1,17 @@
...
+The goal of this file is to have a place to easily commit answers to questions
+in a way that's easily searchable, and can make its way into official
+documentation later.
+
+## Q: How do I provide static IP addresses?
+
+As of OpenShift 4.2, by default the kernel command line arguments
+are persisted.  See this PR: https://github.com/coreos/ignition-dracut/pull/89
+
+In both cases, you can use the traditional Red Hat Linux `/etc/sysconfig/network-scripts`
```
yuqi-zhang: This feels a bit confusing. Maybe instead of "in both cases", we can elaborate a bit by saying: you can set static IP either by:
 - setting it via the kernel command line and having it persist
 - dropping in an ignition snippet to create a `/etc/sysconfig/network-scripts`

ashcrow:
> I'm also curious how many people follow this repo.

As in `watch` or as in users/external devs looking to this repo for `$THING`?

> Are we going to direct people here for questions and make this the canonical RHCOS Q/A? Would it be easier to add a FAQ section to docs.openshift.com?

That's a great question. I believe it depends upon the audience. If the audience is end users docs.openshift.com makes sense. **However**, content in docs.openshift.com does need to come from somewhere and, generally speaking, developer focused docs are a good place to start from. If this is meant more so for folks kicking tires for development/testing/getting up to speed I'd say this is a great place to host a FAQ.

In short, I think this FAQ is a great idea which will help docs folks add content to docs.openshift.com.

FAQ.md
cgwalters: Mmm...not quite, because if you just do the latter, the first boot is DHCP.  I reworded the text.

cgwalters:
> Are we going to direct people here for questions and make this the canonical RHCOS Q/A? Would it be easier to add a FAQ section to docs.openshift.com?

I think we clearly should be submitting PRs to [the docs](https://github.com/openshift/openshift-docs/) more often indeed - but I see this as a "staging ground" for some cases of that.  Basically let's try to have it be a lightweight/fast process, not necessarily a lot of polish - just ensuring that *some* answers are public.

Then the content here can be re-written and polished and turned into docs if that makes sense.

PR #397, opened 2020-01-15
FAQ: add information for making custom RHCOS builds

miabbott:


FAQ.md
```diff
@@ -96,3 +96,15 @@ While this can be helpful for short-term fixes, it is important to remember that
 is in place, as the cluster currently has [no mechanism for reporting](https://github.com/openshift/machine-config-operator/issues/945) that the node has been changed in this
 fashion.  This kind of package replacement can also leave your nodes exposed to potential problems
 that are fixed in newer versions of the package.
+
+## Q: How do I build my own version of RHCOS for testing?
+
+You need the RHCOS manifest configuration (currently hosted on an RHT internal GitLab repo) and
```
yuqi-zhang: Maybe put an url.corp.redhat.com url to point to it?

FAQ.md
```diff
@@ -96,3 +96,15 @@ While this can be helpful for short-term fixes, it is important to remember that
...
+[coreos-assembler](https://github.com/coreos/coreos-assembler).
+
+If you want to replace particular binaries or RPMs in RHCOS, the `coreos-assembler` has
+[override mechanisms](https://github.com/coreos/coreos-assembler/blob/master/README-devel.md#using-overrides)
+to do this.
+
+Also reference the docs from the `machine-config-operator` about
```
yuqi-zhang: Maybe add a small snippet about what the machine-os-content is in this case, or a snippet on what building your own RHCOS for a single machine test vs using in-cluster

cgwalters: Mmm.  I think the former is well covered by the MCO, no?  As for the latter...would still lean a bit towards covering that in the MCO docs to be honest.

miabbott:
Can I `lgtm` my own PR?  Let's find out...

/lgtm

miabbott:
Oops, missed those comments somehow...editing now

FAQ.md
miabbott: Can do!

PR #404, opened 2020-04-22
FAQ: Which RHEL when

cgwalters:
This has been asked a few times.


miabbott:
> `Not mergeable. Needs bugzilla/valid-bug label. `
> 
> Heh...I wonder if we can tune that off for this repo?

Trying to relax the rules here - https://github.com/openshift/release/pull/8521

miabbott:
Let's try this to refresh labels...

/retest

PR #408, opened 2020-07-14
FAQ: add info RPM versions in RHCOS

miabbott:


FAQ.md
```diff
@@ -91,6 +91,23 @@ Here the `81` means it's using RHEL 8.1.
 
 RHEL CoreOS ships RHEL updates after they're released.  Usually, RHEL 8.X updates will land in all OpenShift streams (e.g. `4.3`, `4.4`) at around the same time, and usually first in the development builds.  To see this, inspect the versions per above.
 
+## Q: How do I determine what version of an RPM is included in an RHCOS release?
+
+The contents of each RHCOS release are visible in the [release browser](https://releases-rhcos-art.cloud.privileged.psi.redhat.com/) via the "OS contents" link next to each build.
```
cgwalters, line 96: The fact that the release UI is behind the firewall makes it weird to link to from a public Github repo, but we already have that problem with the main page.

So
/lgtm

miabbott: I thought we had a copy of release-browser running on api.ci, but alas that does not exist

PR #413, opened 2020-10-09
Publish redhat-coreos config sources

cgwalters:
Today the redhat-coreos config git repository (equivalent of https://github.com/coreos/fedora-coreos-config ) lives internally to the RH firewall.  This is problematic for many reasons - we're not consistently publishing the sources here, and also inhibits our integration with the upstream OpenShift CI (Prow).

This PR imports the config.  I propose we move it to Github canonically.


kola-blacklist.yaml
```diff
@@ -0,0 +1,20 @@
+# This file documents currently known-to-fail kola tests. It is consumed by
```
travier, line 1: This file should probably be removed.

miabbott:
I think we can drop the `installer` dir in favor of the `live` dir?

miabbott:
@darkmuggle @bh7cw @mike-nguyen @ravanelli This change will likely intersect with some of the pipeline work you folks are responsible for.

miabbott:
/approve

The first step is getting things imported; I imagine there are going to be follow-ups to refine this.  But as a start, this looks reasonable.

dustymabe:
I have no objections. :heavy_check_mark:

jlebon:
Overall
/approve
also from me. Though it looks like this is also adding support for a "CentOS CoreOS" which I think we should consider independently of the "move configs to GitHub" endeavour.

cgwalters:
> Though it looks like this is also adding support for a "CentOS CoreOS" which I think we should consider independently of the "move configs to GitHub" endeavour.

Yeah true, I can try to drop that.

cgwalters:
OK updated to address comments, dropping draft!

overlay.d/05rhcos/usr/sbin/gcp-routes.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/bash
```
miabbott, line 1: Do we still need this now that it lives in the MCO?

https://github.com/openshift/machine-config-operator/pull/2011

cgwalters: This is just reflecting the state of the internal git repository; let's not try to *also* change it at the same time.

cgwalters: I mean if we succeed at this, then we could do a PR to drop that script and have it tested by Prow launching a cluster in Azure!

miabbott:
Seems like we are in a good position to move forward.

/approve

jlebon:
OK, tested locally and verified that it's up to date.
Let's merge this! The final switchover happens internally.
/lgtm

PR #414, opened 2020-10-15
download-extensions: Avoid duplicating repo enablement list

cgwalters:
I don't know of a reason not to enable all repos when
downloading, same as we do with rpm-ostree.  Let's keep
it in one place.


ashcrow (approved):
Good idea

PR #415, opened 2020-10-15
Document and standardize on rpm-repo names used by api.ci

cgwalters:
Better document the fact that we don't include any `.repo`
files that you need to build this.

Update the manifest to use the repo names used by api.ci;
this *will break* current downstream RHCOS until we adjust
it to match these repo names.

I'll look at a patch for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,3 +7,41 @@ is entangled with various private things and is omitted.
 
 Going forward, this repository will be canonical; more
 to come.
+
+# Building
+
+You need to create a `rhcos.repo` file that looks like this:
+
+```
+# RHEL repos
+[rhel8-baseos]
+baseurl=<url>
+
+[rhel8-appstream]
+baseurl=<url>
+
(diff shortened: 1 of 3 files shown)


cgwalters:
On the plus side though in combination with https://github.com/coreos/coreos-assembler/pull/1795 I was able to do a basic `cosa build` inside a regular unprivileged pod in api.ci with this.

cgwalters:
Requires: https://github.com/openshift/release/pull/12923

cgwalters:
TODO document `curl -L base-4-7-rhel8.ocp.svc.cluster.local > src/config/ocp.repo`

cgwalters:
OK lifting draft on this!  I tested this in concert with [this internal MR](https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1152).

miabbott:
Thanks for pushing this forward, Colin!

/lgtm

PR #418, opened 2020-10-27
manifest: add kexec-tools for kdump

jlebon:
This was discussed in:
https://github.com/coreos/fedora-coreos-tracker/issues/622

### DIFF
--- a/overlay.d/05rhcos/usr/lib/systemd/system-preset/10-rhcos-disabled.preset
+++ b/overlay.d/05rhcos/usr/lib/systemd/system-preset/10-rhcos-disabled.preset
@@ -3,3 +3,7 @@
 # https://github.com/coreos/afterburn/issues/405
 # When we changed to inherit from FCOS we started enabling this service
 disable afterburn-sshkeys@.service
+
+# Explicitly disable: it's enabled by default otherwise but will fail because
+# we don't ship with the kargs needed for enablement.
+disable kdump.service
--- a/rhcos-packages.yaml
+++ b/rhcos-packages.yaml
@@ -18,3 +18,6 @@ packages:
   - rpm-ostree
   # part of container-tools module
   - toolbox
+  # for kdump:
+  # https://github.com/coreos/fedora-coreos-tracker/issues/622
+  - kexec-tools


jlebon:
Similar deps to those described in https://github.com/coreos/fedora-coreos-config/pull/708:

```
Added:
  dracut-squash-049-75.git20200422.el8.x86_64
  ethtool-2:5.0-2.el8.x86_64
  kexec-tools-2.0.20-14.el8.x86_64
  lzo-2.08-14.el8.x86_64
  snappy-1.1.7-5.el8.x86_64
  squashfs-tools-4.3-19.el8.x86_64
```

This brings in `lzo`, which was already in FCOS but not yet in RHCOS.

/hold
while I sanity-check this locally.

cgwalters:
Not opposed to this but I'd like to see at least some minimal investigation of why the heck it depends on `dracut-squash` for example.

jlebon:
> Not opposed to this but I'd like to see at least some minimal investigation of why the heck it depends on `dracut-squash` for example.

Did you see https://github.com/coreos/fedora-coreos-config/pull/708?

jlebon:
OK, dropping hold on this!
/hold cancel

ashcrow:
This change looks good to my eyes BUT it may be something that makes more sense in extensions for OCP/RHCOS. If I remember correctly having kdump on the host but not in use can be a problem for some audits. cc'ing @jaormx to jog my memory :smile:

JAORMX:
I agree with @ashcrow , seems to me like this would be more appropraite in extensions. For compliance, folks are required to disable kdump in their deployments with the following rationale:

> Kernel core dumps may contain the full contents of system memory at the time of the crash. Kernel core dumps consume a considerable amount of disk space and may result in denial of service by exhausting the available space on the target file system partition. Unless the system is used for kernel development or testing, there is little need to run the kdump service.

cgwalters:
Ahh I hadn't realized there was a compliance angle to this.

Well, currently today RHCOS just inherits the "bootable-rpm-ostree.yaml" bits so adding it to FCOS doesn't *preclude* doing it via an extension.

That said:

> Unless the system is used for kernel development or testing, there is little need to run the kdump service.

Note that `kdump.service` won't (shouldn't) run by default unless the `crashkernel=` argument is provided...hm actually I think it really should have `ConditionKernelCommandLine=crashkernel`.

IOW while it's *installed* by default it won't be *enabled* by default on FCOS, and a compliance system can easily check this.

jlebon:
Not against having it as an extension in RHCOS. This PR comes out of rough agreement from the team to have it baked in to reduce friction in enabling what we want to consider an important host feature. In pure FCOS, this means one less reboot. In OCP/OKD however, because the host always incurs a reboot, having it as an extension is actually not much more expensive. But diverging from FCOS has downsides as well, esp. around test sharing.

FWIW, I can imagine FCOS eventually removing it from the host and making it an extension in the future once we have "live extensions" (which don't require reboots).

But anyway, as Colin said, this wouldn't be enabled by default (and in fact, this PR adds a preset exactly to revert that).

ashcrow (approved):
Let's ensure this continues to be disabled by default and we can look at extensions later.

cgwalters:
>  (and in fact, this PR adds a preset exactly to revert that).

Ahh I had missed that.  But then, why does the preset enable it by default?  Hmm well I guess that'd break existing users.

Does my suggestion
> hm actually I think it really should have ConditionKernelCommandLine=crashkernel
seem right? Then we wouldn't need a preset, it'd be "enabled by conditionally off".

jlebon:
> Does my suggestion
> 
> > hm actually I think it really should have ConditionKernelCommandLine=crashkernel
> 
> seem right? Then we wouldn't need a preset override, it'd be "enabled by conditionally off".

I think so. Let's see what the maintainers say: https://src.fedoraproject.org/rpms/kexec-tools/pull-request/5

cgwalters:
/approve
Do we feel the compliance angle is resolved? 
/assign ashcrow
for the lgtm based on that

JAORMX:
> /approve
> Do we feel the compliance angle is resolved? 
> /assign ashcrow
> for the lgtm based on that

Yep. As long as its not enabled by default it should be fine.

cgwalters:
> Yep. As long as its not enabled by default it should be fine.

:+1: 

So on that note...we could add something to `tests/kola/misc-ro/misc-ro.sh`

jlebon:
> So on that note...we could add something to `tests/kola/misc-ro/misc-ro.sh`

Let's do FCOS first, then we can copy/paste into RHCOS: https://github.com/coreos/fedora-coreos-config/pull/716

Issue #426, opened 2020-11-03
remove null LUKS container, transition to Ignition LUKS on upgrade


cgwalters:
Some discussion of this here https://hackmd.io/jBhJUEIqSdOfmnZR-aI0_g?edit

arithx:
My general approach is: remove the [`crypt_rootfs` section from COSA](https://github.com/coreos/coreos-assembler/blob/master/src/create_disk.sh#L158-L199), stop setting the [`rhcos.luks` karg](https://github.com/openshift/os/blob/master/image.yaml#L9), rip out all everything but the unlocking bits from [`coreos-cryptfs`](https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/libexec/coreos-cryptfs), add a bash script (see more below) that runs in the real root every boot for the duration of 4.7.

The bash script running in the real root is there to do two main things: changing the old `rhcos.luks` kargs -> `rd.luks` (which is what the new Ignition LUKS uses) & fixing the partition table in the case of null_cipher (unencrypted) root devices. For the second case right now if we have a non-encrypted root device we have the root device inside a `null_cipher-esb` volume. To avoid performance hits we create a [dm-linear mapping of the partition minus the LUKS header](https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/libexec/coreos-cryptfs#L117-L136), the bash script is just doing that in a permanent way (modifying the partition table changing the starting offset of the partition to omit the LUKS header instead of creating the dm-linear mapping on every boot).

The main idea behind this approach is optimized around
1) remove the legacy ways of touching the root filesystem (Ignition should be handling this in all cases now)
1) not needing an update barrier; for the entire duration of 4.7 we can still detect & unlock previous versions root filesystems, we are just fixing them after we detect them so that we won't need to carry the `coreos-cryptfs` openers starting in 4.8

PR #425, opened 2020-11-03
Bump fedora-coreos-config

cgwalters:
Just on general principle, but specifically in prep for
adding bootupd to RHCOS.

```
Benjamin Gilbert (4):
      overrides: fast-track coreos-installer 0.6.0
      overrides: fast-track coreos-installer 0.7.0-2
      20live/coreos-livepxe-rootfs: drop support for legacy PXE images
      overrides: fast-track rust-afterburn-4.5.1-3.fc32

Clement Verna (2):
      [testing-devel] overrides: drop graduated overrides
      overrides: drop graduated overrides

Colin Walters (17):
      40ignition-ostree: Initial support for rootfs replacement
      40ignition-ostree: Regenerate UUIDs for /boot and /
      Revert "Add root reprovisioning tests"
      overlay: Install /usr/share/licenses/fedora-coreos-config/LICENSE
      live-generator: A few systemd ordering tweaks
      manifest: Add bootupd for x86_64|aarch64 (socket disabled)
      coreos-boot-mount-generator: Use Options=umask=077 for /boot/efi
      overlay/coreos-boot-mount-generator: Small code simplification
      tests/manual: Clarify expected state
      40ignition-ostree: Fix Before= for fsck for root
      growpart: Add btrfs support
      Move documentation removal to toplevel manifest
      manifests/bootloader: Drop armhfp
      Drop bits in overlay to chmod (use cosa fix), add test
      overlay: Enable bootupd.socket by default
      manifests/bootupd: `mkdir /run` for now
      lockfile: Bump bootupd

CoreOS Bot (68):
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
      lockfiles: bump to latest
[...]

PR #427, opened 2020-11-03
40rhcos-fde: gate rootfs reprovisioning behind coreos.rootfs-replace.exp

jlebon:
Right now, with the latest FCOS, we get failures because e.g. the
rootmap code trips up on the dm-linear target trick we use.

This is a transitional state towards using the new rootfs reprovisioning
and LUKS-via-Ignition work. So for now so that we can keep bumping the
FCOS submodule, let's just gate all the reprovisioning services behind a
karg. That way, devs can still re-enable them if they'd like.

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/coreos-luks-open.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/coreos-luks-open.service
@@ -13,6 +13,9 @@ After=coreos-encrypt.service
 # table, udev, etc..
 After=ignition-disks.service
 
+# Need to open device before we can restamp it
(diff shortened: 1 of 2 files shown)


jlebon:
Depending on how close #426 is, we could just not do this at all and go straight to the new world.

cgwalters:
Did you test this locally?  I am hitting some cosa bugs but I believe you this works.

jlebon:
It did at the time I wrote it up initially. I'm retesting with the latest.

jlebon:
/hold

This doesn't work anymore against latest FCOS.

cgwalters:
Probably best to fold any work into https://github.com/openshift/os/pull/425 ?

jlebon:
Merging as part of https://github.com/openshift/os/pull/425.

PR #428, opened 2020-11-03
 manifest: Include bootupd

cgwalters:
Basically pulling in coreos/fedora-coreos-config#595
for RHCOS.  The immediate goal is allowing updating existing
systems in place for Boot Hole.


cgwalters:
This depends on https://github.com/openshift/os/pull/425

cgwalters:
Rebased :surfer:

ashcrow (approved):

PR #430, opened 2020-11-04
docs: Move support entry higher in the FAQ & README

travier:
This should help increase visibility for the official support option.


cgwalters (approved):
/lgtm

PR #432, opened 2020-11-05
overlay.d: remove 41rhcos-ignition-conf

sohankunkerkar:
**Note:** This PR is not mergeable until we cut the new release of Ignition with coreos/ignition#1108 

Ignition now supports a `base.d` directory for base config fragments that allows FCOS and RHCOS to maintain the same `00-core.ign`, and facilitate the distro-specific changes by placing the other base config fragments in a separate `overlay` directory(coreos/ignition#1108).
We don't need to maintain a separate base config for rhcos anymore.

/hold


jlebon:
Could you provide the shortlog for the FCOS bump as part of the commit message? (See e.g. https://github.com/openshift/os/pull/447.) It's helpful for reviewing because it's a sanity-check that the commit you wanted to pull in is there, and that there isn't anything else that gets pulled in that needs more care.

sohankunkerkar:
> Could you provide the shortlog for the FCOS bump as part of the commit message? (See e.g. #447.) It's helpful for reviewing because it's a sanity-check that the commit you wanted to pull in is there, and that there isn't anything else that gets pulled in that needs more care.

@jlebon Thanks for the review. I have addressed your concern in the subsequent push.

jlebon:
Thanks!
/lgtm

PR #433, opened 2020-11-05
Bump fedora-coreos-config to get ordering cycle fix

cgwalters:
```
Colin Walters (1):
      uuid-root: Add an After= for the device

CoreOS Bot (1):
      lockfiles: bump to latest

Jonathan Lebon (4):
      tests/kola: move swap-on-zram to misc-ign-ro
      coreos-update-ca-trust: order after local-fs.target
      tests/kola: add a test for coreos-update-ca-trust.service
      40ignition-ostree: use `lsblk --paths` instead of -o PATH
```


cgwalters:
The new change in https://github.com/coreos/coreos-assembler/pull/1834 breaks current RHCOS without this.

PR #434, opened 2020-11-05
manifest.yaml: include bootupd only for x86 and aarch64

Prashanth684:
bootupd doesn't apply to other arches causing build failures


jlebon:
Ahh yup, this matches FCOS too.
/lgtm

PR #435, opened 2020-11-12
Move to RHEL 8.3 content

travier:
Currently disables LUKS related tests until https://github.com/openshift/os/issues/426 is fixed.

/hold


ashcrow (approved):
:tada: :heart:

ashcrow:
/hold

So we don't merge early accidentally.

arithx:
Worth pointing out that just fixing #426 won't be enough for us to re-enable the LUKS tests. The tests themselves also need to update to create the root LUKS device via Ignition instead of the old method.

arithx:
#436 has two additional commits on top of these aimed at addressing #426

jlebon:
TPM pinning at least is covered by https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/root-reprovision/luks. We'll probably still need a kola test for now for Tang pinning.

arithx:
> TPM pinning at least is covered by https://github.com/coreos/fedora-coreos-config/tree/testing-devel/tests/kola/root-reprovision/luks. We'll probably still need a kola test for now for Tang pinning.

Tang-based root devices might be hard to do in external tests unless we add a method to have it configured on the host / spin up a second machine. We have [GRPA-2887](https://issues.redhat.com/browse/GRPA-2887) filed for updating the existing `rhcos.luks.*` tests (which are native tests for both tpm & tang) to use Ignition LUKS.

jlebon:
> Tang-based root devices might be hard to do in external tests unless we add a method to have it configured on the host / spin up a second machine.

Right, that's what I meant. :) We can remove the native TPM pinning from kola, and only adapt the Tang one.

jlebon:
@arithx @travier, IIUC this isn't actually blocked on #426, right? OTOH, https://github.com/openshift/os/pull/436 is blocked on this so we need to get this in first.

arithx:
> @arithx @travier, IIUC this isn't actually blocked on #426, right? OTOH, #436 is blocked on this so we need to get this in first.

I think we'd want to cherry-pick in https://github.com/openshift/os/pull/436/commits/6b085d4bc4021e6c614ede5caea15d613d9f0456 to this PR but other than that I think we'd be good to ship this.

travier:
> I think we'd want to cherry-pick in [6b085d4](https://github.com/openshift/os/commit/6b085d4bc4021e6c614ede5caea15d613d9f0456) to this PR but other than that I think we'd be good to ship this.

Sounds good to me. Pushed your first commit into this one.

jlebon:
/lgtm

To be clear, this will break LUKS until we merge #436 as well.
Keeping the hold on this; will double-check that we're all on the same page before cancelling it.

cgwalters:
I don't see the value in temporarily breaking LUKS, we're pretty sure we're going to get that fix landed soon right?

jlebon:
> I don't see the value in temporarily breaking LUKS, we're pretty sure we're going to get that fix landed soon right?

Right. Simply that it's a prerequisite for #436 and makes testing 8.3 easier. I don't think we necessarily have to be super strict about this, but I'm not against trying to stay green throughout either.

cgwalters:
I just don't like the idea of this landing in the current 4.7 OpenShift builds.  IMO the LUKS support is important and people will be testing upgrades from 4.6 with it (and expecting it to work in new installs).

One option is to land this in this repo but *not* update the submodule downstream quite yet.  It'd be effectively a temporary fork.

jlebon:
> One option is to land this in this repo but not update the submodule downstream quite yet. It'd be effectively a temporary fork.

Yup, SGTM. Alternatively, I guess we could just close this PR and fold it into #436 so it's done atomically. Will leave it to @travier and @arithx.

miabbott:
> I just don't like the idea of this landing in the current 4.7 OpenShift builds. IMO the LUKS support is important and people will be testing upgrades from 4.6 with it (and expecting it to work in new installs).

OCP CI is in a bad place right now as well (due to https://bugzilla.redhat.com/show_bug.cgi?id=1897604), so any "free" coverage we get through OCP CI jobs is going to be missing.

jlebon:
This was merged as part of #436.
/close

PR #436, opened 2020-11-12
Transition to Ignition LUKS

### LINKED ISSUE #426: remove null LUKS container, transition to Ignition LUKS on upgrade

arithx:
We now have support for both LUKS & moving the root partition via
Ignition. Drop the old method for encrypting the root device as well as
the image.yaml arguments related to it. Also add a new systemd service
that runs at shutdown to modify kargs / the partition table to align more
closely to what Ignition would produce so that we can fully drop
`coreos-cryptfs` unlocking bits in 4.8

Closes #426


overlay.d/05rhcos/usr/lib/systemd/system/rhcos-fix-root.service
```diff
@@ -0,0 +1,17 @@
+# Fixes the partition table / kargs from machines upgrading from
+# older releases before Ignition LUKS was available
+[Unit]
+Description=Fix root device
+
+# Run before shutdown to ensure that we aren't running in a state
+# where the partition table on disk differs from the kernel's view
+Before=shutdown.target reboot.target halt.target
```
cgwalters, line 8: Last time I tried to do this it didn't really work, see
https://github.com/ostreedev/ostree/blob/d046631a0f60c3e31b46eb967da29f7f94c5caaf/src/boot/ostree-finalize-staged.service#L37

Basically use `ExecStop=` - trying to start things at shutdown basically conflicts with the systemd model.

arithx: Hmm; If I leave the unit as is just changing `ExecStart=` to `ExecStop=` then it's correctly moving the partition table but isn't updating the kargs (causing a boot failure).

cgwalters:
> Hmm; If I leave the unit as is just changing ExecStart= to ExecStop= then it's correctly moving the partition table but isn't updating the kargs (causing a boot failure

I think this would need to be `Before=ostree-finalize-staged.service` probably; this is likely racing with that.

arithx:
> I think this would need to be Before=ostree-finalize-staged.service probably; this is likely racing with that.

With that `Before=` target & using `ExecStop=` it's still hitting the same problem of the kargs not being set with the partition table moved.

Given we need to make sure that we've actually finished running the `/usr/libexec/rhcos-fix-root` script before that unit runs (to prevent racing with ostree finalization) should we not just leave it as `ExecStart=` and change to `Before=ostree-finalize-staged.service`

overlay.d/05rhcos/usr/libexec/coreos-cryptfs
```diff
@@ -141,163 +141,7 @@ try_open() {
     fi
 }
 
-token_rm() {
```
darkmuggle, line 144: The red is _awesome_

overlay.d/05rhcos/usr/lib/systemd/system/rhcos-fix-root.service
```diff
@@ -0,0 +1,17 @@
+# Fixes the partition table / kargs from machines upgrading from
+# older releases before Ignition LUKS was available
+[Unit]
+Description=Fix root device
```
bgilbert, line 4: Descriptions use title case.  `Upgrade Root Filesystem Encryption Container`?

jlebon, line 4: How about "CoreOS Migrate Root LUKS Container"? (I'd call the service and script files something along those lines too to be a bit more descriptive.)

jlebon, line 4: Don't we want a `ConditionKernelCommandLine=rhcos.root=crypt_rootfs` here?

overlay.d/05rhcos/usr/libexec/rhcos-fix-root
```diff
@@ -0,0 +1,68 @@
...
+    # check / handle 4k sector size devices
+    sector_size=$(cat /sys/block/$parent/queue/logical_block_size)
+    if [ "$sector_size" == "4096" ]; then
+        startsector=$((startsector / 8))
+        size=$((size / 8))
+    fi
+
+    # validate that the new partition will start at the xfs super block
+    skipcount=$(($startsector * ($sector_size/4)))
+    [ $(dd if=/dev/$parent bs=4 count=1 status=none skip=$skipcount) == XFSB ]
```
bgilbert, line 32: Let's add a proper `if` here, because if this test fails, it'll exit nonzero without a log message.

`count=4 bs=1` would allow avoiding the awkward division by 4 on the previous line.

overlay.d/05rhcos/usr/lib/systemd/system/rhcos-fix-root.service
jlebon: Try using `After=ostree-finalize-staged.service` too. That way we'll stage the kargs modification *before* the service that finalizes it. (Basically, the way this works is systemd runs the `ExecStop`s in reverse order).

overlay.d/05rhcos/usr/libexec/rhcos-fix-root
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+set -euxo pipefail
+
+if [ -f /run/device_mod ]; then
```
jlebon, line 4: I think we can drop the `/run/device_mod` stamp file if we use the `ConditionKernelCommandLine=rhcos.root` bit?

jlebon (commented):
First superficial review. Haven't read the script in depth yet. (How confident are you in that bash script BTW? I know it's RHCOS-specific, though this would be a good candidate for `rdcore` given the high-risk aspect.)

overlay.d/05rhcos/usr/libexec/rhcos-fix-root
bgilbert: I asked for that stamp.  The concern is that the script should not run twice if it's started twice within the same boot.

arithx:
> How confident are you in that bash script BTW?

As confident as I think I could be in a bash script performing operations which could potentially cause a machine not to boot in the case of failure... With the check for the XFS superblock in the non-crypt case we at least shouldn't act in a way that would prevent booting.

> I know it's RHCOS-specific, though this would be a good candidate for rdcore given the high-risk aspect.

My reasoning for why I did it as a bash script instead of putting it into `rdcore` was that ideally we only want to keep this around for 4.7 to fix it the machine the first time it upgrades into 4.7. Once we move to 4.8 we can assume that the root filesystem will be in a similar state to a fresh install that never ran through the legacy LUKS code and as such we can drop both this conversion script & the `coreos-cryptfs` LUKS opener / dm-linear mapper. While we could still do that inside of `rdcore` I'm not sure I'd prefer ripping functionality out of `coreos-installer`/`rdcore` even if that code path would never run again.

Prashanth684:
hitting https://github.com/coreos/coreos-assembler/issues/1496 with cryptsetup from RHEL-8.3 repo (cryptsetup-2.3.3-2.el8.s390x).

jlebon:
Oh right, this PR also needs:

```diff
diff --git a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
index 2d88dd5..0b4db1e 100755
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
@@ -85,21 +85,8 @@ install() {
     inst_script /usr/libexec/coreos-cryptlib

     # Service for first-boot encryption.
-    install_unit "coreos-fde-check-needsnet.service" "ignition-diskful.target"
-    install_unit "coreos-encrypt.service" "ignition-diskful.target"
     inst_simple "$moddir/coreos-luks-open.service" "$systemdsystemunitdir/coreos-luks-open.service"
     # The generator enables the opening service.
     inst_simple "$moddir/coreos-luks-generator" \
         "$systemdutildir/system-generators/coreos-luks-generator"
-
-    # Gate all the rootfs replacement stuff for now behind a development flag
-    # since it conflicts with this module. Soon, we'll nuke this module entirely
-    # anyway in its place.
-    for x in ignition-ostree-rootfs-{detect,save,restore} coreos-inject-rootmap; do
-        mkdir -p $initdir/$systemdsystemunitdir/${x}.service.d/
-        cat > $initdir/$systemdsystemunitdir/${x}.service.d/neuter.conf <<EOF
-[Unit]
-ConditionKernelCommandLine=coreos.rootfs-replace.exp
-EOF
-    done
 }
```

jlebon:
OK, folded that into the "drop old LUKS" commit and also removed the service units we no longer need.

arithx:
Updated! Summing up some OOB discussion:

Leaving `rhcos-upgrade-root-filesystem-container` running at boot up after `multi-user.target`; we don't want to be editing the BLS by hand and `rpm-ostree kargs` is using the DBus which means that we can't run it during the systemd shutdown process (systemd will throw `Refusing activation, D-Bus is shutting down.` during shutdown).

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+    sgdisk -d $partnum -n $partnum:$startsector:+$size -c $partnum:$label /dev/$parent -u $partnum:$guid
+    # delete the rhcos.root karg to prevent unlocking attempts
+    rpm-ostree kargs --delete rhcos.root
+
+    > /run/device_mod
+    exit
+fi
+
+# pre-4.7 root LUKS devices are always named luks-00000000-0000-4000-a000-000000000002
+if [ -h /dev/mapper/luks-00000000-0000-4000-a000-000000000002 ]; then
```
jlebon, line 48: Hmm, feels like this should be an `else` instead, no? I.e. either we're handling the encrypted case, or the non-encrypted case. 

Then we can factor out the `> /run/device_mod` stamp file and drop the explicit `exit`s.

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+    startsector="$(($(cat /sys/block/$parent/$partition/start)+32768))"
+    # the dm device is already the correct size
+    size=$(cat /sys/block/$dm_device_name/size)
+
+    # check / handle 4k sector size devices
+    sector_size=$(cat /sys/block/$parent/queue/logical_block_size)
+    if [ "$sector_size" == "4096" ]; then
+        startsector=$((startsector / 8))
+        size=$((size / 8))
+    fi
```
jlebon, line 28: This looks right, though did you test it on a 4k RHCOS?

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+
+    # validate that the new partition will start at the xfs super block
+    skipcount=$(($startsector * $sector_size))
+    if [ $(dd if=/dev/$parent bs=1 count=4 status=none skip=$skipcount) != XFSB ]; then
+        echo "couldn't detect XFS superblock at expected partition start"
+        exit
```
jlebon, line 34: Is it by design we're not exiting 1 here? Seems like we should so it shows up as a failed service.

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+    fi
+
+    # recreate the partition offset after the LUKS header
+    # we don't want to copy the typeguid because that's pointing to crypt_luks
+    sgdisk -d $partnum -n $partnum:$startsector:+$size -c $partnum:$label /dev/$parent -u $partnum:$guid
+    # delete the rhcos.root karg to prevent unlocking attempts
+    rpm-ostree kargs --delete rhcos.root
```
jlebon, line 41: If we want to match freshly booted 4.7 nodes with the new rootmap code, we should also inject `root=UUID=$filesystem_uuid rw` here (which you can get from `lsblk -no UUID /dev/$partition`).

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+    > /run/device_mod
+    exit
+fi
+
+# pre-4.7 root LUKS devices are always named luks-00000000-0000-4000-a000-000000000002
+if [ -h /dev/mapper/luks-00000000-0000-4000-a000-000000000002 ]; then
+    # check if the rhcos.root karg exists; if it does not then this script has already
+    # modified the kargs and can exit early
+    if [[ ! "$(rpm-ostree kargs)" =~ .*"rhcos.root".* ]]; then
+        exit
```
jlebon, line 52: Should we touch the stamp file here?

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+
+    > /run/device_mod
+    exit
+fi
+
+# pre-4.7 root LUKS devices are always named luks-00000000-0000-4000-a000-000000000002
+if [ -h /dev/mapper/luks-00000000-0000-4000-a000-000000000002 ]; then
+    # check if the rhcos.root karg exists; if it does not then this script has already
+    # modified the kargs and can exit early
+    if [[ ! "$(rpm-ostree kargs)" =~ .*"rhcos.root".* ]]; then
```
jlebon, line 51: Meh, I think the karg is unique enough that we can simplify this to e.g.

```suggestion
    if ! grep -q rhcos.root <(rpm-ostree kargs); then
```

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -0,0 +1,71 @@
...
+    partition=$(ls /sys/block/$dm_device_name/slaves/)
+    luksuuid=$(cryptsetup luksUUID /dev/$partition)
+    fsuuid=$(blkid -s UUID -o value /dev/mapper/luks-00000000-0000-4000-a000-000000000002)
+    clevis_id="$(cryptsetup luksDump /dev/$partition | sed -rn 's|^\s+([0-9]+): clevis|\1|p')"
+    pin=$(cryptsetup token export --token-id "${clevis_id}" "/dev/$partition" \
+            | jq -rM '.jwe.protected' | base64 -d | jq -rM '.clevis.pin')
+
+    set -- --delete rhcos.root --delete rd.luks.options --append root=UUID=$fsuuid --append rw --append rd.luks.name=$luksuuid=luks-$luksuuid
+    if [[ "$pin" != "tpm2" ]]; then
+        set -- "$@" --append rd.neednet=1 --append rd.luks.options=_netdev
```
jlebon, line 65: We should check if `rd.neednet=1` is already specified on the cmdline. It's documented today that to use Tang, users need to add `rd.neednet=1` so in most cases I think this will end up duplicating the karg.

cgwalters (approved):
OK let's give this a try!
/lgtm

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
arithx: Yes I did.

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
arithx: I agree that this could be an `elif` but I think we still want the stamp file. Both cases will technically not do anything but we'd fail the XFS superblock check in the case of nocrypt and in the crypt case we'd exit on `rhcos.root` being missing (if you `systemctl restart rhcos-upgrade-root-filesystem-container` during the same first boot the unit will still be re-ran because we've only staged the kargs update with ostree).

cgwalters:
TBH I have at this point lost track of our vast array of LUKS testing.
Quite a while ago I did this one which is buried in the still-internal Jenkins glue:

```
$ kola run-upgrade --ignition-version v2 -b rhcos -v --find-parent-image --qemu-image-dir tmp/ --output-dir tmp/kola-upgrade
```

which indeed seems to pass with this code :tada:

arithx:
> TBH I have at this point lost track of our vast array of LUKS testing.

For local testing I've been doing the following (for regular non-crypt, non-crypt 4k, & crypt):
1. boot into 4.6
1. do a pull-local upgrade into custom 4.7 image w/ changes
1. systemctl reboot once while in 4.7
1. verify system comes back up and kargs updated / partition table changed

cgwalters:
Yeah the upgrade test there does basically that except starting from 4.2.

cgwalters:
OK a problem this introduced is that the `rpm-ostree kargs` invocation will be "staged" and only actually happen on shutdown (via `ostree-finalize-staged.service`).  Which means that in the following scenario:

- system booted as 4.6
- upgrade to 4.7
- reboot
- <system forcibly powered off>

The kernel command line will be missing the `root=` bits and the luks opener will try to run and fail.

This is a messy topic...we probably need to add some sort of `rpm-ostree kargs --immediate` or just drop to editing the BLS config file directly.

arithx:
> This is a messy topic...we probably need to add some sort of rpm-ostree kargs --immediate or just drop to editing the BLS config file directly.

We'll be moving to editing the BLS directly (via `rdcore`) with the move to the initrd.

PR #438, opened 2020-11-12
FAQ: Where'd my /etc/yum.repos.d go?

cgwalters:
Describe this a bit.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -139,6 +139,25 @@ is in place, as the cluster currently has [no mechanism for reporting](https://g
 fashion.  This kind of package replacement can also leave your nodes exposed to potential problems
 that are fixed in newer versions of the package.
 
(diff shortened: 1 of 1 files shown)


kelvinfan001:
/lgtm
Thanks for this. And just for my understanding, this basically just means that users are *strongly* discouraged to layer new packages right?

cgwalters:
> And just for my understanding, this basically just means that users are strongly discouraged to layer new packages right?

That's
https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md

Note in that design the extensions are still shipped with `machine-os-content`; they're hence still versioned with the release image, and the MCO makes a local repo and hands them to rpm-ostree.

(FCOS however does ship the Fedora package repos enabled by default, that's https://github.com/coreos/fedora-coreos-tracker/issues/401 )

ashcrow (approved):

PR #440, opened 2020-11-17
manifest: make sure clevis is >= 15-1.el8

jlebon:
For RHCOS 4.7, we need to make sure that we use a clevis newer than
what's in 8.3. We're currently cherry-picking from 8.4. Encode that
into the manifest to be more explicit and add a link to the RHBZ.


cgwalters (approved):
/lgtm

PR #443, opened 2020-11-18
rhcos-upgrade-root-filesystem-container: address comments

arithx:
Address some comments that came after the original PR was queued to
merge.

### DIFF
--- a/overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
+++ b/overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euxo pipefail
 
-if [ -f /run/device_mod ]; then
+if [ -f /run/rhcos-rootfs-upgraded ]; then
     exit
 fi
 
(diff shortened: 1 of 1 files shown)


overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
```diff
@@ -10,11 +10,18 @@ if [ -h "/dev/mapper/coreos-luks-root-nocrypt" ]; then
     dm_device=$(readlink /dev/mapper/coreos-luks-root-nocrypt -f)
     dm_device_name=$(basename $dm_device)
     partition=$(ls /sys/block/$dm_device_name/slaves/)
-    parent=$(basename "$(lsblk --noheadings --output PKNAME --paths /dev/$partition | head -n1)")
+    parent=$(lsblk --noheadings --output PKNAME /dev/$partition --nodeps)
     partnum=$(cat /sys/block/$parent/$partition/partition)
+    if [ "$partnum" != "4" ]; then
+        echo "unexpected partition number: $partnum"
+        > /run/rhcos-rootfs-upgraded
```
jlebon, line 17: Hmm, should we really touch the stamp file in error conditions? We didn't actually upgrade anything, right? I think we should drop those so that one *could* rerun the script multiple times if it didn't work all the way through (imagine a debug session where you're iterating on it). This also matches the fact that `set -e` will just exit right away if any of the commands failed.

jlebon (approved):
One comment, LGTM otherwise!

overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
arithx: Fair enough. I'll drop it in the error cases.

PR #444, opened 2020-11-18
manifest: add jlebon-dracut-tmp repo

jlebon:
This is a temporary repo containing a dracut build with the patches we
need for working LUKS.

Will be dropped once we have official builds.


miabbott:
Is it worth an ext test that checks the dracut version pulled in?  That way once the official build is included, we are forced to react and remove the temp repo, etc

jlebon:
Yeah, I like that. Added!

PR #445, opened 2020-11-18
Revert "manifest: add jlebon-dracut-tmp repo"

jlebon:
This reverts commit 5845493ff003dff9c2c0ce432fc75ec1a0e79f7f.

There were concerns about carrying this, even if it's temporary. Let's
revert.


cgwalters (approved):
Thanks.
/lgtm

PR #449, opened 2020-11-20
tests/kola: Lift TPM2 rootfs test from FCOS

jlebon:
This updates the TPM2 test to the new LUKS path.

To dedupe once we have glue to share tests with FCOS.


travier (approved):
LGTM, but it looks like we have a lot of test duplicated between the two repos. Maybe we should find a solution to merge those.

PR #450, opened 2020-11-21
download-extensions: test for hardlinks before creating them

miabbott:
The RHCOS 4.7 pipeline started failing when the `download-extensions`
script tried to create a hardlink that already exists. This changes the
script to test for the presence of the hardlink before trying to
create it.

### DIFF
--- a/scripts/download-extensions
+++ b/scripts/download-extensions
@@ -4,7 +4,7 @@
 # and https://github.com/coreos/fedora-coreos-tracker/issues/401
 # In the future this might live in rpm-ostree, see https://github.com/coreos/rpm-ostree/issues/2055
 set -euo pipefail
-destdir=$1
+destdir=$(realpath "${1}")
 shift
 reposdir=$(pwd)/src/config
 cd "${destdir}"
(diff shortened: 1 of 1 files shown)


miabbott:
Lightly tested this locally, but I think it should work.

Since I'll be on PTO next week, would appreciate if someone bumped the submodule on `redhat-coreos` internally when this merges.

ashcrow (approved):

scripts/download-extensions
```diff
@@ -64,7 +73,9 @@ get_rtheaders() {
     info "Retrieving ${headers} for ${rtrpm}"
     yumdownload "${headers}"
     headersrpm=$(ls kernel-headers*)
-    ln "${headersrpm}" ../..
+    if [ ! -f "${destdir}/extensions/${headersrpm}" ]; then
+        ln "${headersrpm}" "${destdir}/extensions/"
+    fi
```
jlebon, line 78: Hmm, I don't think we actually need this since we're always creating the `kernel-rt` extension first, so we should never hit this case. But WFM as is.

jlebon (approved):
Makes sense to me. Let's unblock the pipeline!
/lgtm

PR #451, opened 2020-11-23
manifest: Add qemu-guest-agent

cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1900759

Basically MCO is buried in PRs, shipping it via a container
is would be another special case there.

Shipping it in RHCOS by default mirrors what we do for the
vmware agent today.  This is important for RHV and will help
with a lot of bugs they have around machineAPI provisioning.

I think OKD will also do the same (package layer it in
machine-os-content built time).
Not proposing we ship it in base FCOS at this time.

Longer term I think we should investigate switching both
this and vmware to be package layered "internally" so
we're not including it obviously on e.g. AWS/Azure instances.

But the potential for conflict is basically zero; the
way the agent works is via a udev rule that keys off
the presence of the virtio socket.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -254,6 +254,8 @@ packages:
  - authselect
  # Dracut module dependencies
  - bsdtar
+ # https://bugzilla.redhat.com/show_bug.cgi?id=1900759
+ - qemu-guest-agent
  # BELOW HERE ARE PACKAGES NOT IN RHEL
  # OpenShift OKD
  #- origin-node origin-hyperkube origin-clients
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -105,6 +105,12 @@ elif [[ $nm_ts -gt $switchroot_ts ]] && on_platform aws; then
 fi
 echo ok conditional initrd networking
 
+# Verify this is shipped
+if ! test -f /usr/bin/qemu-ga; then
+  fatal "missing qemu guest agent"
+fi
+echo ok qemu guest agent
+
 case "$(arch)" in
     x86_64|aarch64)
         if runuser -u core -- ls /boot/efi &>/dev/null; then


ashcrow:
I really don't want to have agents in the base RHCOS images but it is true we had to make an exception for vmware's agent.

> Longer term I think we should investigate switching both this and vmware to be package layered "internally" so we're not including it obviously on e.g. AWS/Azure instances.

Ok. I think this is reasonable. As long as this isn't opening the door to more agents or allowing management of the cluster nodes outside of OCP it's a sane first step between here and package layering/containerization.

LorbusChris:
> I think OKD will also do the same (package layer it in
> machine-os-content built time).
> Not proposing we ship it in base FCOS at this time.

I don't think we currently do in OKD's machine-os-content, but if this is added here, we'll follow along (we certainly already put the vmware agent in there).
cc @vrutkovs 

> Longer term I think we should investigate switching both
> this and vmware to be package layered "internally" so
> we're not including it obviously on e.g. AWS/Azure instances.

:+1: 

> I really don't want to have agents in the base RHCOS images

:+1: :+1:

cgwalters:
Any further thoughts on this?

ashcrow:
cc @miabbott @darkmuggle @bgilbert @jlebon

ashcrow (approved):
I'm OK with this given the context and eventually moving this out to a container/ds/etc..

cgwalters:
Can someone add lgtm?  Thanks!

jlebon:
> Longer term I think we should investigate switching both
this and vmware to be package layered "internally" so
we're not including it obviously on e.g. AWS/Azure instances.

Couldn't we do something similar pretty easily by shipping it as an extension that the MCO auto-enables on the appropriate platform?

cgwalters:
> Couldn't we do something similar pretty easily by shipping it as an extension that the MCO auto-enables on the appropriate platform?

Yes, not really difficult, but runs into the first bit:

> Basically MCO is buried in PRs,

This is a compromise to avoid changing the MCO right now.

bgilbert:
> Longer term I think we should investigate switching both this and vmware to be package layered "internally" so we're not including it obviously on e.g. AWS/Azure instances.

From very early on, the design has been that we ship the same image on all platforms to reduce maintenance overhead.  Is there a substantial harm to doing so?

jlebon:
> Yes, not really difficult, but runs into the first bit:
>
>> Basically MCO is buried in PRs,

OK, gotcha. It seems like the trade-offs here between doing this short-term vs a more proper solution have already been discussed. So,

/approve

Will leave it to @bgilbert do the final `/lgtm`.

cgwalters:
> From very early on, the design has been that we ship the same image on all platforms to reduce maintenance overhead. Is there a substantial harm to doing so?

I'd agree that changing this would have a lot of consequences.  Actually I'd go further and say it's not worth doing until we have the infrastructure in place (mostly in the MCO) to e.g. avoid rolling updates to an AWS cluster when `open-vm-tools` has a CVE we need to fix.  zincati I think already implicitly handles this because the necessary logic for upgrades lives solely in rpm-ostree.

miabbott (approved):
Nothing really to add here; the tradeoffs are understood.  Giving Benjamin the `lgtm` button

PR #452, opened 2020-11-23
.gitmodules: add branch for fedora-coreos-config submodule

jlebon:
Modern git allows specifying a branch in `.gitmodules` so that bumping
it can be done with `git submodule update --remote fedora-coreos-config`
which is nicer than manually fetching and checking out in the submodule
directory.


lucab (approved):
/lgtm

PR #454, opened 2020-11-24
manifest: add ppc64le packages to detect dynamically attached devices

Prashanth684:
same as https://github.com/coreos/fedora-coreos-config/pull/296

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -279,6 +279,9 @@ packages-x86_64:
 
 packages-ppc64le:
   - irqbalance
+  - librtas
+  - powerpc-utils-core
+  - ppc64-diag-rtas
 
 remove-from-packages:
   - - filesystem


jlebon:
Is this missing the same change to presets as the FCOS one?

cgwalters:
One option is to split the manifest bits for this into a separate include that can be shared between the two.  We're a bit overdue to try to extract something like a `fedora-coreos-core.yaml` or so.

But OK as is too.
/approve

Prashanth684:
> Is this missing the same change to presets as the FCOS one?

so isn't  `overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset` common for rhcos too? Unfortunately i don't have a power machine to test on now..but i was thinking that the preset file was common.

jlebon:
> so isn't `overlay.d/05core/usr/lib/systemd/system-preset/40-coreos.preset` common for rhcos too?

Yes, you're right. Thanks!

/lgtm

PR #455, opened 2020-11-30
extensions.yaml: New declarative file for extensions

cgwalters:
I was trying to upload a custom oscontainer with a patched
rpm-ostree to help someone test it:
https://bugzilla.redhat.com/show_bug.cgi?id=1865839

And apparently made it incorrectly; doing the build involves
copy-pasting stuff from the internal Jenkins glue.

Let's accomplish two things:

- Make the list of extensions properly declarative, distinct
  from the code implementing it
- Move the logic for sticking it into the oscontainer as
  part of coreos-assembler (making this whole repo more declarative)
  And also therefore making oscontainer generation more ergonomic
  to run locally.

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -1,0 +1,35 @@
+# RPMs as operating system extensions, distinct from the base ostree commit/image
+# https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md
+# and https://github.com/coreos/fedora-coreos-tracker/issues/401
+# We currently explicitly list dependencies for each extension; see
+# https://github.com/coreos/rpm-ostree/issues/2055
+
+extensions:
+  # https://github.com/coreos/fedora-coreos-tracker/issues/326
+  usbguard:
+    packages:
+      - usbguard
+      - libqb
+      - protobuf
+  # kernel-devel
+  # https://github.com/kmods-via-containers/kmods-via-containers/issues/3
+  # https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/866
+  kernel-devel:
+    packages:
+      - kernel-devel
+      - kernel-core
+      - kernel-headers
+      - kernel-modules
+      - kernel-modules-extra
+  # GRPA-2822
+  # https://github.com/openshift/machine-config-operator/pull/1330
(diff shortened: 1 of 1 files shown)


cgwalters:
Pairs with https://github.com/coreos/coreos-assembler/pull/1916

After we land both we can delete the shell script implementation.

(Though I currently skipped over the special RT kernel versioning handling)

ashcrow (approved):
I like it!

miabbott (approved):
Great idea!

extensions.yaml
```diff
@@ -0,0 +1,35 @@
...
+  # https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/866
+  kernel-devel:
+    - kernel-devel
+    - kernel-core
+    - kernel-headers
+    - kernel-modules
+    - kernel-modules-extra
+
+# Extensions only available on a particular architecture
+extensions-archdep:
```
jlebon, line 24: If we're defining a new format anyway, WDYT about folding this into `extensions` instead and make them each richer objects? E.g.

```yaml
extensions:
  usbguard:
    arches: [...]         # optional; defaults to all arches
    exclude_arches: [...] # optional; defaults to none
    packages: [...]       # required
  ...
```

Feels more "YAML-y".

Good as is too though if you want to tweak the format separately (though it's easier to do before it's in production).

jlebon (approved):
One comment, LGTM as is!

extensions.yaml
jlebon: This would be useful for example when adding more specialized things like matching kernel-headers versions with the one in the host. Something like:

```yaml
kernel-devel:
  packages:
    - kernel-devel
    - kernel-core
    - kernel-headers
    - kernel-modules
    - kernel-modules-extra
  match-evra-of-base-pkg: kernel
```

And then the script would know that when calling `yumdownload` it would append the EVRA of the base package `kernel` to the CLI invocation.

cgwalters: Definitely, it's worth getting this format "right" since it becomes an API/ABI of sorts - not impossible to change but still.

I like the first suggestion.  The kernel thing seems way too special case, I'd vote to just hardcode the logic for that one.  If we could think of another case where we need EVRAs matching perhaps it could be declarative.

Issue #417, opened 2020-10-20
Do a basic build + kola run on PRs


cgwalters:
Previously: https://github.com/openshift/release/issues/4892#issuecomment-531939430 (and see also [this doc](https://github.com/coreos/coreos-assembler/blob/7fc90f45d5fbf45d39441339ac37173dbeb70041/docs/working.md#running-coreos-assembler-in-openshift-on-google-compute-platform))

The biggest change since then is that there's a "build02" cluster in GCP that is apparently born in 4.4.  Since [this openshift/installer commit](https://github.com/openshift/installer/pull/3808/commits/1f2d071f07ffd9e12fb9df64faadc3627db55bbd) RHCOS has enabled the nested virt license by default.

So we need to do one of four things:

- Reprovision build02 as 4.6
- Add a new build03 in GCP from 4.6
- Add a machineset with AWS `i3.metal` nodes (expensive but very powerful)
- Edit the machinesets for build02 in place (not hard but one of the above would be preferable)

After that we need to [Enable kubevirt](https://kubevirt.io/quickstart_cloud/) (or [openshift virtualization](https://docs.openshift.com/container-platform/4.5/virt/about-virt.html) )

cgwalters:
Update: build02 now has nested virt enabled on at least one worker.

Next required step: Get them to run https://github.com/cgwalters/kvm-device-plugin

cgwalters:
And after we have basic sanity checking, the obvious next step here is a "tier 2" possibly opt-in test that also creates a cluster.

PR #456, opened 2020-12-01
05rhcos/*: Fail boot if Ignition uses an older syntax for LUKS in 4.7

sohankunkerkar:
This change ensures that the user provides the correct LUKS syntax in an Ignition config while trying to enable LUKS in 4.7.

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/module-setup.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+install_unit() {
+    local unit="$1"; shift
+    local target="${1:-ignition-complete.target}"; shift
+    local instantiated="${1:-$unit}"; shift
+    inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
+    mkdir -p "$initdir/$systemdsystemunitdir/$target.requires"
+    ln_r "../$unit" "$systemdsystemunitdir/$target.requires/$instantiated"
+}
+
+install() {
+    inst_script "$moddir/rhcos-fail-boot-for-legacy-luks-config" \ 
+        "/usr/libexec/rhcos-fail-boot-for-legacy-luks-config"
+    
+    install_unit rhcos-fail-boot-for-legacy-luks-config.service
+}
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script tries to look for a `files` entry with the `/etc/clevis.json` 
+# path (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists, then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[]? | select(.path==\"${wanted_path}\")" "${ign_config}" > /dev/null; then
+    echo "Your Ignition config specifies LUKS filesystem encryption using the obsolete
+${wanted_path} config file, which is no longer supported. Refusing to boot. 
+Please refer to https://github.com/openshift/openshift-docs/pull/27661 for more 
+information."
+    exit 1
+fi
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config.service
@@ -1,0 +1,16 @@
+# Fail the boot if the user tries to use the legacy 
+# LUKS configuration in an Ignition config to enable 
+# LUKS in RHCOS, so they don't accidentally end up 
+# with an unencrypted system.
+[Unit]
+Description=RHCOS Check For Legacy LUKS Configuration
+Documentation=https://github.com/openshift/openshift-docs/pull/27661
+After=ignition-fetch.service
+Before=ignition-disks.service
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/rhcos-fail-boot-for-legacy-luks-config
+RemainAfterExit=yes


overlay.d/05rhcos/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,23 @@
...
+tmpconfig=${outputdir}/tmpconfig.json
+
+source /etc/os-release
+if [ "$VERSION_ID" == "4.7" ]; then
+    # select the `/etc/clevis.json` entry from a given Ignition config
+    (jq ".storage.files[] | select(.path==\"${wanted_path}\")") \
+        < "${ign_config}" > ${tmpconfig}.fragment
+
+    if [ -s ${tmpconfig}.fragment ]; then
+        echo "Failed to boot due to an older syntax used to configure LUKS"
```
cgwalters, line 20: I think we should get a PR into the OpenShift docs to describe this and change this to hyperlink to that.

overlay.d/05rhcos/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,23 @@
...
+# This script tries to look for a `/etc/clevis.json` reference 
+# (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists and the version of RHCOS is 4.7, 
+# then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+tmpconfig=${outputdir}/tmpconfig.json
+
+source /etc/os-release
```
cgwalters, line 13: This is OK, but `/usr/lib/os-release` is preferred by default.  (But see below)

overlay.d/05rhcos/usr/lib/systemd/system/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,15 @@
+# Fixes the encryption issue with LUKS in 4.7 when a user 
+# tries to use the RHCOS 4.6 or earlier LUKS syntax in an 
+# Ignition config.
+[Unit]
+Description=Fail boot if Ignition uses an older syntax for enabling LUKS in 4.7 
+Before=ignition-disks.service
+After=ignition-fetch.service
```
cgwalters, line 7: It looks like this is intending to run in the initramfs, but it's actually running in the real root.  It will need to become a dracut module (like some of the other ones) to make that work.

overlay.d/05rhcos/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,23 @@
...
+if [ "$VERSION_ID" == "4.7" ]; then
```
cgwalters, line 14: We still want to emit this error in e.g. 4.8 right?  I think if we move this to be properly in the initramfs and also if the unit does `ConditionKernelCommandLine=ignition.firstboot` we don't need a version check at all, because the unit will only run on the first boot.

cgwalters:
For quickly iterating on changes to the initramfs I'd try out [buildinitramfs-fast](https://coreos.github.io/coreos-assembler/working/#using-cosa-buildinitramfs-fast) btw!

overlay.d/05rhcos/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
bgilbert: Should probably be added to https://github.com/openshift/openshift-docs/pull/27661.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,20 @@
...
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+tmpconfig=${outputdir}/tmpconfig.json
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+(jq ".storage.files[] | select(.path==\"${wanted_path}\")") \
+    < "${ign_config}" > ${tmpconfig}.fragment
+
+if [ -s ${tmpconfig}.fragment ]; then
```
jlebon, line 17: Hmm, I think we can avoid a tmpconfig entirely and just use `jq -e`. E.g. something like:

```bash
if jq -e ".storage.files[] | select(.path==\"${wanted_path}\")"; then
    ...
fi
```

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/module-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+    local target="${1:-ignition-complete.target}"; shift
+    local instantiated="${1:-$unit}"; shift
+    inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
+    mkdir -p "$initdir/$systemdsystemunitdir/$target.requires"
+    ln_r "../$unit" "$systemdsystemunitdir/$target.requires/$instantiated"
+}
+
+install() {
+    inst_script "$moddir/rhcos-fail-boot-for-older-luks-syntax" \ 
+        "/usr/sbin/rhcos-fail-boot-for-older-luks-syntax"
```
jlebon, line 18: This path should match the one in the service unit (so `/usr/libexec/...`).

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,16 @@
+# Fixes the encryption issue with LUKS in 4.7 when a user 
+# tries to use the RHCOS 4.6 or earlier LUKS syntax in an 
+# Ignition config.
+[Unit]
+Description=Fail boot if Ignition uses an older syntax for enabling LUKS in 4.7 
+Before=ignition-disks.service
+After=ignition-fetch.service
+ConditionKernelCommandLine=ignition.firstboot
```
jlebon, line 8: The service is already activated only on firstboot via `ignition-complete.target`, so we don't really need this (doesn't hurt either of course, though might as well be consistent with other services like this).

jlebon, line 8: Not sure what the state of target unit failure modes is in el8 systemd, though in Fedora we've had to spam

```
OnFailure=emergency.target
OnFailureJobMode=isolate
```

in constituent service units to be sure. Probably safer to just add it.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,16 @@
...
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
+RemainAfterExit=yes
+
+[Install]
+WantedBy=multi-user.target
```
jlebon, line 16: We shouldn't need this since it's statically enabled via `ignition-complete.target`.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,16 @@
+# Fixes the encryption issue with LUKS in 4.7 when a user 
+# tries to use the RHCOS 4.6 or earlier LUKS syntax in an 
+# Ignition config.
+[Unit]
+Description=Fail boot if Ignition uses an older syntax for enabling LUKS in 4.7
```
jlebon, line 5: WDYT about something like:

> Description=RHCOS Check For Legacy LUKS Configuration

?

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
sohankunkerkar: > Hmm, I think we can avoid a tmpconfig entirely and just use `jq -e`. E.g. something like:
> 
> ```shell
> if jq -e ".storage.files[] | select(.path==\"${wanted_path}\")"; then
>     ...
> fi
> ```
@jlebon  I think we should provide an input i.e. `ign_config` in this case, right?

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/module-setup.sh
sohankunkerkar: ah, good catch!

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
sohankunkerkar: It makes sense to me!

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,15 @@
+# Fixes the encryption issue with LUKS in 4.7 when a user 
+# tries to use the RHCOS 4.6 or earlier LUKS syntax in an 
+# Ignition config.
+[Unit]
+Description=RHCOS Check For Legacy LUKS Configuration 
+Before=ignition-disks.service
+After=ignition-fetch.service
```
bgilbert, line 7: I think it's clearer to list `After` dependencies first so the dependencies are in chronological order.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,15 @@
...
+OnFailure=emergency.target
+OnFailureJobMode=isolate
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/rhcos-fail-boot-for-older-luks-syntax
+RemainAfterExit=yes
+
```
bgilbert, line 15: nit: extra line

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
```diff
@@ -0,0 +1,15 @@
+# Fixes the encryption issue with LUKS in 4.7 when a user 
+# tries to use the RHCOS 4.6 or earlier LUKS syntax in an 
+# Ignition config.
```
bgilbert, line 3: "Fixes the encryption issue" doesn't really explain anything.  In fact, what we're doing is failing the boot if the user tries to do that, so they don't accidentally end up with an unencrypted system.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script tries to look for a `/etc/clevis.json` reference 
+# (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists and the version of RHCOS is 4.7, 
+# then the script ends with exit 1.
```
bgilbert, line 7: We're no longer checking the RHCOS version.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+set -euo pipefail
+
+# This script tries to look for a `/etc/clevis.json` reference
```
bgilbert, line 4: `reference` is a bit unclear.  It looks for a `files` entry with that path.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,16 @@
...
+# (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists and the version of RHCOS is 4.7, 
+# then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[] | select(.path==\"${wanted_path}\")" "${ign_config}";  then
+    echo "Failed to boot due to an older syntax used to configure LUKS"
```
bgilbert, line 14: Let's make sure we give specific information to the user.  Maybe something like "Your Ignition config specifies LUKS filesystem encryption using the obsolete ${wanted_path} config file, which is no longer supported.  Refusing to boot."?

(wrapped to 80 characters, of course)

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
```diff
@@ -0,0 +1,16 @@
...
+# This script tries to look for a `/etc/clevis.json` reference 
+# (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists and the version of RHCOS is 4.7, 
+# then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[] | select(.path==\"${wanted_path}\")" "${ign_config}";  then
```
bgilbert, line 13: If there is no `storage.files` section, this will still work correctly (return non-zero) but will produce a confusing error in the logs:

```
jq: error (at /run/ignition.json:1): Cannot iterate over null (null)
```

sohankunkerkar: ah, good catch. I think this should work: `.storage.files[]?`

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
sohankunkerkar: How about `Legacy LUKS configuration is no longer supported. Refusing to boot.`?

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax.service
sohankunkerkar: Fixed

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-older-luks-syntax
bgilbert: That could work, but the longer string is clearer, I think.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config
```diff
@@ -0,0 +1,18 @@
...
+# This script tries to look for a `files` entry with the `/etc/clevis.json` 
+# path (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists, then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[]? | select(.path==\"${wanted_path}\")" 
+"${ign_config}";  then
```
bgilbert, line 13: Bad line-wrap?

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config
```diff
@@ -0,0 +1,18 @@
...
+
+# This script tries to look for a `files` entry with the `/etc/clevis.json` 
+# path (used to enable LUKS in RHCOS 4.6 or earlier versions) in an 
+# Ignition config. If it exists, then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[]? | select(.path==\"${wanted_path}\")"
```
bgilbert, line 12: This will print the matching entry to stdout if it finds one.

jlebon (approved):
LGTM overall!

Re. link, let's just link to https://github.com/openshift/openshift-docs/pull/27661 for now in the error message until it gets merged and we get a nicer one?

Probably also worth adding it as a `Documentation=` line too in the service unit.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-check-luks-syntax/rhcos-fail-boot-for-legacy-luks-config
```diff
@@ -0,0 +1,18 @@
...
+# Ignition config. If it exists, then the script ends with exit 1.
+
+ignition_cfg="/run/ignition.json"
+wanted_path="/etc/clevis.json"
+
+# select the `/etc/clevis.json` entry from a given Ignition config
+if jq -e ".storage.files[]? | select(.path==\"${wanted_path}\")" "${ign_config}" > /dev/null; then
+    echo "Your Ignition config specifies LUKS filesystem encryption using the obsolete
+${wanted_path} config file, which is no longer supported. Refusing to boot. 
+Please refer https://github.com/openshift/openshift-docs/pull/27661 for more
```
bgilbert, line 15: `Please refer to`

bgilbert (approved):
LGTM!

cgwalters:
This needed https://github.com/openshift/os/pull/463 at least to run.

cgwalters:
(Note there's no useful CI here right now; that's https://github.com/openshift/os/issues/417 and it's being worked on)

PR #457, opened 2020-12-02
kola-denylist: re-enable Tang tests

jlebon:
Dracut patches needed for these landed in 4.7, so these tests pass now.


cgwalters (approved):
/lgtm
Awesome work on following through on this whole chain!

PR #459, opened 2020-12-03
scripts: Remove download-extensions

cgwalters:
There's some logic I didn't currently preserve from this
in the move into coreos-assembler around kernel/kernel-headers
matching.
https://github.com/coreos/coreos-assembler/pull/1916

I will look at re-adding that.  But short term we don't
actually have that problem right now and I'd like to
get the 4.7 pipeline switched over to
the new model so we don't have duplication.

The goal here is to make it very "cosa like" to upload the
oscontainer because it needs to be trivial for other people
to do outside of a pipeline.

Because this is git, when we need some of this code again we
can find it in the history.


cgwalters:
(Also scripts/verify-kernel-versions should be folded into; something more declarative like `allow-skew: true|false` in the `kernel-rt` extension section)

PR #462, opened 2020-12-08
Revert to base 8.3 rpm-ostree

cgwalters:
We need to tag more packages to make this work.

### DIFF
--- a/rhcos-packages.yaml
+++ b/rhcos-packages.yaml
@@ -15,7 +15,7 @@ packages:
   # needed by rpm-ostree for managing users
   - nss-altfiles
   - ostree
-  - "'rpm-ostree >= 2020.7'"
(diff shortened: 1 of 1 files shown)


rhcos-packages.yaml
```diff
@@ -15,7 +15,7 @@ packages:
   # needed by rpm-ostree for managing users
   - nss-altfiles
   - ostree
-  - "'rpm-ostree >= 2020.7'"
+  - rpm-ostree-2020.7-1.el8
```
travier, line 18: Don't we need `2020.4` so `<2020.7`?

cgwalters: Yep done, thanks!

ashcrow (approved):

PR #463, opened 2020-12-08
rhcos-check-luks: Fix trailing space and script variable

cgwalters:
The trailing space causes boot failure in the initramfs because the script
is missing.  Also the variable name was incorrect.


bgilbert:
Ouch.  Thanks for tracking this down.

cc @sohankunkerkar 

/lgtm

sohankunkerkar:
@cgwalters Thanks for catching this issue.

PR #469, opened 2020-12-14
overlay.d: Add 20rhcos-kdump

kelvinfan001:
The latest Dracut release that landed in RHEL 8.3 stopped creating
`$systemdsystemunitdir/initrd.target.wants` dir, causing
`kdump-capture.service` to not be installed correctly and thus making
kdump unable to capture the vmcore during a kernel panic.
There is a patch in kexec-tools to address this issue
https://bugzilla.redhat.com/show_bug.cgi?id=1907253.
For now, we will create this directory ourselves, but we should drop
this once the proper kexec-tools patches land.

### DIFF
--- a/overlay.d/20rhcos-kdump/usr/lib/dracut/modules.d/50rhcos-kdump/module-setup.sh
+++ b/overlay.d/20rhcos-kdump/usr/lib/dracut/modules.d/50rhcos-kdump/module-setup.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+# The latest Dracut release that landed in RHEL 8.3 stopped creating
+# `$systemdsystemunitdir/initrd.target.wants` dir, causing
+# `kdump-capture.service` to not be installed correctly and thus making
+# kdump unable to capture the vmcore during a kernel panic.
+# There is a patch in kexec-tools to address this issue
+# https://bugzilla.redhat.com/show_bug.cgi?id=1907253.
+# For now, we will create this directory ourselves, but we should drop
+# this once the proper kexec-tools patches land.
+
+check() {
+    return 0
+}
+
+depends() {
+    return 0
+}
+
+install() {
+    mkdir -p "$initdir/$systemdsystemunitdir/initrd.target.wants"
+}
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -132,3 +132,10 @@ if ! test -f /etc/iscsi/initiatorname.iscsi; then
     fatal "Missing /etc/iscsi/initiatorname.iscsi"
 fi
 echo "ok iSCSI initiator name"
+
+# Check whether the fixed version of kexec-tools has landed.
+# xref: https://bugzilla.redhat.com/show_bug.cgi?id=1907253
+if grep "add-wants" /usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh; then
+  echo "kexec-tools has been updated; remove workaround from https://github.com/openshift/os/pull/469"
+  sleep 2
+fi


miabbott:
Is it possible to add a test for the fixed version of `kexec-tools` so that when it finally lands we are forced to take action?

Something like:

```diff
diff --git a/tests/kola/misc-ro/misc-ro.sh b/tests/kola/misc-ro/misc-ro.sh
index 36b4061..2c721f7 100755
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -132,3 +132,7 @@ if ! test -f /etc/iscsi/initiatorname.iscsi; then
     fatal "Missing /etc/iscsi/initiatorname.iscsi"
 fi
 echo "ok iSCSI initiator name"
+
+if grep "add-wants" /usr/lib/dracut/modules.d/99kdumpbase/module-setup.sh; then
+    fatal "kexec-tools has been updated; remove workaround from fedora-coreos-config #469"
+fi
```

kelvinfan001:
SGTM. Let me add a test for that.

overlay.d/20rhcos-kdump/usr/lib/dracut/modules.d/50rhcos-kdump/module-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+check() {
+    return 0
+}
+
+depends() {
+    return 0
+}
+
+install() {
+    mkdir -p "$initdir/$systemdsystemunitdir/initrd.target.wants"
```
jlebon, line 20: Wouldn't it serve the purpose to do

```suggestion
    mkdir "$initdir/$systemdsystemunitdir/initrd.target.wants"
```

instead of the added `misc-ro` test?

Then the compose will naturally fail if the directory already exists.

jlebon: We'd need to order it after 99kdumpbase though, which... is annoyingly already using `99`. I guess `99kdumpbase-rhcos`?

kelvinfan001: > Then the compose will naturally fail if the directory already exists.

What if another module added the `initrd.target.wants` directory before the patch in `kexec-tools` lands? Maybe it's not a big deal, but would it be better to have the "reminder mechanism" more directly tied to the fix in `kexec-tools`?

kelvinfan001: > We'd need to order it after 99kdumpbase though

Wouldn't we want to order _before_, so the directory exists when the kdump modules are installed?

jlebon: Ahh I see, I thought the failure manifested itself at runtime. But you're saying the dracut module won't even install the service if the directory is missing?

In that case, yes you're right. :+1:

kelvinfan001: Yep, otherwise the very important `kdump-capture.service` won't be installed and enabled in the kdump initramfs, but `kdump.service` (to enable kdump) deceivingly does not fail and a bad kdump initramfs is generated, so when a kernel panic happens and we boot into the kdump initramfs, the vmcore will not be captured.

cgwalters:
Failing the build seems a bit harsh for this as a general pattern; I probably would have more gone for "emit a notice and sleep 2 seconds" or so.

cgwalters (approved):
/lgtm

PR #471, opened 2020-12-14
overlay.d: Use `systemctl {add-wants, add-requires}`

kelvinfan001:
Use `systemctl`'s built-in way of adding `Wants=` and `Requires=`
dependencies, since this is the recommended way of adding dependencies
and we can do so with a single command.


cgwalters (approved):
Looks like a nice cleanup to me!

cgwalters:
At present there isn't any CI on this repo - did you test this locally?

travier (approved):

kelvinfan001:
I picked a couple services and used `journalctl -u` to see if they ran, and also `systemctl status` to check that they were enabled, and they seemed to be working.

PR #472, opened 2020-12-17
Revert "overlay.d: Add 20rhcos-kdump"

kelvinfan001:
Reverts openshift/os#469

This work around is no longer needed, since we've reverted to using `dracut-049-95.git20200804.el8_3.4`, which does not break kdump.


kelvinfan001:
cc @jlebon

kelvinfan001:
```
# No longer have workaround dracut module
$ ls /usr/lib/dracut/modules.d/ | grep 50rhcos-kdump
# Have the correct dracut version that does not break kdump
$ rpm -q dracut
dracut-049-95.git20200804.el8_3.4.x86_64
# vmcore successfully captured
$ ls /var/crash
127.0.0.1-2020-12-17-16:37:25
```

jlebon:
Thanks!
/lgtm

PR #473, opened 2020-12-17
coreos-generate-iscsi-initiatorname.service: drop ConditionFirstBoot=true

jlebon:
So that even nodes upgrading that hit the previous bug where we wouldn't
generate the iSCSI initiator name get it fixed.

It also matches what the upstream one which will obsolete this does:

https://github.com/open-iscsi/open-iscsi/blob/b680f6e81f2f05f7e721f0aa97ce8aa885b3f0ba/etc/systemd/iscsi-init.service

See: https://bugzilla.redhat.com/show_bug.cgi?id=1901021#c20


cgwalters (approved):
/lgtm

miabbott:
The person that made the comment https://bugzilla.redhat.com/show_bug.cgi?id=1901021#c20 has opened a new downstream BZ for the problem this solves - https://bugzilla.redhat.com/show_bug.cgi?id=1908830

PR #475, opened 2020-12-18
Revert "Revert to base 8.3 rpm-ostree"

miabbott:
We now have `rpm-ostree-2020.7-1.el8_3` available in the 4.7 plashet,
so we can stop using the older version.

Slightly modified the revert to drop any version comparison.

This reverts commit 7d997ddf614ddb06982598ea8e9868757117764c.

### DIFF
--- a/rhcos-packages.yaml
+++ b/rhcos-packages.yaml
@@ -15,7 +15,7 @@ packages:
   # needed by rpm-ostree for managing users
   - nss-altfiles
   - ostree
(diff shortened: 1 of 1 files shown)


jlebon:
Ohhh, [you beat me!](https://github.com/openshift/os/pull/476) :)
/lgtm

jlebon:
Actually, let's just drop any version comparison entirely and rely on highest NEVRA wins?

miabbott:
> Actually, let's just drop any version comparison entirely and rely on highest NEVRA wins?

Sure, lemme fix it up

PR #478, opened 2020-12-18
tests/kola: add NetworkManager version check

dustymabe:
Will keep us reminded of any actions we need to take when
NetworkManager changes.

### DIFF
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -132,3 +132,11 @@ if ! test -f /etc/iscsi/initiatorname.iscsi; then
     fatal "Missing /etc/iscsi/initiatorname.iscsi"
 fi
 echo "ok iSCSI initiator name"
+
+
+# Let's make sure the NetworkManager we use is one of the one-off
+# rebuilds while we're following RHEL 8.3.
+if [[ ! $(rpm -q NetworkManager) =~ 'rhaos4.7' ]]; then
+    fatal "NetworkManager package changed from rhaos4.7 branch. Needs investigation."
+fi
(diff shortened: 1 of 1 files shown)


tests/kola/misc-ro/misc-ro.sh
```diff
@@ -132,3 +132,10 @@ if ! test -f /etc/iscsi/initiatorname.iscsi; then
     fatal "Missing /etc/iscsi/initiatorname.iscsi"
 fi
 echo "ok iSCSI initiator name"
+
+
+# Let's make sure the NetworkManager we use is the one we're expecting
+if [ $(rpm -q NetworkManager) != "NetworkManager-1.26.0-12.1.rhaos4.7.el8.x86_64" ]; then
```
jlebon, line 138: Could we just check for `rhaos4.7` instead so that we don't have to update the pin each time, but we still make sure that we're not pulling in the `baseos` one? (That's what we really care about, right?)

cgwalters:
Also I'd like to add first-class support to rpm-ostree for something like this in a treefile:

```
packages-query:
  - name: NetworkManager
    repo: rhaos-rhel-8
```

which would be like a YAML encoding that would compile down into libdnf/libsolv filtering to say that that package must come from the rpm-md repo named `rhaos-rhel-8`.

tests/kola/misc-ro/misc-ro.sh
dustymabe: yeah, I honestly wanted it to be a little more stringent just so we could make sure things didn't go awry (while we're on 8.3), but I could see relaxing the check just for `rhaos4.7`.

dustymabe: new upload with relaxed check

PR #479, opened 2021-01-05
live: add karg embed areas in GRUB/ISOLINUX configs

jlebon:
(Copied from https://github.com/coreos/fedora-coreos-config/pull/545.)

This provides a buffer into which `coreos-installer iso embed-kargs` can
write additional kernel arguments for the live ISO. It must be located
right after the karg line in configuration files.

Note this doesn't touch `zipl.prm`, which is where ISO kernel args are
stored for s390x. Supporting this for s390x would need more
investigation and testing.

Closes: https://github.com/coreos/coreos-assembler/issues/1988


jlebon:
OK, tested working! This is ready to go now.

arithx (approved):
/lgtm

PR #480, opened 2021-01-05
Bump fedora-coreos-config for various fixes

jlebon:
Specifically,
- the ESP unraiding change (coreos/fedora-coreos-config#794)
- working around buggy udev (coreos/fedora-coreos-config#779)
- --copy-network regression (coreos/fedora-coreos-config#800)
- Azure udev rules in initramfs (coreos/fedora-coreos-config#786)

For the latter, also apply the same `|| exit 1` change from
coreos/fedora-coreos-config#800 to our own
dracut modules for good measure.

```
$ git shortlog --invert-grep --author='CoreOS Bot' 6c9f15c8857772f730cdb0b5b2df143e778c5d0d..8c07b7391473910ba3884ee0d3763743805ac78f
Benjamin Gilbert (6):
      40ignition-ostree: silence mkfs.xfs
      coreos-boot-mount-generator: stop mounting /boot/efi
      40ignition-ostree: rename mount_and_restore_filesystem
      40ignition-ostree: create mountpoint in mount_verbose
      40ignition-ostree: add get_partlabels_for_parttype helper
      40ignition-ostree: copy ESP contents as independent filesystems

Dusty Mabe (1):
      overrides: drop graduated overrides

Jonathan Lebon (7):
      40ignition-ostree/transposefs: also trigger udev on by-label link mismatch for ESP
      40ignition-ostree/transposefs: factor out function to restore fs
      40ignition-ostree/transposefs: relabel the root of reprovisioned filesystems
      40ignition-ostree/transposefs: load zram with num_devices=0
      Add rawhide repo file
      05core: re-order and rename some dracut modules
      05core: add `|| exit 1` to `systemctl add-{requires,wants}` calls

Kelvin Fan (1):
      overlay.d/15fcos: Stop utilizing c-l-h-m private dir

Luca BRUNO (1):
      overrides: drop stale Afterburn entries

Micah Abbott (2):
      use WALinuxAgent-udev package for Azure udev rules
      overlay: add new module for installing Azure udev rules

Prashanth Sundararaman (1):
      tests: Enable TPM test for all arches except s390x

Timothée Ravier (1):
      manifests/fedora-coreos-base: Mask systemd-repart
```

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/module-setup.sh
@@ -19,5 +19,7 @@ check() {
 install() {
     local unit=rhcos-afterburn-checkin.service
     inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
-    systemctl -q --root="$initdir" add-requires ignition-files.service "$unit"
+    # note we `|| exit 1` here so we error out if e.g. the units are missing
+    # see https://github.com/coreos/fedora-coreos-config/issues/799
+    systemctl -q --root="$initdir" add-requires ignition-files.service "$unit" || exit 1
 }
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
@@ -8,7 +8,9 @@ install_unit() {
     local unit=$1; shift
     local target=${1:-ignition-complete.target}
     inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
-    systemctl -q --root="$initdir" add-requires "$target" "$unit"
+    # note we `|| exit 1` here so we error out if e.g. the units are missing
+    # see https://github.com/coreos/fedora-coreos-config/issues/799
+    systemctl -q --root="$initdir" add-requires "$target" "$unit" || exit 1
 }
 
 depends() {
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
@@ -33,6 +33,8 @@ install() {
 
     # We don't support FIPS in diskless cases currently
     target=ignition-diskful.target
-    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips.service
-    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips-finish.service
+    # note we `|| exit 1` here so we error out if e.g. the units are missing
+    # see https://github.com/coreos/fedora-coreos-config/issues/799
+    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips.service || exit 1
+    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips-finish.service || exit 1
 }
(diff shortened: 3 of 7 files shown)


overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/module-setup.sh
```diff
@@ -19,5 +19,7 @@ check() {
 install() {
     local unit=rhcos-afterburn-checkin.service
     inst_simple "$moddir/$unit" "$systemdsystemunitdir/$unit"
-    systemctl -q --root="$initdir" add-requires ignition-files.service "$unit"
+    # note we `|| exit 1` here so we error out if e.g. the units are missing
+    # see https://github.com/coreos/fedora-coreos-config/issues/799
+    systemctl -q --root="$initdir" add-requires ignition-files.service "$unit" || exit 1
```
jlebon, line 24: I guess this dracut module should technically be re-ordered to 31 at least so it's more obvious that it's after `30ignition`. Though the fact that we have `|| exit 1` here clearly demonstrates that it works fine.

travier:
We realized a bit late that not mounting the ESP by default currently breaks bootupd. This might not be a blocker but raising. @cgwalters?

jlebon:
OK, rebumped this to include https://github.com/coreos/fedora-coreos-config/pull/786. Correspondingly as per https://github.com/coreos/fedora-coreos-config/pull/786#issuecomment-754923123, also added a new `overlay.d/` dir to keep shipping the rules in RHCOS until we have a `-udev` subpackage. @miabbott Are you able to test this?

miabbott:
> OK, rebumped this to include [coreos/fedora-coreos-config#786](https://github.com/coreos/fedora-coreos-config/pull/786). Correspondingly as per [coreos/fedora-coreos-config#786 (comment)](https://github.com/coreos/fedora-coreos-config/pull/786#issuecomment-754923123), also added a new `overlay.d/` dir to keep shipping the rules in RHCOS until we have a `-udev` subpackage. @miabbott Are you able to test this?

I'll try to get it tested in Azure proper today

miabbott:
@jlebon I believe we our own dracut module for the Azure udev rules here

```
$ git diff --staged
diff --git a/overlay.d/05rhcos/usr/lib/dracut/modules.d/25coreos-azure-udev/module-setup.sh b/overlay.d/05rhcos/usr/lib/dracut/modules.d/25coreos-azure-udev/module-setup.sh
new file mode 100644
index 0000000..e796e89
--- /dev/null
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/25coreos-azure-udev/module-setup.sh
@@ -0,0 +1,18 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# We want to provide Azure udev rules as part of the initrd, so that Ignition
+# is able to detect disks and act on them.
+#
+# If the WALinuxAgent-udev package is changed to install the udev rules as
+# part of the initramfs, we should drop this module.
+#
+# See https://bugzilla.redhat.com/show_bug.cgi?id=1909287
+# See also https://bugzilla.redhat.com/show_bug.cgi?id=1756173
+
+install() {
+    inst_multiple \
+        /usr/lib/udev/rules.d/66-azure-storage.rules \
+        /usr/lib/udev/rules.d/99-azure-product-uuid.rules
+}
```

jlebon:
> @jlebon I believe we our own dracut module for the Azure udev rules here

Hmm, we should be inheriting that from FCOS though, right?

miabbott:
> > @jlebon I believe we our own dracut module for the Azure udev rules here
> 
> Hmm, we should be inheriting that from FCOS though, right?

You are probably right...let me clean up my repo and make sure I've got it all straight.

miabbott:
I think I must have forgot to `git submodule update` before building/testing.  Looks like the rules are correctly installed in the initramfs; still need to test in Azure for reals.

cgwalters:
> We realized a bit late that not mounting the ESP by default currently breaks bootupd. T

bootupd 0.2.5 that fixes this is built in 4.7.

miabbott:
@jlebon OK, built this cleanly and confirmed that the udev rules are operating properly on Azure.  Thanks for patience.

cgwalters (approved):

jlebon:
Looks like this is ready for a `/lgtm` now? :)

Issue #448, opened 2020-11-19
s390x: Upgrade from old luks to new not working with DASD disks


Prashanth684:
Was testing out https://github.com/openshift/os/pull/436 on some Z systems by locally upgrading using ostree and it works fine on s390x systems with fiber channel disks, but on systems with DASD disks, it's problematic because of how DASD disks work. `sgdisk` doesn't work with DASD only `fdasd` works. There has to be some conversion of sectors to tracks as DASD disks represents data in tracks and not sectors, so there is some translation needed.But that's not even the hard part. fdasd's non interactive mode is severly limited and unlike sgdisk you cannot delete a specific partition or change labels, etc. So in effect, the whole partition layout has to be recreated. I'm wondering if we would have to `dd` the boot and root partitions and save them , use fdasd to recreate the partitions and then copy them over (omitting the luks header offset of course). I am not sure of any other good way to do this.Any ideas are appreciated. cc @nikita-dubrovskii

Prashanth684:
specifically this : https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container#L39 will not work .I also started experimenting with this patch to convert the sectors to tracks and recreate the partitions:
```
--- a/overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
+++ b/overlay.d/05rhcos/usr/libexec/rhcos-upgrade-root-filesystem-container
@@ -19,24 +19,49 @@ if [ -h "/dev/mapper/coreos-luks-root-nocrypt" ]; then
     startsector="$(($(cat /sys/block/$parent/$partition/start)+32768))"
     # the dm device is already the correct size
     size=$(cat /sys/block/$dm_device_name/size)
-
-    # check / handle 4k sector size devices
     sector_size=$(cat /sys/block/$parent/queue/logical_block_size)
-    if [ "$sector_size" == "4096" ]; then
-        startsector=$((startsector / 8))
-        size=$((size / 8))
-    fi
 
     # validate that the new partition will start at the xfs super block
-    skipcount=$(($startsector * $sector_size))
+    skipcount=$(($startsector * 512))
     if [ $(dd if=/dev/$parent bs=1 count=4 status=none skip=$skipcount) != XFSB ]; then
         echo "couldn't detect XFS superblock at expected partition start"
         exit
     fi
 
-    # recreate the partition offset after the LUKS header
[...]

PR #481, opened 2021-01-07
Revert "overlay.d/05rhcos: add rhcos-upgrade-root-filesystem-container"

arithx:
This reverts commit 66d1df13cea80569e19d5f7c73632f58e13a288f.

This upgrade script cannot live in the real root (we need to perform different actions on other architectures; see #448). Revert for 4.7 and implement an initrd based option in 4.8.


cgwalters (approved):
OK right...and we had kept `coreos-luks-open.service` because we needed it until *after* the reboot from the migration script we're deleting.

jlebon (approved):

arithx:
> OK right...and we had kept `coreos-luks-open.service` because we needed it until _after_ the reboot from the migration script we're deleting.

Yup. It's been stripped down to just the unlocking / dm-linear mapping bits (see 2fba76ec51001e6402d62bb3df13fb484fce5130 for the detailed view) but it is still present.

jlebon:
(Will shepherd this through together with https://github.com/openshift/os/pull/482.)

PR #482, opened 2021-01-07
tests/misc-ro: drop `ls /boot/efi` check

jlebon:
We don't mount `/boot/efi` anymore, so the `ls` invocation succeeds
because it's just targeting an empty dir.

Matches https://github.com/coreos/fedora-coreos-config/pull/794.

### DIFF
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -113,9 +113,6 @@ echo ok qemu guest agent
 
 case "$(arch)" in
     x86_64|aarch64)
-        if runuser -u core -- ls /boot/efi &>/dev/null; then
-            fatal "Was able to access /boot/efi as non-root"
-        fi
         # This is just a basic sanity check; at some point we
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, played with using symlinks to share the tests we want with FCOS, though seems like `kola run` doesn't pick it up. We should investigate that; might be easy way to make this less painful.

jlebon:
Looks like CI is flaking:

```
 error: build error: failed to pull image: After retrying 2 times, Pull image still failed due to error: error reading pull credentials: open : no such file or directory
2021/01/07 21:51:08 No custom metadata found and prow metadata already exists. Not updating the metadata.
2021/01/07 21:51:08 Ran for 25s
error: some steps failed:
  * could not run steps: step src failed: the build src failed after 23s with reason PullBuilderImageFailed: Failed pulling builder image. 
```

Just going to override it since I think this is safe, and Prow doesn't yet run kola tests anyway.

/override ci/prow/unit

PR #485, opened 2021-01-14
05rhcos: add /etc/mke2fs.conf to the initrd to fix mkfs.ext4

bgilbert:
RHEL e2fsprogs has a bug where mke2fs will fail if it falls back on the compiled-in mke2fs.conf.

https://bugzilla.redhat.com/show_bug.cgi?id=1889464
https://bugzilla.redhat.com/show_bug.cgi?id=1916382


cgwalters (approved):
/lgtm

PR #487, opened 2021-01-20
initrd: include 'sg_inq' binary and 'scsi-sg3_id.rules' 

nikita-dubrovskii:
Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1918244

On s390x systems with IBM 2810XIV discs multipath couldn't be configured because SCSI_IDENT_*
udev properties are not set at boot time:
```
sda: blacklisted, udev property missing
sdb: blacklisted, udev property missing
```

Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# Fix for https://bugzilla.redhat.com/show_bug.cgi?id=1918244
+# On s390x systems with IBM 2810XIV discs multipath couldn't be configured
+# because SCSI_IDENT_* udev properties are not set at boot time
+install() {
+    inst_simple sg_inq
+    inst_rules 61-scsi-sg3_id.rules
+}


nikita-dubrovskii:
This is a workaround, i'll continue discovering why SCSI_IDENT_LUN_EUI64 is not set

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
jlebon, line 1: Can we change this to:

```suggestion
#!/bin/bash
set -euo pipefail
```

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.sh
```diff
@@ -0,0 +1,21 @@
...
+    model=$(cat /sys/class/block/${dev}/device/model)
+    if [ "${model}" == "2810XIV" ]; then
+	enable=1
+	break
+    fi
+done
+
+if [ -n ${enable} ] && [ ! -e /etc/multipath.conf ] ; then
+    mkdir -p /etc/multipath/multipath.conf.d
+    mpathconf --enable
```
jlebon, line 18: This is doing what `multipathd-configure.service` does already, so couldn't we just order ourselves after that service only to run the `sed` command?

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.sh
```diff
@@ -0,0 +1,21 @@
...
+    sed -i "s/(SCSI_IDENT_|ID_WWN)/(SCSI_IDENT_|ID_WWN|ID_SERIAL)/g" /etc/multipath.conf
```
jlebon, line 19: From researching a bit about multipath configs, IIUC what this is doing is augmenting the blacklist exception to include any device which has an `ID_SERIAL` udev property, right? Whereas usually it would only be those with `SCSI_IDENT_...` and `ID_WWN...` properties. Looking at `multipath.conf(5)` from [here](https://github.com/opensvc/multipath-tools/blob/8dcfb54346c110610cccf923000171a10dbf656a/multipath/multipath.conf.5#L1330-L1332), it says:

> The default \fIblacklist exception\fR is: \fB(SCSI_IDENT_|ID_WWN)\fR, causing
well-behaved SCSI devices and devices that provide a WWN (World Wide Number)
to be included, and all others to be excluded.

Is there a concern here that `ID_SERIAL` is too generic and may enable multipath on devices that normally *should* be blacklisted?

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/module-setup.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+depends() {
+    echo systemd
+}
+
+install_and_enable_unit() {
+    unit="$1"; shift
```
jlebon, line 10: `local` and below

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.service
```diff
@@ -0,0 +1,17 @@
...
+After=systemd-udev-trigger.service systemd-udev-settle.service
+Before=local-fs-pre.target multipathd.service multipathd-configure.service
+DefaultDependencies=no
+Conflicts=shutdown.target
+
+ConditionKernelCommandLine=rd.multipath=default
+
+[Service]
+Type=oneshot
+ExecStart=/usr/sbin/ibm2810xiv-multipath-config
```
jlebon, line 14: For `oneshot` services, it's good to use `RemainAfterExit=yes`.

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.sh
jlebon: If this affects a specific model, couldn't we craft a more specific exception instead?

jlebon:
> i'll continue discovering why SCSI_IDENT_LUN_EUI64 is not set

:+1: Note BTW this probably also affects traditional RHEL, so a higher-level fix addressing both would be better.

overlay.d/05rhcos/usr/lib/dracut/modules.d/89ibm2810xiv-mutipath-config/ibm2810xiv-multipath-config.sh
nikita-dubrovskii: So i found the missing part - `scsi-sg3_id.rules`.

Prashanth684:
> > i'll continue discovering why SCSI_IDENT_LUN_EUI64 is not set
> 
>  Note BTW this probably also affects traditional RHEL, so a higher-level fix addressing both would be better.

maybe the BZ needs to modified and moved to RHEL instead ?

overlay.d/05rhcos/usr/lib/dracut/modules.d/01scsi-rules/module-setup.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
```
jlebon, line 3: Can you add a comment here on why we need this as well as a link to the RHBZ for posterity?

jlebon (approved):
Nice work debugging this! :)
One comment, otherwise LGTM.

jlebon (approved):
/lgtm

PR #488, opened 2021-01-20
building: Update instructions to use build02

cgwalters:
We have nested virt finally enabled!


miabbott:
Unfortunately, DPTP disabled creation (today!) of personal projects on the build cluster, so we can't vet these instructions as written.  There's an aos-devel thread discussing possible path forward.

cgwalters:
Rebasing this "pure docs" PR as a CI canary check.

PR #489, opened 2021-01-20
kola-denylist: skip crio test on x86_64

miabbott:
We are seeing crio failures that appear to be network related which
are blocking RHCOS 4.7 builds.  It doesn't appear to be an explicit
crio issue, but needs further investigation.

See https://bugzilla.redhat.com/show_bug.cgi?id=1917470

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -5,11 +5,15 @@
   tracker: https://bugzilla.redhat.com/show_bug.cgi?id=1782026
   arches:
    - s390x
+# making two entries for crio.base because they are separate issues
 - pattern: crio.base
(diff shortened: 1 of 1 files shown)


ashcrow (approved):

miabbott:
@jlebon split up the items as requested in standup; works as expected locally

Note: lists of tracker issues do work:

```
⚠️  Skipping kola test pattern "crio.base":                                                                                                                                                                                                                                                                                    
⚠️  ['https://github.com/kubernetes/kubernetes/issues/87325', 'https://bugzilla.redhat.com/show_bug.cgi?id=1917470'] 
```

PR #493, opened 2021-02-01
[release-4.7] Target fedora-coreos-config branch rhcos-4.7 and bump

jlebon:
```
Jonathan Lebon (1):
      Revert "overlay/boot-mount-generator: Mount /boot read-only,nodev,nosuid"
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fedora-coreos-config"]
 	path = fedora-coreos-config
 	url = https://github.com/coreos/fedora-coreos-config
-	branch = testing-devel
+	branch = rhcos-4.7


travier (approved):

travier:
Looks like the CI did not check this one (not an issue here but something to keep in mind for later).

travier:
Ping @cgwalters as this might need to be fixed from an another repo but I don't know which one.

miabbott:
@travier I believe the job is defined here - https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/os/openshift-os-master-presubmits.yaml

Looking at the MCO repo as an example, I think we need another yaml for the 4.7 branch

e.g. https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/machine-config-operator/openshift-machine-config-operator-release-4.7-presubmits.yaml

PR #494, opened 2021-02-01
chrony NM dispatcher: Account for dhclient dispatcher script in /usr/lib/

LorbusChris:
In current dhcp-client Fedora RPMs,
the dhclient dispatcher script is placed at
`/usr/lib/NetworkManager/dispatcher.d/11-dhclient`
as opposed to
`/etc/NetworkManager/dispatcher.d/11-dhclient`
used before.

This changes the chrony dispatcher script to account for the
dhclient dispatcher script in either place.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -299,4 +299,5 @@ remove-from-packages:
   - - dhcp-client
     - "/etc/dhcp/dhclient.conf"
     - "/etc/NetworkManager/dispatcher.d/11-dhclient"
+    - "/usr/lib/NetworkManager/dispatcher.d/11-dhclient"
     - "/usr/lib64/pm-utils/sleep.d/56dhclient"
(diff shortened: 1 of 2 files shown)


LorbusChris:
@cgwalters does this work even if the newly excluded file here is not present in the RPM yet?

jlebon:
I'm confused by this. The commit message indicates that it's to account for Fedora packaging differences, but we're strictly concerned with RHCOS here, no? For FCOS IIRC, we rely on that functionality having migrated to chrony.

LorbusChris:
The PR is a reaction to this comment: https://src.fedoraproject.org/rpms/chrony/pull-request/5#comment-66021

This change isn't strictly necessary here, but I thought it might be confusing to have that many versions of this script living in various places..

jlebon:
Ahh gotcha, thanks. Maybe we can add a link to this in the commit message? Something like:

> This isn't strictly needed for RHCOS, but we want to keep it in sync with the chrony script: https://src.fedoraproject.org/rpms/chrony/pull-request/5

But LGTM as is too.
/approve

LorbusChris:
@jlebon updated :slightly_smiling_face:

PR #496, opened 2021-02-04
extensions: adapt for new rpm-ostree support

jlebon:
With https://github.com/coreos/rpm-ostree/pull/2439 and
https://github.com/coreos/coreos-assembler/pull/2028, support for
extensions is now more robust.

The extensions.yaml file is now fed directly to rpm-ostree. We just need
to do some minor tweaks:
- drop listing usbguard deps, rpm-ostree fetches them automatically
- split the devel bits into two extensions: one is an OS extension which
  contains the devel and headers packages, and the other is a
  development extension containing kernel packages already in the base
  (marking it as a development extension will signal the MCO that it's
  not intended to be directly layered on the host)
- use `match-base-evr: kernel` on those to enforce that the kernel
  development packages have the same EVR as the base kernel package in
  the OSTree

### DIFF
--- a/oscontainer.yaml
+++ b/oscontainer.yaml
@@ -1,2 +1,9 @@
 # Configuration for oscontainer uploads; this is the base image.
 base: registry.access.redhat.com/ubi8/ubi:latest
+# These are the packages whose NEVRAs are included as labels on the image.
+labeled-packages:
+  - kernel kernel-rt-core
+  - ostree rpm-ostree
+  - ignition
+  - systemd
+  - runc cri-o
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -1,26 +1,35 @@
 # RPMs as operating system extensions, distinct from the base ostree commit/image
 # https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md
 # and https://github.com/coreos/fedora-coreos-tracker/issues/401
-# We currently explicitly list dependencies for each extension; see
-# https://github.com/coreos/rpm-ostree/issues/2055
+
+repos:
+  - rhel-8-nfv
 
 extensions:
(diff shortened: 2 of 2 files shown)


jlebon:
Needs https://github.com/coreos/coreos-assembler/pull/2028.

miabbott:
Let's just hold this until the `cosa` PR merges

/hold

jlebon:
cosa PR is merged now!
/hold cancel

/retest

oscontainer.yaml
```diff
@@ -1,2 +1,9 @@
 # Configuration for oscontainer uploads; this is the base image.
 base: registry.access.redhat.com/ubi8/ubi:latest
+# These are the packages whose NEVRAs are included as labels on the image.
+labeled-packages:
+  - kernel kernel-rt-core
+  - ostree rpm-ostree
+  - ignition
+  - systemd
+  - runc
```
miabbott, line 9: add `cri-o`?

miabbott:
One suggestion, but otherwise LGTM!

/approve

oscontainer.yaml
jlebon: Good idea, done!

miabbott:
`fatal: Missing /dev/kvm`

I think @cgwalters was chasing this down as part of 

https://github.com/openshift/release/pull/16197
https://github.com/openshift/release/pull/16193

jlebon:
So... since CI is broken but doesn't cover this anyway, and this is blocking pipelines, let's just get this in? Then we get to actual regressions in this new path sooner and can fix them sooner too. :)

cgwalters:
Yeah, that's fine by me.
/override ci/prow/unit

jlebon:
Now just needs a `/lgtm`!

miabbott:
/lgtm

For xref purposes, CI failures tracking in #506

PR #502, opened 2021-02-19
manifest: Add conntrack (tools) but without the daemon

cgwalters:
https://github.com/coreos/fedora-coreos-tracker/issues/404
https://bugzilla.redhat.com/show_bug.cgi?id=1925698
https://github.com/openshift/machine-config-operator/pull/2421

This will help us work around a believed kernel bug for OpenShift right now.  We may remove this later.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -270,6 +270,10 @@ packages:
  # https://bugzilla.redhat.com/show_bug.cgi?id=1877905
  # https://bugzilla.redhat.com/show_bug.cgi?id=1886201
  - perl-interpreter
+ # https://github.com/coreos/fedora-coreos-tracker/issues/404
+ # https://bugzilla.redhat.com/show_bug.cgi?id=1925698
+ # https://github.com/openshift/machine-config-operator/pull/2421
+ - conntrack-tools
 
 packages-x86_64:
   # Temporary add of open-vm-tools. Should be removed when containerized
@@ -301,3 +305,7 @@ remove-from-packages:
     - "/etc/NetworkManager/dispatcher.d/11-dhclient"
     - "/usr/lib/NetworkManager/dispatcher.d/11-dhclient"
     - "/usr/lib64/pm-utils/sleep.d/56dhclient"
+  # Remove the systemd unit; we only want the binary to be used
+  # by MCD or kubelet.  See above.
+  - - conntrack-tools
+    - /usr/lib/systemd/system
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -130,6 +130,9 @@ if ! test -f /etc/iscsi/initiatorname.iscsi; then
 fi
 echo "ok iSCSI initiator name"
 
+rpm -q conntrack-tools
+test ! -f /usr/lib/systemd/system/conntrackd.service
+echo "ok conntrack tools without daemon"
 
 # Let's make sure the NetworkManager we use is one of the one-off
 # rebuilds while we're following RHEL 8.3.


travier (approved):

travier:
Will need a 4.7 backport

LorbusChris:
I wonder whether this should be done in upstream FCOS for now. Things like this would otherwise have to be managed in https://github.com/openshift/okd-machine-os/, broadening the gap between FCOS and the OKD machine OS

cgwalters:
I agree this is creating an OKD gap; we could address that by having okd-machine-os just install the conntrack-tools package alongside the rest of the stuff like kubelet etc.

I am less certain about blocking this quick-fix-for-OCP on adding this to FCOS - that's basically a permanent commitment, although sentiment seemed to be in favor-ish.

jlebon:
As a quick hack, this seems fine to me!

/approve

But long-term we should either get that package split out, or e.g. moved to the MCD as was discussed. Given that, I'd rather not do this hack at all in FCOS if we can, because `remove-from-packages` is really not great (and honestly, we should be looking at dropping the ones we currently have).

jlebon:
> Things like this would otherwise have to be managed in openshift/okd-machine-os, broadening the gap between FCOS and the OKD machine OS

Because FCOS is more general that RHCOS, I think some gap will basically always exist and we should embrace that and figure out how to manage it best (e.g. like the [extensions strengthening work](https://github.com/coreos/coreos-assembler/pull/2028)),

lucab:
This is going to change cri-o behavior, which performs some [networking-related logic](https://github.com/cri-o/cri-o/blob/v1.20.0/internal/hostport/hostport_manager.go#L190) based on auto-detection of binary presence.
Ideally, before landing this, there should be a tri-state knob (`on | off | autodetect`) in cri-o configuration so that the expected behavior can be pinned.

cgwalters:
> This is going to change cri-o behavior,

Right but...is that code always something we want anyways?  I am not sure.

One option to make this even more obviously a hack for current OCP is to move the binary to e.g. `/usr/opt/openshift-private/bin/conntrack`, and that would also defeat crio finding it too - if indeed that is a problem.

mrunalp:
@aojea ptal from the crio hostport manager perspective.

aojea:
> This is going to change cri-o behavior, which performs some [networking-related logic](https://github.com/cri-o/cri-o/blob/v1.20.0/internal/hostport/hostport_manager.go#L190) based on auto-detection of binary presence.
> Ideally, before landing this, there should be a tri-state knob (`on | off | autodetect`) in cri-o configuration so that the expected behavior can be pinned.

cri-o manage the hostport in the containers, without the conntrack binary crio has a bug in certain UDP scenarios and it also doesn't pass one e2e test https://github.com/kubernetes/kubernetes/issues/91216.

I don't know the reasons why this is this way honestly, but the correct behaviour for crio, since is the owner of the hostport logic, is to use the conntrack binary to delete the stale entries ... the conntrack logic doesn't modify any behaviour, it fixes a bug

/lgtm

mrunalp:
@aojea Thanks! Now the question is if the crio code is enough for the gcp issue or we still need the changes proposed in https://github.com/openshift/machine-config-operator/pull/2421 cc: @michaelgugino

cgwalters:
/hold
CI failure looks unrelated, not sure what's going on.  Will ping DPTP Monday.

ashcrow (approved):

cgwalters:
Ohh I see the problem:

> "0/26 nodes are available: 18 Insufficient devices.kubevirt.io/kvm, 2 node(s) had taint {ci.openshift.io/ci-search: true}, that the pod didn't tolerate, 3 node(s) had taint {node-role.kubernetes.io/infra: }, that the pod didn't tolerate, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate."

cgwalters:
We debugged this to https://github.com/openshift/release/commit/7a961fa5da8eee4aecd9e2bfcfcf16be53858e1c which steered jobs for this repo back to build01, which means we lost nested virt.  DPTP is investigating a fix.

cgwalters:
https://github.com/openshift/release/pull/16193 merged, let's see
/retest

cgwalters:
OK now we ran on build02, but are missing `/dev/kvm`.

cgwalters:
Now retrying after https://github.com/openshift/release/pull/16197

cgwalters:
I added some more debugging
/retest

miabbott:
So nice to have working CI again!  Thank you so much @cgwalters!

/hold cancel

PR #508, opened 2021-02-26
remove gcp-routes hack from os

aojea:
The logic is handled by the machine-config-operator

Signed-off-by: Antonio Ojea <aojea@redhat.com>


aojea:
@cgwalters are these scripts supposed to be here?
How can I test an image with this PR, clusterbot is failing to me if I ask him to use thisPR

aojea:
@squeed should this be here? or only in the MCO?

LorbusChris:
AFAIK this is still required here, since it has to be present on the bootstrap node where the MCO never runs.

Long term solution we aim for is to replace the gcp-routes entirely with nm-cloud-setup (see https://github.com/coreos/fedora-coreos-tracker/issues/320)

aojea:
> AFAIK the gcp routes script is still required here, since it has to be present on the bootstrap node where the MCO never runs.
> 
> Long term solution we aim for is to replace the gcp-routes entirely with nm-cloud-setup (see [coreos/fedora-coreos-tracker#320](https://github.com/coreos/fedora-coreos-tracker/issues/320))

then, how can be replaced from the non-bootsrap nodes, we'll have a pod doing the job https://github.com/openshift/machine-config-operator/pull/2422 and this will interfere

michaelgugino:
The bootstrap node was the whole reason we added the iptables rules drop in the first place.  Seems like we could make this run on only the bootstrap node, potentially?

darkmuggle:
/lgtm
In light of https://github.com/openshift/machine-config-operator/blob/master/templates/master/00-master/gcp/files/opt-libexec-openshift-gcp-routes-sh.yaml we don't need the script

yuqi-zhang:
The MCO script only applies for masters. Would it need to be applied for workers as well?

Bootstrap generation is handled via installer so I think this may need to be split across multiple locations

aojea:
> Bootstrap generation is handled via installer so I think this may need to be split across multiple locations

/hold
I don't have the overall view of how many Openshift GCP loadbalancers are needed, so if there are more components than the apiserver on the nodes that need a GCP LB that relies on this script, we can break it
@squeed do you have more context about this?

LorbusChris:
From what I've gathered the VIP handling on the bootstrap node doesn't have to be as intricate as the gcp-routes services does it.
I think the core team's plan was to replace the script on the bootstrap node with `nm-cloud-setup`, and then let gcp-routes take over on the masters once the MCO runs.
/cc @sohankunkerkar @lucab

PR #509, opened 2021-03-04
ci: build extensions

miabbott:
Now that we have a better story for build extensions along side the
OS, let's start doing that as part of our CI.


miabbott:
Ideally we have some tests related to consuming the extensions, too

jlebon:
Nice, I like it!
Looks like CI is still down here.
/retest

Issue #506, opened 2021-02-24
CI is broken tracker


cgwalters:
Splitting this from https://github.com/openshift/os/pull/502 

So far what appears to have happened is a config change broke the steering of jobs to build02 where we have nested virt enabled.

But now, AFAICS our jobs are running on build02 again, and:

kvm-device-plugin is healthy:
```
$ oc get ds
NAME                DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
kvm-device-plugin   42        42        42      42           42          <none>          39d
$ 
```

Looking at the pods from one CI run:

```
$ oc project ci-op-d7w80vyk
Now using project "ci-op-d7w80vyk" on server "https://api.build02.gcp.ci.openshift.org:6443".
$ oc get pods
NAME        READY   STATUS      RESTARTS   AGE
bin-build   0/1     Error       0          33m
src-build   0/1     Completed   0          35m
$
```

That's on `node/build0-gstfj-w-c-gttfw.c.openshift-ci-build-farm.internal`.

Looking at that node, I see:

```
Capacity:
  attachable-volumes-gce-pd:  127
  cpu:                        16
  devices.kubevirt.io/kvm:    2
  ephemeral-storage:          1257739244Ki
  hugepages-1Gi:              0
  hugepages-2Mi:              0
  memory:                     61832252Ki
  pods:                       250
Allocatable:
  attachable-volumes-gce-pd:  127
  cpu:                        15
  devices.kubevirt.io/kvm:    2
  ephemeral-storage:          1158058743528
  hugepages-1Gi:              0
  hugepages-2Mi:              0
  memory:                     58584124Ki
  pods:                       250
```

So we have `devices.kubevirt.io/kvm` on that node.

And that node is backed by `machine/build0-gstfj-w-c-gttfw` which was booted with `Image:        projects/rhcos-cloud/global/images/rhcos-46-82-202011260640-0-gcp-x86-64` which looks right.

cgwalters:
I came across this helpful doc https://cloud.google.com/compute/docs/troubleshooting/troubleshooting-nested-vms

And got the output of that, seems we have `/dev/kvm` just fine and have `vmx`:

```
$ oc --context build02 --as system:admin debug node/build0-gstfj-w-c-gttfw.c.openshift-ci-build-farm.internal -- grep -c vmx /proc/cpuinfo
...
16
$
[...]

PR #513, opened 2021-03-10
ci: Rework to split build/test

cgwalters:
Aiming to fix https://github.com/openshift/os/issues/506

I think the problem here may be related to the fact that
we're trying to do all of our CI as part of a build.

Instead, let's function like e.g. the CI for coreos/rpm-ostree
which creates a derived container from the source code.  For
us, this derived container is basically the 2-tuple (source, cosa).

For coreos/rpm-ostree we actually do generate binaries as well
as test code from the buildroot, i.e. a 3 tuple (binaries, test binaries, cosa).
But here we don't need that, there's no actual "build" step.

Split off `validate.sh` as a separate thing that can be a
clearly separate step.


cgwalters:
Should pair with https://github.com/openshift/release/pull/16649

cgwalters:
Note that the reason this PR passed is that I neutered the CI back to just running `validate.sh`.  After the this merges I'll try ratcheting back in the real CI too.

miabbott:
:crossed_fingers: 

/lgtm

PR #525, opened 2021-03-25
Add sandboxed-containers extension

fidencio:
This reworks af93514887350da53e035d1b841eea36542cdd42 as, since it was
merged, a few changes happened in the enhancement proposal and it's been
decided to go with a "sandboxed-containers" extensions instead.

As the "qemu-kiwi" extension is no longer needed, we're removing it
(well, reworking it) and adding the "sandboxed-containers" one instead.

The sandboxed-containers extension is added to the MCO with by the
following PR:
https://github.com/openshift/machine-config-operator/pull/2456

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -43,10 +43,13 @@ extensions:
       - kernel-rt-modules
       - kernel-rt-modules-extra
       - kernel-rt-devel
-  # https://github.com/openshift/machine-config-operator/pull/2376/
+  # https://github.com/openshift/machine-config-operator/pull/2456
+  # https://github.com/openshift/enhancements/blob/master/enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
   # GRPA-3123
-  qemu-kiwi:
+  # - kata-containers (RHAOS)
+  #   - qemu-kiwi (advanced-virt)
+  sandboxed-containers:
     architectures:
       - x86_64
     packages:
-      - qemu-kiwi
+      - kata-containers


fidencio:
On a note, I'm note sure if I need to list **all** the dependencies as I did, but better be safe than sorry here.

extensions.yaml
```diff
@@ -43,10 +44,27 @@ extensions:
...
       - x86_64
     packages:
+      - kata-containers
       - qemu-kiwi
+      - ipxe-roms-qemu
+      - libpmem
+      - pixman
+      - seabios-bin
+      - seavgabios-bin
+      - sgabios-bin
```
miabbott, line 70: As long as `qemu-kiwi` has its dependencies properly enumerated in the RPM, `rpm-ostree` will do the necessary depsolving to pull them in.

You could probably even reduce the extension to just specify `kata-containers` and it should pull in `qemu-kiwi` and then all the deps

extensions.yaml
```diff
@@ -4,6 +4,7 @@
 
 repos:
   - rhel-8-advanced-virt
+  - rhel-8-appstream
```
miabbott, line 7: You can drop this repo from this manifest; `rpm-ostree` will use the repos in the `manifest.yaml` config for finding RPMs.  This list is better thought of as "I need these additional repos for extensions"

extensions.yaml
fidencio: Perfect!

extensions.yaml
fidencio: Perfect.  Do I also have to add `http://rhsm-pulp.corp.redhat.com/content/dist/layered/rhel8/x86_64/rhocp/4.8/` in the same way that it was done for the advanced-virt one?

Asking because the `kata-containers` RPM will show up there (not yet released, thus not yet there).

miabbott: I think `rhocp` is the same as the RHOSE one?

https://github.com/openshift/release/blob/db8e82f518f55d46d80493744d074e5d657b2259/core-services/release-controller/_repos/ocp-4.8-rhel8.repo#L25-L28

miabbott: Maybe not; probably worth asking the ART or release folks

fidencio: Yep, not sure yet. :-)
But AFAIR, it'll be released in a "tech-preview" kind of repo / CDN, so, if needed, I'll add the needed bits to `redhat-cores` as already done for the advanced-virt repo, and propagate it here accodingly.

@miabbott, thanks a lot for the pointers and the review, very much appreciated!

fidencio:
This one depends on https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1242

miabbott:
> This one depends on https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1242

This PR can actually merge without that change; but to make CI pass, you'll need to update the `openshift/release` repo with a PR for the `rhocp` repo similar to what was done here - https://github.com/openshift/release/pull/16258

fidencio:
> > This one depends on https://gitlab.cee.redhat.com/coreos/redhat-coreos/-/merge_requests/1242
> 
> This PR can actually merge without that change; but to make CI pass, you'll need to update the `openshift/release` repo with a PR for the `rhocp` repo similar to what was done here - [openshift/release#16258](https://github.com/openshift/release/pull/16258)

Aha! On that, thanks for the pointer!

extensions.yaml
```diff
@@ -43,10 +44,21 @@ extensions:
...
-  qemu-kiwi:
+  # The dependency tree here is:
+  # - kata-containers (rhocp)
+  #   - qemu-kiwi (advanced-virt)
+  #     - ipxe-roms-qemu (appstream)
+  #     - libpmem (appstream)
+  #     - pixman (appstream)
+  #     - qemu-kvm-common (advanced-virt)
+  #     - seabios-bin (advanced-virt)
+  #     - seavgabios-bin (advanced-virt)
```
cgwalters, line 58: Fine as is, I'm just not sure that anyone will really care about the bios level packages.  Calling out `kata-containers -> qemu-kiwi` though makes total sense to me.

cgwalters:
```
rhel-8-rhocp': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Curl error (60): SSL peer certificate or SSH remote key was not OK for https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/rhocp/4.8/os/repodata/repomd.xml [SSL certificate problem: self signed certificate in certificate chain]
```

Odd that it's just failing for that repo but not the others.

cgwalters:
https://github.com/openshift/release/pull/17198#issuecomment-808468511

miabbott:
> ```
> rhel-8-rhocp': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Curl error (60): SSL peer certificate or SSH remote key was not OK for https://cdn.redhat.com/content/dist/layered/rhel8/x86_64/rhocp/4.8/os/repodata/repomd.xml [SSL certificate problem: self signed certificate in certificate chain]
> ```
> 
> Odd that it's just failing for that repo but not the others.

Since that repo was just added to the release-controller today, I'm going to guess it missed the nightly repo sync that happens.

miabbott:
> [openshift/release#17198 (comment)](https://github.com/openshift/release/pull/17198#issuecomment-808468511)

Fabiano fixed the baseurl in https://github.com/openshift/release/pull/17202/

extensions.yaml
```diff
@@ -5,6 +5,7 @@
 repos:
   - rhel-8-advanced-virt
   - rhel-8-nfv
+  - rhel-8-rhocp
```
cgwalters, line 8: IOW it should work to just drop this line.

fidencio: This is something I'm not exactly sure @cgwalters.  Apart from myself breaking the CI with my change in the `openshift/release` (sorry), the whole situation of where or when the `kata-containers` package will show up is still not clear to me.

I've reverted the changed on `openshift/release` as it was impacting other bits of the project, and I'll try to sync up with all the folks possibly involved into the release process.  For now, let me hold this one.

fidencio: Situation is finally solved, this line has been dropped and seems that we're good to go (huge thanks to the AOS-ART team!)

extensions.yaml
fidencio: Fixed!

fidencio:
@miabbott, @cgwalters, CI has finally passed on this one (yay!).  I think we can have this one merged sooner than later and it'll unblock the MCO PR.

PR #526, opened 2021-04-06
Remove now unneeded sync-sources.sh script

travier:


cgwalters (approved):

ashcrow (approved):

miabbott:
`+ fatal 'NetworkManager package changed from rhaos4.7 branch. Needs investigation.' ` looks like RHEL released a new version of `NetworkManager` that we need to get bumped for RHCOS.

Nice to have another CI system while the internal build cluster is on fire.

travier:
We can probably override this one as it is unlikely to need testing :)

cgwalters:
(Agree but it'd still be good to get CI fixed, although we can also just sprint to 8.4)

PR #527, opened 2021-04-06
Update for RHEL 8.4 Beta

travier:
Collection of small changes to move to RHEL 8.4.

PR #528, opened 2021-04-08
overlay: Add 15rhcos-logrotate (BZ#1780079)

cgwalters:
We landed the package addition in FCOS, but the package
there ships with a timer unit, but not in RHEL8.  That
seems like it's not going to happen, so backport the timer
unit manually.

For https://bugzilla.redhat.com/show_bug.cgi?id=1780079


miabbott:
Probably need to land #527 first to make CI green, but looks good otherwise.

/lgtm

cgwalters:
/hold
until ci is good

travier:
Let's get this one in before we branch?

PR #510, opened 2021-03-05
tests/kola: use root reprovisioning tests from FCOS

jlebon:
These all work on RHCOS right now.

It overlaps with the `luks/tpm` test, so drop that. Rename the
`luks/ignition` test to `luks-var` to be clearer.

PR #529, opened 2021-04-09
tests/kola: use var-mount test from FCOS

jlebon:
Now that symlinks are supported, let's do the same thing as #510 for
the var-mount test and dedupe.

Prep for backporting
https://github.com/coreos/fedora-coreos-config/pull/942.


jlebon:
Blocked on https://github.com/openshift/os/pull/527

jlebon:
Let's merge this? We need it for coverage from the functionality added in https://github.com/openshift/os/pull/530.

PR #532, opened 2021-04-13
[release-4.7] Bump fedora-coreos-config for clevis-luks-askpass.path enablement

jlebon:
Specifically https://github.com/coreos/fedora-coreos-config/pull/942 for
https://bugzilla.redhat.com/show_bug.cgi?id=1947490.

```
Jonathan Lebon (3):
      tests/kola/var-mount: convert to Butane config
      tests/kola/var-mount: test mounting on reboot
      presets: enable clevis-luks-askpass.path
```


jlebon:
Note: this bug is easy to work around and I don't think it warrants a full bootimage bump at this point. But staging it meanwhile means that we have CI coverage for it and that it'll make it to the next warranted bootimage bump.

PR #512, opened 2021-03-09
manifest: Use 'system-configuration' manifest

travier:
Packages have been split out from fedora-coreos-base to improve sharing with RHCOS.

PR #533, opened 2021-04-14
Add dhcp-client to manifest for dhclient binary

miabbott:
During the switch to using RHEL 8.4 Beta, the `dhcp-client` package dropped out of the content set.  (Was not able to quickly determine why)

This re-adds the package to the manifest, as we have OCP customers that are still relying on the `dhclient` binary for their network setup.  See:  https://bugzilla.redhat.com/show_bug.cgi?id=1908462

This also adds a test to ensure that the `dhclient` binary does not fall out of the compose.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -206,6 +206,9 @@ packages:
  - openvswitch2.13
  - dnsmasq
  - NetworkManager-ovs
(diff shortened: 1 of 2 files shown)


miabbott:
/hold

My `git` skills failed me this morning...looks like I undid some of #512

PR #535, opened 2021-04-15
manifest.yaml: Do not install rdma-core on s390x

Prashanth684:
rdma-core is needed for the infiniband subsystem which is not supported for s390x. This is causing a couple of systemd units to fail
when booting a system with RoCE cards (https://bugzilla.redhat.com/show_bug.cgi?id=1782876). Disabling the rdma-core as it is not needed for mellanox RoCE cards.


cgwalters (approved):
/lgtm

PR #538, opened 2021-04-16
Bug 1945646: start gcp-routes.sh script directly

miabbott:
Starting the script via bash puts the process in the initrc_t domain,
which is more restrictive than starting it directly and putting it
into unconfined_service_t. Starting the script direcrtly should quiet
some of the AVC messages that are being sent to the audit.log

See: https://bugzilla.redhat.com/show_bug.cgi?id=1945646#c11

### DIFF
--- a/overlay.d/05rhcos/usr/lib/systemd/system/gcp-routes.service
+++ b/overlay.d/05rhcos/usr/lib/systemd/system/gcp-routes.service
@@ -6,8 +6,8 @@ After=network.target
 
 [Service]
 Type=simple
-ExecStart=/bin/bash /usr/sbin/gcp-routes.sh start
-ExecStopPost=/bin/bash /usr/sbin/gcp-routes.sh cleanup
+ExecStart=/usr/sbin/gcp-routes.sh start
+ExecStopPost=/usr/sbin/gcp-routes.sh cleanup
 User=root
 RestartSec=30
 Restart=always


miabbott:
Built this locally and tested that the service starts normally in GCP; see https://bugzilla.redhat.com/show_bug.cgi?id=1945646#c13

overlay.d/05rhcos/usr/lib/systemd/system/gcp-routes.service
```diff
@@ -6,8 +6,8 @@ After=network.target
 
 [Service]
 Type=simple
-ExecStart=/bin/bash /usr/sbin/gcp-routes.sh start
-ExecStopPost=/bin/bash /usr/sbin/gcp-routes.sh cleanup
+ExecStart=/usr/sbin/gcp-routes.sh start
+ExecStopPost=/usr/sbin/gcp-routes.sh cleanup
```
jlebon, line 10: Kinda odd to have something in `/usr/sbin` use `.sh`. If this isn't something intended to be used directly by users, it should be in `/usr/libexec` instead. Anyway, not new so definitely not a blocker!

jlebon (approved):
It doesn't really make sense to me that those are [treated differently](https://bugzilla.redhat.com/show_bug.cgi?id=1945646#c11) by SELinux. Reminds me of [this hack](https://github.com/coreos/coreos-assembler/pull/2094) which actually does the opposite move to also work around SELinux issues. Can't live with it, can't live without it!

/lgtm

PR #540, opened 2021-04-16
Add presets for CLHM's `issuegen`-related units

kelvinfan001:
CLHM v0.21+ used in FCOS no longer has `{issue,motd}gen`-related units,
but RHCOS is still on older versions of CLHM that require these units.
These presets should be removed once RHCOS catches up to CLHM v0.21+.


kelvinfan001:
Depends on https://github.com/coreos/fedora-coreos-config/pull/967.

cgwalters (commented):
Thanks, I have no issues with this.  Or, without it.

kelvinfan001:
One of @cgwalters' puns needs to be nominated for "Best (worst) comment on a pull or merge request" for the "Emerging OS Awards 2021".

PR #544, opened 2021-04-29
Revert "kola-denylist: skip crio test on x86_64"

travier:
This reverts commit 62c25b7a664bd4239630d484088df0aece34fb3b.

See https://bugzilla.redhat.com/show_bug.cgi?id=1917470


travier:
I can not reproduce locally. Let's see if it's a quick failure here. I'll revert that in other pipelines for wider flake testing.

cgwalters (approved):
/approve
/lgtm

PR #545, opened 2021-04-29
manifest: Exclude unneeded udisks2

travier:
With RHEL 8.4, fwupd got pulled in as provider of dbxtool and that also
included udisks2 as a recommends. But udisks2 is apparently only needed
for encrypted swap checks in fwupd and is not even included in FCOS, so
manually exclude it.

For BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1953105

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -185,6 +185,8 @@ exclude-packages:
   # https://github.com/coreos/rpm-ostree/pull/1789/files/a0cd999a8acd5b40ec1024a794a642916fbc8ff8#diff-fc2076dc46933204a7a798f544ce3734
   # People need to use `rpm-ostree kargs` instead.
   - grubby
+  # udisks2 is a fwupd recommends only need for encrypted swap checks
+  - udisks2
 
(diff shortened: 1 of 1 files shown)


travier:
@hughsie Could you confirm that udisks2 is not strictly required for firmware updates in fwupd? Thanks!

hughsie:
> udisks2 is not strictly required for firmware updates in fwupd

We indeed use it to find the ESP partition, but admins can hardcode this in /etc/fwupd/uefi_capsule.conf instead if required. If you're not using any plugins that use the ESP then you don't need udisks. I guess that depends on if you're ever likely to see a capsule update in reality on openshift.

cgwalters (approved):

cgwalters:
> We indeed use it to find the ESP partition, but admins can hardcode this in /etc/fwupd/uefi_capsule.conf instead if required. 

Regarding the ESP: https://github.com/coreos/fedora-coreos-config/pull/794

Ultimately we may need a shared userspace component for dealing with the ESP, which may be bootupd.  Not sure.

PR #546, opened 2021-05-03
overlay.d: s390x: add to ramdisk missing zipl

nikita-dubrovskii:
This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1955621
When starting OCP in fips mode it immediate fails during install:
```
systemd[1]: Starting Check for FIPS mode...
rhcos-fips[763]: Found /etc/ignition-machine-config-encapsulated.json in Ignition config
rhcos-fips[763]: FIPS mode required; updating BLS entries
rhcos-fips[763]: Appending 'fips=1 boot=LABEL=boot' to /run/rhcos-fips/sysroot/boot/loader/entries/ostree-1-rhcos.conf
rhcos-fips[763]: /usr/sbin/rhcos-fips: line 78: zipl: command not found
```

Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
@@ -17,6 +17,12 @@ install() {
         bwrap \
         env
 
+    local _arch=${DRACUT_ARCH:-$(uname -m)}
+    if [[ "$_arch" == "s390x" ]]; then
+        inst_multiple zipl
+        inst /lib/s390-tools/stage3.bin
+    fi
+
     inst_script "$moddir/rhcos-fips.sh" \
         "/usr/sbin/rhcos-fips"
     inst_script "$moddir/coreos-dummy-ignition-files-run.sh" \
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -76,7 +76,9 @@ firstboot() {
               echo $(grep $line $f) >> $tmpfile
           done
       done
-      grep options $tmpfile | cut -d ' ' -f2- > $optfile
+      echo "Appending 'ignition.firstboot' to ${optfile}"
+      options="$(grep options $tmpfile | cut -d ' ' -f2-) ignition.firstboot"
+      echo $options > "$optfile"
       zipl --verbose \
            --target "${tmpsysroot}/boot" \
            --image $tmpsysroot/boot/"$(grep linux $tmpfile | cut -d' ' -f2)" \
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -70,16 +70,18 @@ firstboot() {
       # We need to call zipl with the kernel image and ramdisk as running it without these options would require a zipl.conf and chroot
       # into rootfs
       tmpfile=$(mktemp)
+      optfile=$(mktemp)
       for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
           for line in title version linux initrd options; do
               echo $(grep $line $f) >> $tmpfile
           done
       done
+      grep options $tmpfile | cut -d ' ' -f2- > $optfile
       zipl --verbose \
            --target "${tmpsysroot}/boot" \
            --image $tmpsysroot/boot/"$(grep linux $tmpfile | cut -d' ' -f2)" \
            --ramdisk $tmpsysroot/boot/"$(grep initrd $tmpfile | cut -d' ' -f2)" \
-           --parmfile $tmpfile
+           --parmfile $optfile
     fi
 
     echo "Rebooting"


overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
```diff
@@ -15,7 +15,10 @@ install() {
         chroot \
         sync \
         bwrap \
-        env
+        env \
+        zipl
```
cgwalters, line 19: Don't we need to make this architecture conditional?  It looks like in upstream dracut there's some usage of
`case "${DRACUT_ARCH:-$(uname -m)}" in`

nikita-dubrovskii: Sure, will do

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -59,9 +59,15 @@ firstboot() {
     mkdir -p "${tmpsysroot}/boot"
     mount /dev/disk/by-label/boot "${tmpsysroot}/boot"
 
+    # https://github.com/dracutdevs/dracut/blob/master/modules.d/01fips/fips.sh#L146
+    # Dracut looks for .vmlinux-xyz.hmac, which in our case is installed not under /boot,
+    # but under /boot/ostree/rhcos-xxxyyyzz/. To avoid installation failure we have to
+    # specify BOOT_IMAGE karg.
```
jlebon, line 65: On GRUB-supported platforms, the `BOOT_IMAGE` arg is already injected by GRUB into the command-line so this should Just Work. For s390x, we should check how the dracut code is intended to work, since this is something that traditional RHEL would hit as well.

nikita-dubrovskii: Here comes interesting part - `zipl` by default sets `BOOT_IMAGE=0` (https://github.com/ibm-s390-linux/s390-tools/blob/master/zipl/boot/menu.c#L188). And because we don't use `grub`, `dracut` fails. 
Also there is some strange issue during `cosa build` : 
```
dracut: *** Including module: fips ***
mknod: /tmp/dracut/dracut.yRQ17k/initramfs/dev/random: Operation not permitted
dracut: Cannot create /dev/random
dracut: To create an initramfs with fips support, dracut has to run as root
```

nikita-dubrovskii: Also i wonder why https://github.com/ostreedev/ostree/pull/1962/ doesn't work.

```
[    5.802578] dracut-pre-trigger[719]: Mounting /dev/disk/by-label/boot as /boot 
[    5.841955] SGI XFS with ACLs, security attributes, no debug enabled 
[    5.876049] XFS (dasda2): Mounting V5 Filesystem 
[    5.882649] EXT4-fs (dasda1): mounted filesystem with ordered data mode. Opts: (null) 
[    5.883933] dracut-pre-trigger[719]: Checking integrity of kernel 
[    5.886162] dracut: FATAL: FIPS integrity test failed 
[    5.886166] dracut: Refusing to continue 
[    5.886269] dracut-pre-trigger[719]: Warning: /boot//.vmlinuz-4.18.0-240.22.1.el8_3.s390x.hmac does not exist 
[    5.886280] dracut-pre-trigger[719]: Warning: /boot//.vmlinuz-4.18.0-240.22.1.el8_3.s390x.hmac does not exist 
[    5.917621] XFS (dasda2): Ending clean mount 
[    5.917951] XFS (dasda2): Quotacheck needed: Please wait. 
[    5.982502] systemd[1]: Shutting down. 
```


Please correct me if i'm wrong - `rpm-ostree` doesn't create anything during install (because we just copy `osmet` image) and later during `firstboot` . Or should it create those `*.hmac` during `cosa build***` ?

jlebon: >  Also there is some strange issue during cosa build :

We can ignore those issues, see: https://github.com/coreos/rpm-ostree/pull/1951

> Here comes interesting part - zipl by default sets BOOT_IMAGE=0 (https://github.com/ibm-s390-linux/s390-tools/blob/master/zipl/boot/menu.c#L188). And because we don't use grub, dracut fails.
> Also i wonder why ostreedev/ostree#1962 doesn't work.

Ahh OK, I think I see what's going on here.

So, on OSTree systems, the kernel, HMAC file, and initramfs are placed under `/boot/ostree/$name-$hash/`.

On GRUB systems, the `BOOT_IMAGE` is a path to the kernel image, and the dracut FIPS code uses that to also find the HMAC file.

It looks like on s390x, the `BOOT_IMAGE` is an integer representing the index of the configuration that was booted into? Does this roughly correspond to the `version` field of the BLS config?

Anyway, I think what's going on here is that dracut doesn't know how to use this field and so just ignores it and falls back to trying it in the root of `/boot`. This works fine on traditional systems because the kernel and HMAC files are indeed placed at the root of the boot partition. So this only breaks on OSTree systems.

Hmm, ideally we would teach dracut how to use this index to figure out the kernel path from it (and so the HMAC path). Do you think mapping the index back to the BLS is feasible or is it too error-prone? I suspect there might not actually be any traces of where the configuration came from once it's written out by `zipl` (or maybe there's some extra metadata we can shove in there?). /cc @martinezjavier

I don't think we should pre-inject `BOOT_IMAGE` like this patch is doing. Definitely wouldn't want to do this on GRUB platforms, since it contains dynamic information about e.g. which devices was actually used. But even on s390x, it's possible that there are tools which expect an integer there for their purposes.

nikita-dubrovskii: >So, on OSTree systems, the kernel, HMAC file, and initramfs are placed under /boot/ostree/$name-$hash/.

exactly. and becuase 
>on s390x, the BOOT_IMAGE is an integer representing the index of the configuration that was booted

i had to modify BOOT_IMAGE + https://github.com/nikita-dubrovskii/s390-tools/commit/70dde0e7d64f39a6f4747306c15ebc053ba01b19 . I've asked `zipl` experts about this, maybe it would be possible to mimic grub here.

So with all those patches system runs (rescue shell) with `fips`, but `ignition` is skipped. 
Could `fips` affect `ignition` ? I use my own simple `rhcos.ign` augmented with https://github.com/coreos/coreos-assembler/blob/master/mantle/kola/tests/fips/fips.go#L25

jlebon: > I've asked zipl experts about this, maybe it would be possible to mimic grub here.

Oh that would be the ideal fix if possible!

> So with all those patches system runs (rescue shell) with `fips`, but `ignition` is skipped.

Hmm, not sure offhand. Can you paste all the logs in e.g. a gist? You should see the system boot, run `ignition-fetch`, inject `fips=1 boot=LABEL=boot`, and then immediately reboot (and then run through all the Ignition stages).

nikita-dubrovskii: Discussion about `zipl` is going on. But looks like we can not drop `BOOT_IMAGE=N` because some further tooling is relying on this. I've found why `ignition` was skipped and made an update, so now system runs thru all ignition steps, but anyhow i end up with a rescue shell. Here is an installation log - https://bugzilla.redhat.com/show_bug.cgi?id=1955621#c8 , maybe i've missed smth?

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -75,6 +84,8 @@ firstboot() {
               echo $(grep $line $f) >> $tmpfile
           done
       done
+      # we reboot before ignition runs, so let's keep this opt
```
nikita-dubrovskii, line 87: without this after reboot `/proc/cmdline` contains only `$ignition_firstboot`, so all ignition steps are skipped

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -75,6 +84,8 @@ firstboot() {
...
+      echo "ignition.firstboot" >> $tmpfile
```
nikita-dubrovskii, line 88: without this after reboot /proc/cmdline contains only $ignition_firstboot, so all ignition steps are skipped

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
nikita-dubrovskii: Here is a link to zipl BOOT_IMAGE issue - https://github.com/ibm-s390-linux/s390-tools/issues/78

jlebon: There's an even more obvious practical reason why hardcoding `BOOT_IMAGE` won't work: it'll become out of date as soon as the node upgrades to a new image with a different kernel.

@Prashanth684 Do you know the answer to

> Hmm, ideally we would teach dracut how to use this index to figure out the kernel path from it (and so the HMAC path). Do you think mapping the index back to the BLS is feasible or is it too error-prone?

Prashanth684: right now i see that BOOT_IMAGE is set to 0. i tried an rhcos upgrade and didn't see it change so not sure if that will help us. @nikita-dubrovskii does BOOT_IMAGE update to 1 when you do an upgrade?

jlebon: From my naive reading of https://github.com/ibm-s390-linux/s390-tools/blob/master/zipl/boot/menu.c, I think it represents the index of the menu entry that was selected, so 0 is expected. The question is whether we can go from there and map it back to e.g. the vmlinuz path. Because if so, then we can teach dracut to use it.

Prashanth684: oh..right...yeah it's always 0.  not sure if there is a good way to map it to the kernel/initramfs.let me ask around.

Prashanth684: so after talking to a few folks looks like there is no good way to map the index back to the BLS entry.

jlebon: In that case, I think I agree that https://github.com/ibm-s390-linux/s390-tools/issues/78 is the best way to solve this, even if it happens at "boot record writing" time. If we can't change the existing semantics of `BOOT_IMAGE=N`, then maybe best to have a separate variable for this, e.g. `BOOT_IMAGE_PATH`.

One hack we could do is to have ostree on s390x symlink all the kernel and HMAC files from the root of the boot partition, but... yuck. (And you get into the question of what to do if somehow two `vmlinuz` files named the same across deployments somehow don't have the same checksum.)

jlebon: Let's chat in https://github.com/ibm-s390-linux/s390-tools/issues/78.

miabbott:
I spoke with the OCP Multi Arch team and was told that FIPS on s390x has never been supported, so the original BZ has been closed in favor of an RFE - https://issues.redhat.com/browse/MULTIARCH-690

It was suggested that we can stop pursuing this fix and let the Multi Arch team do the necessary enablement of FIPS.

jlebon:
Chatted with @nikita-dubrovskii about this.

To re-iterate, IMO ideally we wouldn't carry any special code for this here but instead patch dracut as necessary. The reason it happens to work in non-OSTree s390x systems today is because kernel images are placed in the root of the boot partition. But any other system like OSTree which strays from that will trip on the same issue.

There's no room to change `BOOT_IMAGE` to conform the other arches because of backwards compatibility (see https://github.com/ibm-s390-linux/s390-tools/issues/78). One idea was to have libostree (specifically, the zipl bootloader backend) manually inject a `BOOT_IMAGE_PATH=/ostree/$osname-.../vmlinuz-...` karg when translating the BLS configs and then teach dracut's FIPS module to look for both `BOOT_IMAGE` and `BOOT_IMAGE_PATH`. There's some concerns however about using up precious kernel cmdline space (apparently, there's only room for 896 bytes of data).

Maybe instead, we should just roll with it and simply teach dracut how to use the `BOOT_IMAGE=N` karg. So here's my idea: we change dracut on s390x to first check the root of the boot partition as usual, but otherwise fall back to looking at BLS entry N (the value of `BOOT_IMAGE`) and extract the kernel path from there in order to find the HMAC file. This isn't fully correct, because it assumes that there are *only* BLS entries (which is safe to assume for FCOS/RHCOS), but we would've failed anyway so it's an improvement. (And worst case, validation simply fails still.)

_time passes_

OK, ended up writing a dracut patch for this:

```diff
diff --git a/modules.d/01fips/fips.sh b/modules.d/01fips/fips.sh
index 821c26a2..fcaa798e 100755
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -124,6 +124,21 @@ do_fips() {
     else
         BOOT_IMAGE="$(getarg BOOT_IMAGE)"
 
+        # On s390x, BOOT_IMAGE isn't a path but an integer representing the
+        # entry number selected. Let's try the root of /boot first, and
+        # otherwise fallback to trying to parse the BLS entries if it's a
+        # BLS-based system.
+        if [[ $(uname -m) == s390x ]]; then
+            if [ -e "/boot/vmlinuz-${KERNEL}" ]; then
+                BOOT_IMAGE="vmlinuz-${KERNEL}"
+            elif [ -d /boot/loader/entries ]; then
+                bls=$(ls -v -r /boot/loader/entries/*.conf | sed -n "$((BOOT_IMAGE + 1))p")
+                if [ -e "${bls}" ]; then
+                    BOOT_IMAGE=$(grep ^linux "${bls}" | cut -d' ' -f2)
+                fi
+            fi
+        fi
+
         # Trim off any leading GRUB boot device (e.g. ($root) )
         BOOT_IMAGE="$(echo "${BOOT_IMAGE}" | sed 's/^(.*)//')"
```

@nikita-dubrovskii Do you want to try that and hack on top of it if needed?

madeelibm:
@jlebon @nikita-dubrovskii  I have tested the fips.sh changes and it works flawlessly!

jlebon:
@madeelibm Thanks for testing, I opened https://github.com/dracutdevs/dracut/pull/1539 for it.

jlebon:
@nikita-dubrovskii Can you drop the second commit in this PR?

nikita-dubrovskii:
> @nikita-dubrovskii Can you drop the second commit in this PR?

i did. today had no time to test it, but i guess we anyhow should keep `sed -e "/^options / s/$/ ignition.firstboot /" -i "$tmpfile"`
tomorrow will check

madeelibm:
ignition.firstboot is needed, please see: https://github.com/openshift/os/pull/565

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -70,16 +70,22 @@ firstboot() {
...
+      optfile=$(mktemp)
       for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
           for line in title version linux initrd options; do
               echo $(grep $line $f) >> $tmpfile
           done
       done
+
+      echo "Appending 'ignition.firstboot' to ${tmpfile}"
+      sed -e "/^options / s/$/ ignition.firstboot/" -i "$tmpfile"
+      grep options $tmpfile | cut -c 9- > $optfile
```
jlebon, line 82: Let's use `cut -d' ' -f2-` instead?

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -70,16 +70,22 @@ firstboot() {
...
       tmpfile=$(mktemp)
+      optfile=$(mktemp)
       for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
           for line in title version linux initrd options; do
               echo $(grep $line $f) >> $tmpfile
           done
       done
+
+      echo "Appending 'ignition.firstboot' to ${tmpfile}"
+      sed -e "/^options / s/$/ ignition.firstboot/" -i "$tmpfile"
```
jlebon, line 81: Optional: I think we could combine these two steps the way it's done here: https://github.com/coreos/coreos-assembler/blob/0a5bb5180acf4bdcea224a31b2ed1394a3297c6d/src/create_disk.sh#L359.

Though... looking at that `create_disk.sh` code, I think it's also missing a `| cut -d' ' -f2-`. So right now it's passing the `options` keyword into `--parmfile`. I'm guessing `zipl` just treats it as another karg, right? Looks like the original patch where `zipl` support was added does not have this issue: https://github.com/coreos/coreos-assembler/commit/100c2e512ecb89786a53bfb1c81abc003776090d.

jlebon: https://github.com/coreos/coreos-assembler/pull/2233

nikita-dubrovskii: >I'm guessing zipl just treats it as another karg, right?

yep

jlebon (approved):
Thanks!

PR #547, opened 2021-05-05
ci: Test bootupd

cgwalters:
Don't merge.

### DIFF
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -117,6 +117,7 @@ case "$(arch)" in
         # will implement "project-owned tests run in the pipeline"
         # and be able to run the existing bootupd tests:
         # https://github.com/coreos/fedora-coreos-config/pull/677
+        # dummy change for testing
         bootupctl status
         ok bootupctl
         ;;


cgwalters:
One thing I was investigating is whether there was bootupd version drift, but it seems we have `0.2.5` everywhere - and in any case there's not much different between 0.2.4 and 0.2.5 in this respect.

miabbott:
For the sake of completeness, this is the error we saw in the internal RHCOS pipeline (spotted by @travier)

```
+ /usr/bin/bootupctl backend generate-update-metadata /
error: file /boot/efi/EFI/BOOT/fbx64.efi: No such file or directory
error: file /boot/efi/EFI/redhat/grubx64.efi: No such file or directory
error: file /boot/efi/EFI/BOOT/BOOTX64.EFI: No such file or directory
error: file /boot/efi/EFI/redhat/shimx64.efi: No such file or directory
error: file /boot/efi/EFI/redhat/mmx64.efi: No such file or directory
error: file /boot/efi/EFI/redhat/shimx64-redhat.efi: No such file or directory
error: file /boot/efi/EFI/redhat/BOOTX64.CSV: No such file or directory
error: generating metadata failed: Failed to invoke rpm -qf
error: bwrap(/usr/bin/rpmostree-postprocess-inline-0): Child process killed by signal 1
```

miabbott:
CI job shows the same error noted above

cgwalters:
OK I verified that using `quay.io/coreos-assembler/coreos-assembler:v0.11.0` works.  But no idea *why* yet.

cgwalters:
Let's land https://github.com/openshift/release/pull/18338 then we'll do a PR here to change our Dockerfile to use it.

jlebon:
I *think* the issue might be https://github.com/openshift/os/pull/548. Something like: base rpm doesn't understand the rpmdb, so `rpm -qf` gets confused.

jlebon:
Looking at https://github.com/rpm-software-management/rpm/blob/847c6f062c267c4be643be9c202141bd330a7891/lib/query.c#L466-L478 at least, it does seem like `rpm -qf` falls back to doing a `stat` if it can't query the rpmdb and then printing the error message we see in the compose because it can't find the file (which is expected during the treecompose).

cgwalters:
OK, our CI is working again.

PR #548, opened 2021-05-05
manifest.yaml: explicitly use `rpmdb: bdb`

jlebon:
With cosa moving to f34, the rpm-ostree default rpmdb backend there
changed to sqlite. But for RHCOS we want to stick with bdb since that's
what el8 rpm understands.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -13,6 +13,11 @@ include:
   # RHCOS owned packages
   - rhcos-packages.yaml
 
+# The Fedora default is sqlite now, but we want to stick with bdb.
+# XXX: set back to `bdb` once rpm-ostree is fixed:
+# https://github.com/coreos/rpm-ostree/pull/2808
+rpmdb: b-d-b
+
 arch-include:
   x86_64:
     - fedora-coreos-config/manifests/grub2-removals.yaml


cgwalters (approved):

jlebon:
> error: Parsing src/config/manifest.yaml: serde-yaml: rpmdb: unknown variant `bdb`, expected one of `b-d-b`, `sqlite`, `n-d-b` at line 17 column 8

Ohhh yuck.

OK, will fix this in rpm-ostree, but for unblocking RHCOS, let's just use `b-d-b` and then we can switch it back once we have the fix. :arrow_up:

jlebon:
But actually, there's a much larger problem here. RPM will drop support for writing BDB in the near future, so we won't even be able to compose to bdb.

This is going to require major changes in rpm-ostree because right now we're directly using the librpm from the build environment to write out the rpmdb.

Hmm, OK it looks like there are plans to backport sqlite *read* support to el8, but not write (https://bugzilla.redhat.com/show_bug.cgi?id=1938928). So e.g. package layering wouldn't work. I guess what we could do then is we could detect that case and auto-convert back to bdb... yuck.

Anyway, to unblock RHCOS let's stick with f33 cosa for now

jlebon:
Ahhh, looks like that is the case already:

```
warning: using dummy database, installs not possible
```

jlebon:
Created an `rhcos-4.8` branch in cosa at the v0.11.0 tag and opened MRs internally for making use of it.

jlebon:
> Hmm, OK it looks like there are plans to backport sqlite _read_ support to el8, but not write ([bugzilla.redhat.com/show_bug.cgi?id=1938928](https://bugzilla.redhat.com/show_bug.cgi?id=1938928)). So e.g. package layering wouldn't work. I guess what we could do then is we could detect that case and auto-convert back to bdb... yuck.

Well, it's not as bad actually, because we could convert to bdb at treecompose time. I.e. rpm-ostree would first use librpm to create the sqlite rpmdb as it currently does, and then immediately run `rpm` in a bwrap container in the rootfs to have it converted into bdb. But would be nicer of course if we had https://bugzilla.redhat.com/show_bug.cgi?id=1938928#c2 and left bdb behind.

cgwalters:
The reason I pushed really hard to use Fedora as the build environment is that we were bit badly in the Atomic Host days around the "bootstrap problem" - every time we added a new feature we wanted to use at build time, we needed to prepare a build to ship to customers just so we could use it at build time - i.e. the self hosting problem, which is the same one that slows down RPM evolution actually.

But now that C8S exists, in theory we could create a cosa based on C8S and have much more freedom to iterate on e.g. new rpm-ostree builds and actually use them to ship quickly.  (While still having the code trickle down to RHEL after a cycle)

So the answer may be for RHCOS8 that we cut a C8S coreos-assembler - we could probably stick with Fedora for RHCOS9 of course, but eventually the same C9S will apply.

PR #550, opened 2021-05-13
faq: Link to `single` docs from FCOS

cgwalters:
In prep for getting this into product docs.


bgilbert:
Last time I checked, `single` doesn't work on RHCOS unless a root password is set.  Has that changed?

bgilbert:
Looks like it works in OCP 4.7 and above.

PR #557, opened 2021-05-21
ci: use RHEL 8.3 for RHCOS 4.7

miabbott:
Update the CI test to fetch RHEL 8.3 repos.

This should be reverted once we intend to start using RHEL 8.4 in
RHCOS 4.7.

See:  openshift/release#18664


miabbott:
Are we blocking merges based on the bug rules?  This one is only going to affect CI for 4.7, so seems safe to merge without a BZ

PR #560, opened 2021-06-04
overlay: split gcp-routes in its own overlay

vrutkovs:
This commit moves `gcp-routes` scripts, services and service-presets to a separate overlay.

This is useful for okd-machine-os, as it would like to reuse gcp-routes script and service instead of copying it. However OKD can't use the whole 05rhcos overlay as it refers to services which are not installed in OKD.


miabbott:
Seems safe enough; will let others weigh in.

/approve

travier (approved):

PR #561, opened 2021-06-08
system-preset: reduce priority for disabled services

lucab:
This renames `rhcos-disabled` presets to appear later in sorting
order, and specifically after `20-ignition.preset`.
The file contains entries to disable some services including
`afterburn` and `kdump`. While those services should be disabled
by default, users have to be able to selectively allow them
through Ignition.
As such RHCOS presets should always sort after `20-ignition.preset`
to avois shadowing user configuration.

NB: for preset files, "the entry in the file with the lexicographically
earliest name will be applied".

Ref: https://www.freedesktop.org/software/systemd/man/systemd.preset.html#Preset%20File%20Format
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1969208


lucab:
For reference, this is how RHCOS 4.7 presets look like before this change:
```
# ls -l {/usr/lib,/etc}/systemd/system-preset/
/etc/systemd/system-preset/:
total 4
-rw-r--r--. 1 root root 49 Jun  8 12:10 20-ignition.preset

/usr/lib/systemd/system-preset/:
total 28
-rw-r--r--. 2 root root  363 Jan  1  1970 10-rhcos-disabled.preset
-rw-r--r--. 2 root root 1419 Jan  1  1970 40-coreos.preset
-rw-r--r--. 2 root root  839 Jan  1  1970 43-manifest-rhcos.preset
-rw-r--r--. 2 root root  264 Jan  1  1970 85-display-manager.preset
-rw-r--r--. 2 root root 3846 Jan  1  1970 90-default.preset
-rw-r--r--. 2 root root  951 Jan  1  1970 90-systemd.preset
-rw-r--r--. 2 root root   10 Jan  1  1970 99-default-disable.preset
```

jlebon (approved):
Nice catch!

lucab:
For the sake of completeness, there is also another preset affecting kdump later on:
```
# grep kdump.service /usr/lib/systemd/system-preset/90-default.preset 
enable kdump.service
```

So, we do have to sort the RHCOS default after `20-ignition.preset` but before `90-default.preset`.
I picked `35-` arbitrarily with the intent of keeping RHCOS presets in the range [30, 50).

PR #569, opened 2021-06-22
Update faq.md

SirVL:
Update the URL of the screenshot since the old URL leads to a 404 page.


jlebon:
/approve

We'll just have to revert this when the OpenShift repos move to `main`.

cgwalters (approved):
/lgtm

PR #572, opened 2021-06-28
OWNERS: update with current CoreOS members

miabbott:


miabbott:
Saw that Luca could not `/lgtm` on https://github.com/openshift/os/pull/571 and made no sense that we don't have everyone on the team as part of the OWNERS

ashcrow (approved):

cgwalters:
I think it would work to symlink this to one of the other versions too.  But we can do that as a followup
/lgtm

PR #574, opened 2021-06-28
overlay: combine reboots for FIPS and Ignition kargs

arithx:
The FIPS and Ignition kargs reboots are entirely separate right now, and also race with each other.  Afterburn checkin also races with Ignition kargs because `ignition-kargs.service` doesn't handle the reboot itself.  Sequence all of these to avoid the races, and have the FIPS code write the same flag file read by `coreos-kargs-reboot.service` so it handles reboots for both.

Also use `rdcore kargs` to update the BLS entry rather than open-coding it, and do some other minor cleanups.

This should be covered by the existing kargs and FIPS tests.

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -86,8 +86,10 @@ firstboot() {
            --parmfile $optfile
     fi
 
-    echo "Rebooting"
-    systemctl --force reboot
+    echo "Scheduling reboot"
+    # Write to /run/coreos-kargs-reboot to inform the reboot service so we
+    # can apply both kernel arguments & FIPS without multiple reboots
+    > /run/coreos-kargs-reboot
 }
 
 finish() {
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -28,7 +28,6 @@ firstboot() {
     ign_usercfg_msg=$(journalctl -q MESSAGE_ID=57124006b5c94805b77ce473e92a8aeb IGNITION_CONFIG_TYPE=user)
     if [ -z "${ign_usercfg_msg}" ]; then
         noop "No Ignition config provided."
-        exit 0
     fi
     if [ ! -f "${IGNITION_CONFIG}" ]; then
         fatal "Missing ${IGNITION_CONFIG}"
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -53,16 +53,10 @@ firstboot() {
             ;;
     esac
 
-    echo "FIPS mode required; updating BLS entries"
+    echo "FIPS mode required; updating BLS entry"
 
-    mkdir -p "${tmpsysroot}/boot"
-    mount /dev/disk/by-label/boot "${tmpsysroot}/boot"
-
-    for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
-        echo "Appending 'fips=1 boot=LABEL=boot' to ${f}"
-        sed -e "/^options / s/$/ fips=1 boot=LABEL=boot/" -i "$f"
-    done
-    sync -f "${tmpsysroot}/boot"
+    rdcore kargs --boot-device /dev/disk/by-label/boot \
+        --append fips=1 --append boot=LABEL=boot
 
     if [[ $(uname -m) = s390x ]]; then
       # Similar to https://github.com/coreos/coreos-assembler/commit/100c2e512ecb89786a53bfb1c81abc003776090d in the coreos-assembler
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
@@ -22,12 +22,13 @@ ConditionKernelCommandLine=ignition.platform.id=azure
 # userdata, so we need to wait until after Ignition fetch.
 # (Waiting for fetch-offline is not sufficient if the config
 # references network resources, which it usually does in RHCOS.)
-#
-# In addition, the kargs stage might reboot the machine, after
-# which we need to be able to fetch the config again.  Removal
-# of the virtual CD persists across reboots, so we need to defer
-# it until after the kargs reboot.
-After=ignition-fetch.service ignition-kargs.service
+After=ignition-fetch.service
+
+# FIPS or Ignition kargs may schedule a reboot of the machine, which
+# happens in coreos-kargs-reboot.service.  Afterward we need to be
+# able to fetch the config again.  Removal of the virtual CD persists
+# across reboots, so defer it until after the kargs reboot.
+After=coreos-kargs-reboot.service
 
 [Service]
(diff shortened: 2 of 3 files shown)


overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/rhcos-early-reboot.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+if [ -f "/run/ignition-modified-kargs" ] || [ -f "/run/fips-modified" ]; then
+    echo "Rebooting"
+    systemctl --force reboot
```
cgwalters, line 7: We could split the conditionals and use e.g. `systemctl reboot --message="rhcos-early-reboot: modified kargs" to help admins (and us) debug *why* a reboot happened later (and systemd also nicely groups together the reboot flag with the message in the journal).

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/coreos-kargs-reboot.service
```diff
@@ -0,0 +1,24 @@
...
+# This runs after ignition-kargs & before ignition-disks so that it can optionally reboot
+# if kargs were modified via Ignition. This is done in a two-stage fashion so that other
+# mechanisms which may want to reboot (e.x. FIPS) can also hook in here and only reboot
+# once from the initrd.
+After=ignition-kargs.service
+Before=ignition-disks.service
+
+# On RHCOS we also need to run after FIPS and reboot in the case that
+# either Ignition kargs or FIPS has performed modifications.
+After=rhcos-fips.service
```
cgwalters, line 16: Totally fine as is, though in the future we could add our own little synthetic `coreos-ignition-extra.target` or something?  OTOH, hopefully it's unlikely that we discover a 3rd reason to reboot.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/coreos-kargs-reboot.service
```diff
@@ -0,0 +1,24 @@
...
+Before=ignition-complete.target
+ 
+# This runs after ignition-kargs & before ignition-disks so that it can optionally reboot
+# if kargs were modified via Ignition. This is done in a two-stage fashion so that other
+# mechanisms which may want to reboot (e.x. FIPS) can also hook in here and only reboot
+# once from the initrd.
+After=ignition-kargs.service
+Before=ignition-disks.service
+
+# On RHCOS we also need to run after FIPS and reboot in the case that
```
cgwalters, line 14: Offhand...I think nothing stops us from moving the FIPS bits into FCOS too.  It'd be an open question whether and how much we test it.  Definitely don't need to change it in this PR, just thinking out loud.

cgwalters (approved):
LGTM overall!  Mostly just minor optionals.

Could have a test in this repo too, but not a blocker.

bgilbert:
Ready for review!

travier:
I vaguely remember a dicussion where we said we could not combine both as FIPS enablement requires that all Ignition remote fetches should happen after FIPS is enabled thus making it impossible to wait for the config to reboot. How are we addressing that?

darkmuggle:
> I vaguely remember a dicussion where we said we could not combine both as FIPS enablement requires that all Ignition remote fetches should happen after FIPS is enabled thus making it impossible to wait for the config to reboot. How are we addressing that?

FIPS requires that any cryptographic operations happen using an approved cipher, else the operation is considered done in plain text. So long as we are not persisting confidential information to disk and _reboot_ before performing any file or disk operation, then we would be fine.

bgilbert:
Note that the kargs stage uses the same config handling as FIPS: the config is re-fetched after reboot.

@darkmuggle clarified OOB that kargs can't count as "persisting confidential information" since `/boot` is always unencrypted.

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/fips-reboot.conf
```diff
@@ -0,0 +1,12 @@
+[Unit]
+Description=CoreOS Early Reboot
+# We need to convert /run/ignition-modified-kargs into a triggering
```
jlebon, line 3: Hmm, maybe instead the flag file should just be called something more agnostic upstream. E.g. `/run/coreos-reboot-required` or something. Then we can just have it touch the same file and avoid this drop-in.

Ahh, except `rdcore kargs --create-if-changed` uses `O_EXCL` so we'd have to run the FIPS code *after* `ignition-kargs.service` and not before.

bgilbert: The other thing we could do is just make it a triggering condition upstream, with a single trigger.

bgilbert: https://github.com/coreos/fedora-coreos-config/pull/1135

bgilbert: https://github.com/coreos/fedora-coreos-config/pull/1135

jlebon: WFM though I still wonder whether we really need multiple triggering conditions at all. Why not simplify further and have a single "reboot required" trigger? Avoids having to ship new dracut modules just to add drop-ins entirely.

bgilbert: Okay, I can make that change.  How would you feel about dropping the O_EXCL so we're not managing that ordering dependency indefinitely?

jlebon: Yeah, makes sense. I can take care of that.

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
```diff
@@ -54,16 +53,10 @@ firstboot() {
...
-    mkdir -p "${tmpsysroot}/boot"
-    mount /dev/disk/by-label/boot "${tmpsysroot}/boot"
-
-    for f in "${tmpsysroot}"/boot/loader/entries/*.conf; do
-        echo "Appending 'fips=1 boot=LABEL=boot' to ${f}"
-        sed -e "/^options / s/$/ fips=1 boot=LABEL=boot/" -i "$f"
-    done
-    sync -f "${tmpsysroot}/boot"
+    rdcore kargs --boot-device /dev/disk/by-label/boot \
+        --append fips=1 --append boot=LABEL=boot
```
cgwalters, line 59: (visualize this with me holding pom-poms and jumping up and down)
Go go more Rust!  Down with shell!
Go go more Rust!  Down with shell!

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/fips-reboot.conf
bgilbert: Went ahead and created https://github.com/coreos/coreos-installer/pull/577 so I'd have something to link to.

cgwalters (approved):
(Only re-skimmed discussion since I last reviewed, this is a superficial LGTM)

overlay.d/05rhcos/usr/lib/dracut/modules.d/41rhcos-early-reboot/fips-reboot.conf
bgilbert: The flag file change was https://github.com/coreos/fedora-coreos-config/pull/1136, for the record.

bgilbert:
Updated!  And integrated the changes with Afterburn checkin.

jlebon (approved):
Nice!
/lgtm

PR #576, opened 2021-07-09
Bump fedora-coreos-config

bgilbert:
Note that `overlay: remove /boot/ignition on upgrade if present` is FCOS-specific.

```
Ben Howard (1):
      Buildroot: add golang

Benjamin Gilbert (6):
      dracut: drop references to ignition-setup-base.service
      overlay: remove /boot/ignition on upgrade if present
      35coreos-ignition: depend on coreos-live; assume is-live-image exists
      35coreos-ignition: add ignition-setup-user.service from Ignition repo
      35coreos-ignition: update ignition-setup-user for this repo
      40ignition-ostree: update an ignition-fetch dep to fetch-offline

Dusty Mabe (2):
      image-base.yaml: set bootfs_metadata_csum_seed to true
      Revert "kola-denylist.yaml: add ext.config.podman.rootless-systemd"

Jonathan Lebon (13):
      35coreos-ignition: use cp instead of mv
      coreos-propagate-multipath-conf: trigger emergency.target on failure
      coreos-propagate-multipath-conf: run After=initrd-root-fs.target
      40ignition-ostree: explicitly add zram kmod in initrd
      ci/buildroot: bump nocache hack to defeat Quay.io caching
      05core: add coreos-ignition-firstboot-complete.service
      15fcos: add missing dash in "user provided"
      15fcos: make "ssh" all caps
      05core: move coreos-ignition-firstboot-complete logic to separate script
      overlay.d: drop .sh extensions on /usr/libexec shell scripts
      05core/firstboot-complete: also delete /boot/ignition if it exists
      35coreos-live: stop overriding NetworkManager-wait-online timeout to 5s
      overrides: fast-track fedora-coreos-pinger-0.0.4-11.fc34
```


cgwalters:
Note this pulls in ` image-base.yaml: set bootfs_metadata_csum_seed to true` but it's a no-op because we have a separate `image.yaml` here which does not inherit the FCOS one.

/lgtm

Issue #552, opened 2021-05-14
be buildable with f34 cosa somehow, or bust


cgwalters:
Followup to https://github.com/openshift/os/pull/547 and https://github.com/openshift/os/pull/548

Right now we're in an ugly situation where because f34 rpm dropped support for bdb, yet rhel8 only works with bdb we cannot use f34 to build rhcos(8).

Options:

- Just require the `rhcos-4.8` cosa-f33 branch for a while (6 months?) and see if the rpm maintainers end up doing something here (e.g. sqlite support in rhel8).
- jlebon's idea of somehow using the host librpm to write as sqlite, and then pipe the data to the target rpm to downconvert to bdb  (I think we'd need to carry f33 rpm for this which is the only one that does both?  How ugly would it be to e.g. build a statically linked f33 rpm binary and include it in our container?)  Hmm, maybe we could convince the rpm maintainers to ship something like `/usr/lib/rpm/rpm-sqlite2bdb --dbin /path/to/srcdb --dbout /path/to/destdb` in rhel8?
- Switch over to using a C8S coreos-assembler for *both* fcos and rhcos
- Switch over to C8S coreos-assembler just for rhcos

Or something else?

cgwalters:
jlebon correctly found https://bugzilla.redhat.com/show_bug.cgi?id=1938928 but I lost track of that BZ in the PRs/issues here.  I commented there, I think rpm sqlite readonly support is enough.  Let's leave this open to track work doing the conversion.

cgwalters:
Ultimately the conversion code I think should land in rpm-ostree, i.e. all that happens here is we set `rpmdb: bdb` and rpm-ostree knows to do the dance internally instead of us doing it as a postprocess script or whatever.

jlebon:
Yeah, I think that's the easiest approach, even if it's not great. One note is that the RHBZ is planned for 8.5, so we might have to hold until RHCOS rebases.

miabbott:
Do we have a path forward on this issue?

4.8 is reaching code freeze soon, which means we would just start using `coreos-assembler:main` again.

Could we change the internal `coreos-assembler:main` BuildConfig to use F33?

```  strategy:
    dockerStrategy:
      env:
      - name: GIT_URL
        value: https://github.com/coreos/coreos-assembler
      - name: GIT_REF
        value: main
      - name: OPENSHIFT_GIT_HACK
        value: "TRUE"
      from:
        kind: ImageStreamTag
        name: fedora:33
      noCache: true
[...]

PR #577, opened 2021-07-09
manifest: Explicitly set rpmdb: bdb

cgwalters:
Since that's what it needs to be.  There is read-only
sqlite support queued in
https://bugzilla.redhat.com/show_bug.cgi?id=1938928
But we need read-write for package layering etc.

xref https://github.com/openshift/os/issues/552

This is effectively a no-op right now because the default
backend in F33 rpm-ostree is bdb, but by setting this
explicitly we'll get a useful error message after
https://github.com/coreos/rpm-ostree/pull/2978

And then ideally we teach rpm-ostree how to actually deal
with this.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -31,6 +31,9 @@ repos:
   - rhel-8-fast-datapath-aarch64
   - rhel-8-server-ose
 
+# https://bugzilla.redhat.com/show_bug.cgi?id=1938928
+rpmdb: b-d-b
+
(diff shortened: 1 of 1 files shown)


cgwalters:
OK right https://github.com/coreos/rpm-ostree/pull/2979

miabbott:
/lgtm

Great work getting us off the `rhcos-4.8` version of `cosa`!

PR #579, opened 2021-07-13
manifest: chmod 600 /etc/sudoers.d/*

bgilbert:
Apparently there are security scanners that object to mode 644.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1981979.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -161,6 +161,13 @@ postprocess:
      # NB: we don't use -f here so we break when this is no longer needed
      rm -v /etc/iscsi/initiatorname.iscsi
 
+  # Make security scanners happy
+  # https://bugzilla.redhat.com/show_bug.cgi?id=1981979
+  - |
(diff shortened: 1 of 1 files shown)


miabbott:
The same permissions exist on FCOS; I think this should be moved to fedora-coreos-config

jlebon:
Wonder at this point if it'd be cleaner to just support a [statoverride](https://github.com/ostreedev/ostree/blob/f216a3c170b8cd636241e9a86f0b87fb183859e7/man/ostree-commit.xml#L232-L236) file at e.g. `overlay.d/stat-override.txt` that we just feed to ostree while committing the overlays in cosa.

bgilbert:
Moved to fedora-coreos-config in https://github.com/coreos/fedora-coreos-config/pull/1111, using statoverride support added in https://github.com/coreos/coreos-assembler/pull/2293.

PR #580, opened 2021-07-13
rhcos-afterburn-checkin: wait for Ignition fetch and kargs stages

bgilbert:
If we check in before Ignition config fetch, Azure will remove the config drive and config fetch will fail.  If we check in before Ignition kargs and kargs reboots the machine, config fetch will fail on the next boot.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1980679.

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
@@ -4,7 +4,17 @@ ConditionKernelCommandLine=ignition.platform.id=azure
 
 # Since we don't care about the actual success of the boot
 # from the OS perspective, check in as soon as we are able.
-After=network.target
+#
+# On Azure, checkin causes removal of the virtual CD with the
+# userdata, so we need to wait until after Ignition fetch.
+# (Waiting for fetch-offline is not sufficient if the config
+# references network resources, which it usually does in RHCOS.)
+#
+# In addition, the kargs stage might reboot the machine, after
+# which we need to be able to fetch the config again.  Removal
+# of the virtual CD persists across reboots, so we need to defer
+# it until after the kargs reboot.
+After=ignition-fetch.service ignition-kargs.service
 
 [Service]
 Environment=AFTERBURN_OPT_PROVIDER=--cmdline


miabbott:
/approve

Makes sense, thanks!

jlebon (approved):
/lgtm

I wish we could make this service consistent with upstream/FCOS. I.e. either moving the latter to the initramfs, or making it more easily configurable to meet the requirements of OCP. E.g. OCP could just add a drop-in via Ignition to have the check-in happen earlier than it does in FCOS.

bgilbert:
> I wish we could make this service consistent with upstream/FCOS. I.e. either moving the latter to the initramfs, or making it more easily configurable to meet the requirements of OCP. E.g. OCP could just add a drop-in via Ignition to have the check-in happen earlier than it does in FCOS.

ISTM performing checkin before provisioning has completed is a fairly OpenShift-specific requirement.  In the general case (FCOS) I don't think we should check in until we've provisioned successfully, since otherwise checkin has no useful semantics.  Potentially the OS could support both paths, and check the Ignition config for an `/etc/coreos/azure-checkin-early` flag that would enable/disable the corresponding units.

cgwalters:
> ISTM performing checkin before provisioning has completed is a fairly OpenShift-specific requirement. 

Probably?  But I don't fully understand the history behind this.  What *is* the requirement?  Jerry's message mentions Terraform but...that doesn't help that much.  IIRC there's a BZ for this that is probably lost to time.  [My search](https://bugzilla.redhat.com/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=POST&bug_status=MODIFIED&bug_status=ON_DEV&bug_status=ON_QA&bug_status=VERIFIED&bug_status=RELEASE_PENDING&bug_status=CLOSED&component=RHCOS&email1=walters%40redhat.com&emailassigned_to1=1&emailcc1=1&emailreporter1=1&emailtype1=substring&f0=OP&f1=OP&f3=CP&f4=CP&j2=OR&list_id=12012619&longdesc=ignition%20azure&longdesc_type=allwordssubstr&product=OpenShift%20Container%20Platform&query_format=advanced) doesn't seem to find it.



>  Potentially the OS could support both paths, and check the Ignition config for an /etc/coreos/azure-checkin-early flag that would enable/disable the corresponding units.

Right, something like that if we have to.

cgwalters:
Ahh wait maybe it is there: https://bugzilla.redhat.com/show_bug.cgi?id=1756175#c7 right?

PR #581, opened 2021-07-14
rhcos-afterburn: Add original downstream commit message

cgwalters:
When we imported the internal git repo to here we lost all the
history.  Since we touched this file recently and I found
myself wondering why it exists, let's copy the exact git commit
message (and timestamp) here so it isn't lost.

This file already has a comment which this partially duplicates,
but the commit message e.g. references Terraform and kube
which the existing comment does not.

I think we should try to dig up more of the original discussion
around this because we're already being bitten by this
nontrivial divergence between FCOS/RHCOS and we should think
really hard about a way to have OpenShift/Terraform work
well without this divergence.


jlebon:
Yeah, agreed with last paragraph.

/lgtm

PR #582, opened 2021-07-14
Bug 1982002: [4.8] rhcos-afterburn-checkin: wait for Ignition fetch stage

bgilbert:
If we check in before Ignition config fetch, Azure will remove the config drive and config fetch will fail.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1982002.  Backports #580.


miabbott:
The BZ bot doesn't understand the semantics between fixing a BZ in the code here and actually **shipping** the fix as part of a boot image bump, so I am going to wave the invalid bug label.

miabbott:
@bgilbert WDYT about pulling in 950cbe6 as part of this for completeness sake?

bgilbert:
Done!

PR #588, opened 2021-07-19
[release-4.7] Bump fedora-coreos-config submodule to latest

jlebon:
```
Jonathan Lebon (1):
      35coreos-live: stop overriding NetworkManager-wait-online timeout to 5s
```

For https://bugzilla.redhat.com/show_bug.cgi?id=1967483.


miabbott:
While BZ#1967483 is fixed with this code change, actually **delivering** the fix to customers requires a boot image bump in the installer via BZ#1964753.  The BZ bot does not understand this dependency, so we have to manually remove the `invalid-bug` flag.

PR #589, opened 2021-07-23
Bump fedora-coreos-config

bgilbert:
Notable changes:

- Fix race in Ignition kargs reboot
- Fix race between Ignition kargs and `coreos-copy-firstboot-network.service`
- Fix race in `coreos-propagate-multipath-conf.service`
- `chmod 600 /etc/sudoers.d/coreos-sudo-group`

```
Benjamin Gilbert (7):
      overrides: fast-track Ignition 2.11.0-2.fc34
      tests/installer-cleanup: only run on QEMU
      ci: encourage adding Bodhi update link to fast-track overrides
      overlay: add empty statoverride files in all overlays
      overlay: chmod 600 /etc/sudoers.d/coreos-sudo-group
      35coreos-ignition: perform kargs reboot with --force
      coreos-kargs-reboot: rename flag file

Dusty Mabe (15):
      unfreeze dracut, adapt for NM via systemd in initrd
      overrides: fast-track dracut-055-3.fc34
      35coreos-network: order coreos-copy-firstboot-network before ignition-kargs
      35coreos-live: add requirement on loop kmod
      ci: don't do a strict build for mechanical streams
      lockfiles: fast-track selinux-policy-34.14-1.fc34
      lockfiles: fast-track ostree-2021.3-1.fc34
      tests/manual: silence audit messages to the console
      tests/manual: fcct is now known as butane
      lockfiles: fast-track container-selinux-2.164.1-1.git563ba3f.fc34
      manifests: remove outdated comment
      manifests: make the systemd-resolved neutering F34 only
      lockfiles: fast-track systemd-248.5-1.fc34
      lockfiles: fast-track kernel-5.13.4-200.fc34
      lockfiles: update bodhi link for kernel-5.13.4-200.fc34

Jonathan Lebon (7):
      35coreos-multipath: add Before=initrd-parse-etc.service in coreos-propagate-multipath-conf.service
      overrides: fast-track coreos-installer-0.9.1-2.fc34
      manifests: move fedora-coreos-pool to fedora-coreos.yaml
      ci: add `branched` to list of mechanical streams
      15fcos: order coreos-check-ignition-config.service before systemd-user-sessions.service
      15fcos: remember when Ignition ran and print on console
      ci/remove-graduated-overrides: query by arch too

Micah Abbott (1):
      tests: add test for watching files labeled kubernetes_file_t

Timothée Ravier (1):
      Revert "overlay.d/09misc: Fix mode for some files in /etc"
```


cgwalters (approved):

bgilbert:
CI is failing because it's not using the tip of the cosa `rhcos-4.8` branch, but an older commit coreos/coreos-assembler@0830918dc07506fe71f68c60d9cda58734895823 that doesn't have coreos/coreos-assembler@7f275cadae6b0bf613da61077f5549218898e8d7.  Do we need to manually bump a CI config?

bgilbert:
@darkmuggle reports that Prow is having issues.

cgwalters:
Wait wait, let's be careful about overriding CI here.  There's no longer another gate to ship code.

bgilbert:
@cgwalters According to @darkmuggle, Prow is not close to being in working order again, and he's working on a PR to disable it for this repo.  If you have a better idea, we have another shot in https://github.com/openshift/os/pull/574.

jlebon:
Now that https://github.com/openshift/os/issues/552 is fixed, we should be able to use git main of cosa again for CI, which should fix the missing statoverride issue.

cgwalters:
https://github.com/openshift/release/pull/20570

cgwalters:
CI passed on e.g. https://github.com/openshift/os/pull/586 which was pretty recent, I am not aware of something new and fundamental blocking working CI here.  

I for sure wouldn't want to run an operating system that wasn't gated rigorously on CI, and neither would our customers.  I'm not saying we can't override, sometimes it's needed.  But I think we also need to do our best to keep some CI going.

bgilbert:
@cgwalters I'm in favor of CI.  I was working on the advice of someone who knows the RHCOS CI situation much better than I.  With feature freeze happening today, it seemed better to land the code and then fix up any breakage during the stabilization period.

darkmuggle:
I have attempted to work with the Prow team to get this fixed. The root cause was narrowed down to a Prow bug where the Docker build root was getting a permission denied.

I suggested disabling the Prow CI simply because we can't test it today until the Prow bug is cleared. Further, we do have layers of CI where ART runs these tests in Jenkins behind the firewall. The tests here while value can fail differently since we are not using the same CI images and build roots. 

Obviously, I'm :100: for CI, but this test is redundant. The value in Prow based CI is the public visibility. I'm working on getting this fixed and have been all week.

cgwalters:
OK, sorry I wasn't aware of that.

> Obviously, I'm 100 for CI, but this test is redundant. The value in Prow based CI is the public visibility. 

I don't think that's quite true though, because it's really useful to have each PR tested individually rather than have the main build fail.  Plus, ultimately Prow e.g. has credentials for cloud based testing that we can start using, and we don't want to use the "upload image" credentials for that.

I think this will fix the new flakes:  https://github.com/openshift/release/pull/20591

cgwalters:
Thinking about this long term, ideally...we move the RHCOS build to use lockfiles just like FCOS - and those lockfile bumps are tested by Prow CI too.  That would basically remove the need to rerun most of the CI tests internally too.   And further, bootimage bumps to `openshift/installer` are already covered by CI - we could further add our own CI there.  And if we also did OS level testing for `machine-os-content` promotion jobs (which would be a big deal) then we could reduce the role of the internal flow.

bgilbert:
The changes in this PR passed CI in https://github.com/openshift/os/pull/574.

PR #591, opened 2021-07-23
rhcos-packages: Add bootupd

travier:


jlebon:
We're already pulling in bootupd from the `fedora-coreos-config/manifests/bootupd.yaml` includes, so a commit message here on what this is fixing would've been nice. I'm guessing this is related to the SCOS efforts, though we shouldn't have to guess. :)

travier:
Sorry, I went too fast on this one as I had not realized that it was included in the `manifest` and thought that it was mostly a declarative / documentation list.

The initial goal was to make sure that all packages that we maintain for RHCOS are listed in this file and `bootupd` was missing thus the addition. Whether or not all of those packages will be included in Stream for the SCOS effort remains to be determined.

PR #592, opened 2021-07-26
rhcos-packages: Make bootupd arch dependend

jschintag:
Do not include bootupd for s390x and ppc64le

### DIFF
--- a/rhcos-packages.yaml
+++ b/rhcos-packages.yaml
@@ -5,8 +5,6 @@ packages:
   ## Delivered via RHAOS puddles
   # Extra runtime
   - afterburn
-  # Bootloader updater
-  - bootupd
   # User experience
   - console-login-helper-messages-issuegen console-login-helper-messages-profile
   # installer iso packages
@@ -23,3 +21,9 @@ packages:
   # for kdump:
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
   - kexec-tools
+
+# See https://github.com/coreos/bootupd
+packages-x86_64:
+   - bootupd
+packages-aarch64:
+   - bootupd


travier:
Could you provide more context? What's the issue here?

travier:
Oh, it's included in `manifest.yaml`. Let's add a comment then instead of removing it?

jschintag:
@travier  The RPM for `bootupd` is not used for s390x and power, as such it should be arch specific, which i think it already is.

travier:
To match the existing format, could you make it like this: https://github.com/coreos/fedora-coreos-config/blob/35ba0419cd3daaf424c4c8d74654d51403932d2c/manifests/fedora-coreos-base.yaml#L194

jschintag:
It is already included in `manifest.yaml` https://github.com/openshift/os/blob/42cf9bfd20b57ea566583d1cc945881326b5a421/manifest.yaml#L19. Do you want it inside the `rhcos-packages.yaml` or just a comment referencing it?

Prashanth684:
also, if this does need to be added, it can be done with arch specific packages like:
```
packages-x86_64:
   - bootupd
packages-aarch64:
   - bootupd
```

jschintag:
I moved the includes into `rhcos-package.yaml` @travier

travier:
Sorry I wasn't clear enough. I'd prefer we keep the include in the original files to keep things clearer and keep this files as just a list of packages, with arch specific includes. We could even consider removing the include for this file from the manifest directly as it should be 100% redundent.

travier:
Either a comment or an arch include directly. It will be redundant as said in the comment as the top of the file but that's mostly for easier tracking.

jschintag:
I reverted my changes to `manifest.yaml`

travier (approved):
/lgtm

travier:
Fixes: https://github.com/openshift/os/pull/591
Fixes: https://github.com/openshift/os/pull/591/commits/170f78cac0b5014e53c1973d242b1c7ec71b29ed

PR #593, opened 2021-07-26

cgwalters:

let's settle for parallel.

Some discussion on that here:

around and make `podman run registry.ci.openshift.org/rhcos/rhcos:4.8` e.g.


cgwalters:
Hmm, right actually we're going OOM *inside* the VM, not outside.  I wonder if this is something like the size of the content set for RHCOS?  But it's *really* weird that we're consistently getting ENOMEM from `fstatat`.

cgwalters:
YES!  CI passed, can I get a lgtm?

kelvinfan001 (approved):
/lgtm

PR #596, opened 2021-07-28
extensions: Add oci-seccomp-bpf-hook

jhrozek:
The oci-seccomp-bpf-hook package enables recording of secomp profiles,
which, apart from being useful on its own, is a part of the Security
Profiles Operator:
    https://github.com/openshift/enhancements/pull/745

This package was previously evaluated for inclusion into FCOS:
    https://github.com/coreos/fedora-coreos-tracker/issues/887
and it was recommended that we try adding the hook as an extension
instead.

Note that bcc (a dependency of oci-seccomp-bpf-hook) seems to have a weak
dependency on bcc-tools which is not needed for this case. I'm unsure
if there is a way to tell the extensions system to not install the weak
dependency though.

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -11,6 +11,11 @@ extensions:
   usbguard:
     packages:
       - usbguard
+  # https://github.com/coreos/fedora-coreos-tracker/issues/887
+  # CMP-927
+  oci-seccomp-bpf-hook:
+    packages:
+      - oci-seccomp-bpf-hook
   # https://github.com/kmods-via-containers/kmods-via-containers/issues/3
   # https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/866
   # These are currently overlaid onto the host so that they can be bind-mounted


jhrozek:
So the build-test-qemu job failed installing kata containers which seem unrelated but it did make me realize that the dependencies for bcc/oci-seccomp-bpf-hook/crun were only fixed recently and might have not made its way into RHEL. I know that the bcc dep fixes will be included in 8.5, but I don't think before.
There is more details in https://issues.redhat.com/browse/CMP-927 and of course feel free to ask me for more details.

miabbott:
> So the build-test-qemu job failed installing kata containers which seem unrelated but it did make me realize that the dependencies for bcc/oci-seccomp-bpf-hook/crun were only fixed recently and might have not made its way into RHEL. I know that the bcc dep fixes will be included in 8.5, but I don't think before.

Kata failure is #597, but should be resolved as of this writing

RHCOS/OCP has no immediate plans to use RHEL 8.5 content, so the dependency fixes would need to be backported to RHEL 8.4.z or explicitly listed under the entry in `extensions.yaml` for the `oci-seccomp-bpf-hook` extension.

cgwalters:
Is this still pulling in LLVM?  I am quite wary of adding all of that to extensions even, it's a huge dependency and everyone who wants to apply OS updates needs to download it.

I'm not totally opposed to be clear, but this is really a large thing.

jhrozek:
I'm sorry this stalled for so long. First, summer happened and then 4.9 deadlines. Sorry. Let me try to restart the discussions and reply to your comments individually.

miabbott:
> Is this still the case and will be going forward? I was under the impression that OCP would follow RHEL-8 stream to use 8.5 at some point.

We found that the lifecycle for the odd-numbered RHEL 8 y-stream releases (8.3, 8.5, etc) is too short to be used by OCP/RHCOS without significant effort by multiple teams (RHEL dev, releng, RHCOS, etc) with not enough benefit for the extra effort. Additionally, we've found that switching RHEL 8 y-streams as part of an OCP z-stream release can introduce more problems that it solves.

Our working plan is to only use the even-numbered RHEL 8 y-stream releases for future OCP/RHCOS versions, which gains us an extended support lifecycle for the RHEL 8 bits. Additionally, we will "lock" an OCP/RHCOS release to a RHEL 8 EUS release for the duration of the OCP y-stream lifecycle.

This means that we will not have any OCP/RHCOS versions that consume RHEL 8.5 and will continue to use RHEL 8.4 EUS for the remaining duration of OCP 4.7/4.8/4.9 and likely OCP 4.10.

jhrozek:
[...]

> This means that we will not have any OCP/RHCOS versions that consume RHEL 8.5 and will continue to use RHEL 8.4 EUS for the remaining duration of OCP 4.7/4.8/4.9 and likely OCP 4.10.

Thank you for the informative reply. I will ask for the backports, then.

miabbott:
/remove-lifecycle stale

@jhrozek Is the inclusion of `oci-seccomp-bpf-hook` still desired in RHCOS?  OCP/RHCOS 4.11 will be using RHEL 8.6 content, so I think some of the dependency requirements would be solved.

jhrozek:
> /remove-lifecycle stale
> 
> @jhrozek Is the inclusion of `oci-seccomp-bpf-hook` still desired in RHCOS? OCP/RHCOS 4.11 will be using RHEL 8.6 content, so I think some of the dependency requirements would be solved.

OK, so to be frank, the only use-case I can think of right now is better support for seccomp profiles recording in the Security Profiles Operator. Because this PR seemed stalled, we came up with an alternative which was pretty much tailing audit.log and we need to use this mechanism to record SELinux profiles. We also have a built-in BPF recorder for the operator, but that one sadly needs very new libbpf and libelf releases which are not available in RHEL-8 at all, only RHEL-9.

I assume that once we add a new extension we have to keep and support it? If yes, then I would probably "park" this requirement, we can close this PR for the time being. SPO can go ahead and use watching audit.log for a start and move eventually to the built-in BPF recorder. This way, we wouldn't have to add anything to RHCOS and would keep all the functionality inside the operator.

If you know other consumers of the hook, then by all means, let's go ahead and add it.

Issue #597, opened 2021-07-28
`error: Packages not found: kata-containers`


cgwalters:
And now we have a *new* problem in CI.
```
 Importing rpm-md...done
rpm-md repo 'rhel-8-baseos'; generated: 2021-07-21T00:05:02Z solvables: 8801
rpm-md repo 'rhel-8-appstream'; generated: 2021-07-27T19:45:38Z solvables: 20182
rpm-md repo 'rhel-8-server-ose'; generated: 2021-07-28T12:54:53Z solvables: 338
rpm-md repo 'rhel-8-fast-datapath'; generated: 2021-06-21T14:20:59Z solvables: 431
rpm-md repo 'rhel-8-fast-datapath-aarch64'; generated: 2021-07-28T12:51:15Z solvables: 0
rpm-md repo 'rhel-8-nfv'; generated: 2021-07-20T20:32:19Z solvables: 468
rpm-md repo 'rhel-8-advanced-virt'; generated: 2021-07-06T13:17:39Z solvables: 1699
error: Packages not found: kata-containers 
```

cgwalters:
Indeed, on the internal mirror at least, the package just seems to have vanished?  Going to go check the CDN

fidencio:
@cgwalters, thanks for bringing this to my attention.  It's happening as there's no 4.9 build for kata-containers, which is unfortunate and we're doing a new build right now.

Sorry for the breakage.

cgwalters:
I'm not seeing a new build so I pushed https://github.com/openshift/os/pull/598 for now.

PR #595, opened 2021-07-27
kola-denylist: disable crio.base tests for all arches

miabbott:
See: https://bugzilla.redhat.com/show_bug.cgi?id=1986239

PR #598, opened 2021-07-28
extensions: Drop kata-containers for now

cgwalters:
Until a new version is built.

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -48,8 +48,9 @@ extensions:
   # GRPA-3123
   # - kata-containers (RHAOS)
   #   - qemu-kiwi (advanced-virt)
-  sandboxed-containers:
-    architectures:
-      - x86_64
-    packages:
-      - kata-containers
+  # FIXME temporarily removed due to https://github.com/openshift/os/issues/597
+  # sandboxed-containers:
+  #   architectures:
+  #     - x86_64
+  #   packages:
+  #     - kata-containers


darkmuggle:
Before merging do we have an ETA on when https://github.com/openshift/os/issues/597 might be cleared by the Kata folks?

cgwalters:
There's a build going now, if that succeeds let's close this.  

But it will also be useful to see if this PR hits any further CI issues past that, so let's keep it open.

cgwalters:
```
 kola -p qemu-unpriv --output-dir tmp/kola run --qemu-nvme=true basic
2021-07-28T18:28:57Z kola: yaml: unmarshal errors:
  line 10: cannot unmarshal !!seq into string 
```

Ooh!  Yep a *new* CI failure.  I bet this is from https://github.com/openshift/os/pull/595/files

fidencio:
> There's a build going now, if that succeeds let's close this.

The new build succeeded, by the way.  Thanks for the raising this to our attention and, as mentioned on #597, sorry for the breakage.

miabbott:
> ```
>  kola -p qemu-unpriv --output-dir tmp/kola run --qemu-nvme=true basic
> 2021-07-28T18:28:57Z kola: yaml: unmarshal errors:
>   line 10: cannot unmarshal !!seq into string 
> ```
> 
> Ooh! Yep a _new_ CI failure. I bet this is from https://github.com/openshift/os/pull/595/files

That's frustrating; I strongly believe I tested that change locally with `cosa:latest`...but anyways #599

cgwalters:
Yep `buildextend-extensions` works here now.

PR #601, opened 2021-07-30
Revert "image: Switch to `ostree-format: oci`"

cgwalters:
This reverts commit 8e0a9d20e8e286cf87ebe9442ba663c75bf3006e.

We need to ship https://github.com/ostreedev/ostree-rs-ext/pull/58
before we can do this.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -11,5 +11,3 @@ squashfs-compression: gzip
 # Disable networking by default on firstboot. We can drop this once cosa stops
 # defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
-
-ostree-format: "oci"


Prashanth684:
> This reverts commit [8e0a9d2](https://github.com/openshift/os/commit/8e0a9d20e8e286cf87ebe9442ba663c75bf3006e).
> 
> We need to ship [ostreedev/ostree-rs-ext#58](https://github.com/ostreedev/ostree-rs-ext/pull/58)
> before we can do this.

i am not sure how soon we can ship https://github.com/ostreedev/ostree-rs-ext/pull/58 but for now , the rhcos pipeline is using cosa 4.8, so really this should not be an issue.But i leave this up to you.

cgwalters:
> the rhcos pipeline is using cosa 4.8

I think we should switch rhcos 4.9 to use cosa latest.

saqibali-2k (approved):

PR #602, opened 2021-07-30
Bug 1988476: remove dhclient binary

miabbott:
We decided back in the RHCOS 4.6 timeframe that we wanted to remove the `dhclient` binary after the switch to using the DHCP client provided by `NetworkManager` in the initrd.  We had customers that were relying on the binary, so we kept it part of RHCOS for a period of time to allow for transition away from its usage.

In the 4.8 release notes [1], we documented the intended removal of the binary and this is the implementation of that removal.

This reverts a number of commits related to `dhclient` and the `dhcp-client` package.  There is one new commit adding `dhcp-client` to `exclude-package` in the manifest due to the new dependency coming from `chrony`.


miabbott:
This builds successfully and I verified that `ext.config.chrony.dhcp-propagation` still passes.  Still running additional tests locally.

@sohankunkerkar would appreciate your review here/

/assign @sohankunkerkar

tests/kola/misc-ro/misc-ro.sh
```diff
@@ -166,3 +161,4 @@ if test -f /usr/lib/systemd/system/console-login-helper-messages-issuegen.servic
   fi
 fi
 echo "ok console-login-helper-messages presets present"
+
```
travier, line 164: Extra line?

travier (commented):
One minor comment but LGTM otherwise

jlebon (approved):
Nice!
/lgtm

PR #603, opened 2021-07-30
manifest: adapt for rpm-ostree modularity support

jlebon:
With the latest rpm-ostree support for modules, rpm-ostree will no
longer pull in modular packages by default, which is something we've
been relying on so far. So if we don't adapt, composes would start
failing. This is a good thing, because it forces us to be more explicit
wrt modularity:
1. enable the `container-tools:rhel8` module for the podman stack
2. enable the `virt:rhel` module for `qemu-guest-agent`

While we're here, use the new `repo-packages` key to clarify some more
of our intentions:
1. always pull in the `kernel` from BaseOS
2. always pull in `nss-altfiles` from AppStream
3. always pull in `toolbox` from RHAOS until we move over to the
   `container-tools:rhel8` one
4. always pull in `cri-o` and `conmon` from RHAOS to ensure they're
   synced

All of the choices above should just be a reflection of our current
stance. I've derived them from the various `exclude=` hacks we have in
our repo files internally, and analyzing where we're currently getting
what from. So this shouldn't significantly change our composes here.
Though I will note that we now drop quite a few `Perl`-related modular
packages, which I don't think we were pulling in intentionally.

This then also allows us to drop all the `exclude=` hacks as well.

Closes: https://github.com/coreos/rpm-ostree/issues/3035

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -261,6 +261,28 @@ packages:
  # https://github.com/openshift/machine-config-operator/pull/2421
  - conntrack-tools
 
+repo-packages:
+  # we always want the kernel from BaseOS
+  - repo: rhel-8-baseos
+    packages:
+      - kernel
+  # we want the one shipping in RHEL, not the equivalently versioned one in RHAOS
+  - repo: rhel-8-appstream
+    packages:
+      - nss-altfiles
+  - repo: rhel-8-server-ose
+    packages:
+      # eventually, we want the one from the container-tools module, but we're
+      # not there yet
+      - toolbox
+
+modules:
+  enable:
+    # podman stack
+    - container-tools:rhel8
+    # qemu-guest-agent
+    - virt:rhel
+
 packages-x86_64:
   # Temporary add of open-vm-tools. Should be removed when containerized
   - open-vm-tools
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -6,6 +6,11 @@ repos:
   - rhel-8-advanced-virt
   - rhel-8-nfv
 
+modules:
+  enable:
+    # for qemu-kiwi
+    - virt:av
+
 extensions:
   # https://github.com/coreos/fedora-coreos-tracker/issues/326
   usbguard:
--- a/rhcos-packages.yaml
+++ b/rhcos-packages.yaml
@@ -16,7 +16,6 @@ packages:
   - nss-altfiles
   - ostree
   - rpm-ostree
-  # part of container-tools module
   - toolbox
   # for kdump:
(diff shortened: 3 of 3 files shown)


jlebon:
Need to hold this until new rpm-ostree release with https://github.com/coreos/rpm-ostree/pull/2760 hits cosa.

jlebon:
> This then also allows us to drop all the `exclude=` hacks as well.

https://url.corp.redhat.com/16481c3

cgwalters (approved):
This looks really good!  Awesome work!

travier (approved):

manifest.yaml
```diff
@@ -259,6 +259,32 @@ packages:
...
+      - toolbox
+      # we want these from RHAOS to make sure they're in sync
+      # https://github.com/openshift/installer/pull/3062#issuecomment-582614009
+      - cri-o
+      - conmon
+
+modules:
+  enable:
+    # podman stack
+    - container-tools:rhel8
```
miabbott, line 284: While I'm glad we are able to select the module we use in the manifest, we should consider the support stance of each stream (i.e. 2.0/3.0/rhel8).  I don't want us to be in a position where we have selected a stream that is not going to be supported for the entirety of the 4.9 lifecycle.

jlebon: Hmm, my understanding is that `rhel8` is the moving target which should be valid for all of el8. It's essentially what we've implicitly been following since its packages are usually the latest NEVRA, which is what rpm-ostree has been picking so far being modularity-unaware. (In fact, [in a few days](https://errata.devel.redhat.com/advisory/79166) we'll be bumping to podman 3.2.3 once the module is updated.) So this is just matching current reality.

But yeah, we should probably get in touch with the containers team on this to clarify things.

miabbott: >  So this is just matching current reality.

That's enough to satisfy my concerns.  Thanks, Jonathan!

manifest.yaml
```diff
@@ -259,6 +259,32 @@ packages:
...
+      - nss-altfiles
+  - repo: rhel-8-server-ose
+    packages:
+      # eventually, we want the one from the container-tools module, but we're
+      # not there yet
+      - toolbox
+      # we want these from RHAOS to make sure they're in sync
+      # https://github.com/openshift/installer/pull/3062#issuecomment-582614009
+      - cri-o
+      - conmon
```
jlebon, line 279: Self-note for when we're ready for this: we can drop this now as per commit message in https://url.corp.redhat.com/8ef7f78.

jlebon:
Updated this, but now requires https://github.com/rpm-software-management/libdnf/pull/1328.

PR #607, opened 2021-08-05
ci: Rework again to aid execution via PRs to coreos-assembler

cgwalters:
In working on https://github.com/openshift/release/pull/20929
I realized the previous PR https://github.com/openshift/os/pull/606
awkward to use.  Plus as jlebon pointed out we should support
branching.

Rework things so that as much shared code as possible lives in `prow-build-test-qemu.sh`
which assumes it's in a cosa workdir with a configured source git.

A key to avoiding duplication is using a conditional for testing
writability of `src/config`, and to do the clone dance internally
rather than externally.

The naming of scripts here is confusing; I created
`ci/prow-thisrepo-entrypoint.sh` which is a symlink to the existing
`build-test-qemu.sh`.  Then we can switch openshift/release over
to using the former, and drop the latter.

### DIFF
--- a/ci/build-test-qemu.sh
+++ b/ci/build-test-qemu.sh
@@ -1,42 +1,10 @@
 #!/bin/bash
 set -xeuo pipefail
-# record information about cosa + rpm-ostree
-if test -d /cosa; then
-    jq . < /cosa/coreos-assembler-git.json
-fi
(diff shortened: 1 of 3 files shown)


miabbott:
Looks reasonable to me!

/lgtm

PR #608, opened 2021-08-18
kola-denylist: enable crio test for all arches

Prashanth684:
With https://github.com/coreos/coreos-assembler/pull/2359 landing and now that the pause image has been building on all arches for a while, enable the crio test for all arches.


jlebon (approved):

cgwalters:
Note right now, we're only running a subset of tests on PRs, so we're not actually testing this change here.

A good thing to do would be to add `/test kola-all` or so.

PR #609, opened 2021-08-25
test: check package versions and downgrades

mike-nguyen:
Add two tests:
- Verify that all packages from the plashets have the correct OCP
  release
- Verify that no packages are downgraded from the previous OCP release

### DIFF
--- a/tests/kola/package/test.sh
+++ b/tests/kola/package/test.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# kola: { "architectures": "!s390x ppc64le", "minMemory": 2048,  "tags": "needs-internet" }
+
+ok() {
+    echo "ok" "$@"
+    exit 0
+}
+
+fatal() {
+  echo "$@"
+  exit 1
+}
+
+# Verify all rhaos packages contain the same OpenShift version
+test_package_versions() {
+  if [[ $(rpm -qa | grep rhaos | grep -v $OPENSHIFT_VERSION) ]]; then
+    fatal "Error: rhaos packages do not match OpenShift version"
+  fi
+}
+
+# Verify there are no downgraded packages
+test_downgraded_packages() {
+  RELEASE=$OPENSHIFT_VERSION
+  STREAM=fast-$RELEASE
+  GRAPH=$(curl -sfH "Accept:application/json" "https://api.openshift.com/api/upgrades_info/v1/graph?channel=$STREAM")
+  if [[ $? -ne 0 ]]; then
+    fatal "Unable to get graph"
+  fi
+
+  # There are no released builds on master so no need to check downgraded packages
+  if [[ $(echo $GRAPH | jq 'has("nodes") and has("edges") and (.nodes | length == 0)') =~ "true" ]]; then
+    ok "No released stream"
+  fi
+
+  # The cincinatti graph defines nodes as a list of objects and edges as a list
+  # of list of two integers. Nodes are releases, and edges are updates.
+  # [x, y] is [from release index, to release index]. The release index can
+  # change every request for the cincinatti graph!
+  #
+  # Use jq to find the node that contains no "from release index" edge.
+  #   1. Find all the unique "from release indexes"
+  #   2. Find all the release indexes
+  #   3. Get the difference between all the releases and the "from release
+  #      indexes".  This is the latest because there is no update from this
+  #      release.
+  PAYLOAD=$(echo $GRAPH | jq -r '. as $graph | [$graph.edges[][0]] | unique as $from | $graph.nodes | to_entries as $indexed | [$indexed[].key] | unique as $nodes | ($nodes - $from)[] as $latest | $indexed[] | select(.key == $latest) | .value.payload')
+  VERSION=$(oc adm release info $PAYLOAD -o json | jq -r '.displayVersions."machine-os".Version')
+  OCP_COMMIT=$(curl -sSL https://art-rhcos-ci.s3.amazonaws.com/releases/rhcos-$RELEASE/$VERSION/x86_64/meta.json | jq -r '."ostree-commit"')
+
+  curl -SL https://art-rhcos-ci.s3.amazonaws.com/releases/rhcos-$RELEASE/$VERSION/x86_64/rhcos-$VERSION-ostree.x86_64.tar -o $STREAM.tar
+
+  mkdir -p repo && tar xvf $STREAM.tar -C $_ && rm -rf $STREAM.tar
+
+  ostree pull-local repo
+
+  RHCOS_COMMIT=$(rpm-ostree status --json | jq -r .deployments[0].checksum)
+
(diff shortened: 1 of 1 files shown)


tests/kola/package/test.sh
```diff
@@ -0,0 +1,74 @@
...
+  #   1. Find all the unique "from release indexes"
+  #   2. Find all the release indexes
+  #   3. Get the difference between all the releases and the "from release
+  #      indexes".  This is the latest because there is no update from this
+  #      release.
+  PAYLOAD=$(echo $GRAPH | jq -r '. as $graph | [$graph.edges[][0]] | unique as $from | $graph.nodes | to_entries as $indexed | [$indexed[].key] | unique as $nodes | ($nodes - $from)[] as $latest | $indexed[] | select(.key == $latest) | .value.payload')
+  VERSION=$(oc adm release info $PAYLOAD -o json | jq -r '.displayVersions."machine-os".Version')
+  OCP_COMMIT=$(curl -sSL https://art-rhcos-ci.s3.amazonaws.com/releases/rhcos-$RELEASE/$VERSION/x86_64/meta.json | jq -r '."ostree-commit"')
+
+  curl -SL https://art-rhcos-ci.s3.amazonaws.com/releases/rhcos-$RELEASE/$VERSION/x86_64/rhcos-$VERSION-ostree.x86_64.tar -o $STREAM.tar
```
cgwalters, line 52: This will break after we try again with https://github.com/openshift/os/pull/593

I think it'd be a bit cleaner to download the `meta.json` and then do `jq .images.ostree.path` on it, then we can conditionalize on whether it ends in `.ociarchive` or `.tar` as is done in e.g.
https://github.com/coreos/coreos-assembler/blob/88efaff63b38542763336e12249b11f53d606776/src/cosalib/cmdlib.py#L251

Also avoid duplicating the URL would be good.

tests/kola/package/test.sh
```diff
@@ -0,0 +1,74 @@
...
+  # [x, y] is [from release index, to release index]. The release index can
+  # change every request for the cincinatti graph!
+  #
+  # Use jq to find the node that contains no "from release index" edge.
+  #   1. Find all the unique "from release indexes"
+  #   2. Find all the release indexes
+  #   3. Get the difference between all the releases and the "from release
+  #      indexes".  This is the latest because there is no update from this
+  #      release.
+  PAYLOAD=$(echo $GRAPH | jq -r '. as $graph | [$graph.edges[][0]] | unique as $from | $graph.nodes | to_entries as $indexed | [$indexed[].key] | unique as $nodes | ($nodes - $from)[] as $latest | $indexed[] | select(.key == $latest) | .value.payload')
```
cgwalters, line 48: This is some veritable jq sorcery.

tests/kola/package/test.sh
```diff
@@ -0,0 +1,74 @@
...
+
+  curl -SL https://art-rhcos-ci.s3.amazonaws.com/releases/rhcos-$RELEASE/$VERSION/x86_64/rhcos-$VERSION-ostree.x86_64.tar -o $STREAM.tar
+
+  mkdir -p repo && tar xvf $STREAM.tar -C $_ && rm -rf $STREAM.tar
+
+  ostree pull-local repo
+
+  RHCOS_COMMIT=$(rpm-ostree status --json | jq -r .deployments[0].checksum)
+
+  if [[ $(rpm-ostree db diff $OCP_COMMIT $RHCOS_COMMIT | grep -A1000 Downgraded) ]]; then
```
cgwalters, line 60: Hm, we should add something like `db diff --downgrades` or something so this test could just validate it's empty.

cgwalters (commented):
One thing I think we should bear in mind as a general rule is that our team can and IMO should also add tests to the main OpenShift tests, not just our kola tests.   For example I did one a while ago in https://github.com/openshift/origin/pull/25993  (That one was specifically about in-cluster stuff so it *needed* to be there of course)

I think this one would also be a good potential candidate to live in the openshift e2e tests, because it isn't actually testing the operating system itself, just metadata about the build.  It's also tied to OCP and the release image.

While RHEL currently does not revert packages unfortunately, to me the ability to do so is a key selling point of image based updates.  I guess if we explicitly wanted to revert, we could just denylist this test temporarily or something?

We also had a similar debate in FCOS over here: https://github.com/coreos/fedora-coreos-config/issues/841

tests/kola/package/test.sh
mike-nguyen: @cgwalters it looks like the version of `rpm-ostree` in the latest rhcos does not have the `ex-container` functionality.  I see  that you converted the oci-archive to tar inside cosa before and copied it to the test system [0] but this is an external test, and I don't think that's possible.  Do you have any ideas to work around this?

Some ideas:
- create a ostree container to handle the conversion 
- rewrite test as a mantle kola test and do the same thing as [0]

I also saw that in the future the container may contain a http server.  Would anything need to change in the oci-archive or would it just ab an additional http server to serve the ostree repo?

[0] https://github.com/coreos/coreos-assembler/blob/350c1d93bc3343c8f523d642664bccd8dc0424c9/mantle/kola/tests/rhcos/upgrade.go#L95

cgwalters: There's more alternatives:

 - Add `cosa build --disallow-downgrades` and use it in the pipeline by default (and have a declarative file of allowed downgrades?)
 - Add the above, but implement in rpm-ostree (implicity adding `allowed_downgrades: []` i.e. empty YAML list would disallow downgrades, then the list is an allowlist)

cgwalters: Either way, this is basically the same thing as https://github.com/coreos/fedora-coreos-config/pull/892 right?  So if we deduplicate in either coreos-assembler or rpm-ostree it would be a win (less pipeline code, more easily reproduced outside of pipelines, etc.)

I kind of like the `allowed-downgrades` in rpm-ostree natively but it's also pretty easy to implement outside it in a cosa workdir which is what the FCOS pipeline PR above is doing.

mike-nguyen: [jlebon's comment](https://github.com/coreos/fedora-coreos-config/pull/892#discussion_r594648764) may apply here because it would be checking from the last build and not the last release. Maybe this test should be scrapped and changed to an upgrade test from the previous releases and db diff  instead of just doing a db diff.

jlebon: I think going the `cosa build --disallow-downgrades` + YAML path could work. There'd be overhead involved, but maybe that's OK. I do like that with https://github.com/coreos/fedora-coreos-config/pull/892, it's just an easy "rubber-stamping" thing: we don't have to do anything else than just force merge the PR. In the pipeline itself, nothing cares about downgrades.

For RHCOS, that's harder to do without lockfiles because unexpected downgrades could happen anytime up to and including during the prod pipeline build. So it's much harder to make it as lightweight as in FCOS.

mike-nguyen:
closing in favor of  https://github.com/openshift/os/pull/635 and https://github.com/coreos/coreos-assembler/pull/2459

PR #611, opened 2021-08-30

cgwalters:
This is a re-submit; we reverted in https://github.com/openshift/os/pull/601
but the issue is fixed with newer rpm-ostree.

let's settle for parallel.

Some discussion on that here:

arounnd and make `podman run registry.ci.openshift.org/rhcos/rhcos:4.8` e.g.


miabbott (approved):
Since this doesn't interfere with the current ability to upload the oscontainer, this seems worth another try.

/lgtm

cgwalters:
reverted in https://github.com/openshift/os/pull/613

PR #612, opened 2021-09-01
ci: Run upgrade tests and all kola tests (--parallel 2)

cgwalters:
Turns out the `image-format: oci` change broke the rhcos upgrade
test, let's add coverage of that.

While we have the patient open, also run all kola tests.


cgwalters:
Cool, this correctly failed:
```
 --- FAIL: rhcos.upgrade.luks/setup (16.66s)
cluster.go:143: error: opening repo: opendir(objects): No such file or directory 
```

And then once https://github.com/openshift/os/pull/613 merges we can rebase this and verify it passes.

miabbott (approved):
/lgtm

PR #616, opened 2021-09-03
manifest: force container-selinux from OSE repo

miabbott:
We tagged a new version of `container-selinux` for inclusion in RHCOS
4.9 in order to fix RHBZ#1999245. So we have to specify that we want
it to come from the `rhel-8-ose-server` repo instead of the normal
RHEL repo.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -275,6 +275,10 @@ repo-packages:
       # eventually, we want the one from the container-tools module, but we're
       # not there yet
       - toolbox
+      # we tagged a new version of container-selinux so that we can include the fix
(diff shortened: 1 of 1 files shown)


saqibali-2k (approved):

travier (approved):
LGTM. We should probably set a reminder somewhere about that.

miabbott:
> LGTM. We should probably set a reminder somewhere about that.

I'll create a reminder on our Slack channel

ravanelli (approved):
LGTM

miabbott:
Thanks for the reviews everyone; merging!

Issue #615, opened 2021-09-02
`multipath.day1` is flaking


miabbott:
s390x was failing because we needed to bump `coreos-installer` - https://github.com/openshift/os/pull/614

...but a new version of `coreos-installer` is on its way.

Now we are seeing the test fail in x86_64:

```
--- FAIL: multipath.day1 (83.92s)
        multipath.go:125: mount /boot has non-multipath source /dev/disk/by-label/dm-mpath-boot
```

Nothing jumping out in the console/journal, though.

jschintag:
`s390x` is still failling, now with a new error
```
--- FAIL: multipath.day1 (21.48s)
        harness.go:1120: Cluster failed starting machines: machine 42262dc0-142a-40d0-bc12-9674c668d4fa entered emergency.target in initramfs
```
Need to investigate, we are seing this message in fcos as well.

jschintag:
Ok, investigating further, s390x is still waiting for the fix for https://bugzilla.redhat.com/show_bug.cgi?id=1974411 to land.

miabbott:
Wasn't able to get the kola archive for the x86_64 failure, but I ran it in a loop locally for 50 iterations (`kola run multipath.day1 --multiply 50`)and wasn't able to reproduce the failure.

miabbott:
[kola_artifacts.zip](https://github.com/openshift/os/files/7106951/kola_artifacts.zip)

kola aritifacts from a failing s390x run

miabbott:
snippet from console.txt from the above failed s390x run

```
[    8.471687] ignition[780]: Ignition 2.12.0                                                                                                                  
[    8.471738] ignition[780]: Stage: kargs                                                                                                                     
[    8.471829] ignition[780]: reading system config file "/usr/lib/ignition/base.d/00-core.ign"
[  OK  ] Started dracut pre-mount hook.                                                                                                                        
[    8.473662] systemd[1]: Started dracut pre-mount hook.                                                                                                      
[    8.477311] ignition[780]: no config dir at "/usr/lib/ignition/base.platform.d/qemu"                                                                        
[    8.477918] ignition[780]: kargs: op(1): [started]  updating kernel arguments                                                                               
[    8.645201] EXT4-fs (dm-1): mounted filesystem with ordered data mode. Opts: (null)
[...]

PR #617, opened 2021-09-03
kola-denylist: exclude multipath.day1 temporarily on s390x

Prashanth684:
multipath.day1 test is failing on s390x because coreos-installer has not been updated to the latest to include the fix to run zipl (https://github.com/coreos/coreos-installer/pull/533)
@jschintag has opened https://src.osci.redhat.com/rpms/coreos-installer/pull-request/28 which updates coreos-installer to 0.10.0, but in the meanwhile, exclude the test so pipeline builds are not blocked.

Update: Now this test is blocked by https://bugzilla.redhat.com/show_bug.cgi?id=1974411

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -12,3 +12,7 @@
    - s390x
 - pattern: coreos.ignition.journald-log
   tracker: https://github.com/coreos/coreos-assembler/issues/1173
+- pattern: multipath.day1
+  tracker: https://github.com/openshift/os/issues/615
+  arches:
+   - s390x


miabbott:
@Prashanth684 the fix for 1974411 is in https://github.com/coreos/fedora-coreos-config/pull/1077 which should be part of RHCOS by now (I confirmed it is part of the x86_64 builds), so this test should be passing, right?

Could you build/boot RHCOS for s390x and verify that coreos/fedora-coreos-config#1077 is included?

Prashanth684:
> @Prashanth684 the fix for 1974411 is in [coreos/fedora-coreos-config#1077](https://github.com/coreos/fedora-coreos-config/pull/1077) which should be part of RHCOS by now (I confirmed it is part of the x86_64 builds), so this test should be passing, right?
> 
> Could you build/boot RHCOS for s390x and verify that [coreos/fedora-coreos-config#1077](https://github.com/coreos/fedora-coreos-config/pull/1077) is included?

looks like it @miabbott . it doesn't seem to have to do with 1974411 but with zipl failing:
```
"Error: Unsupported device driver 'device-mapper'\nError: \"zipl\" \"--blsdir\" \"/tmp/coreos-installer-zipl-bls-t4Fn5L/loader/entries\" \"--config\" \"/tmp/coreos-installer-zipl.Khpfzo\" failed with exit code: 1\n"
```

Prashanth684:
@miabbott while the discussion happens here: https://github.com/openshift/os/issues/615 can we get this merged so the pipelines can continue building ? @jschintag we know that day1 multipath works on baremetal right? it's only qemu that's  a problem ?

PR #618, opened 2021-09-03
40rhcos-fips: drop redundant zipl run with coreos-installer 0.10.0

bgilbert:
cc @jlebon @nikita-dubrovskii


jlebon (approved):
That's a nice cleanup. Hopefully in the future we can simplify this even more by having the MCO drive the FIPS kargs directly via Ignition. Then here we'll just need the `fips-mode-setup` bits.

PR #619, opened 2021-09-08

cgwalters:
This is the 3rd attempt at this, trying again now that `master` is 4.10 and https://github.com/coreos/coreos-assembler/pull/2403 merged.

let's settle for parallel.

Some discussion on that here:

arounnd and make `podman run registry.ci.openshift.org/rhcos/rhcos:4.8` e.g.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -11,3 +11,5 @@ squashfs-compression: gzip
 # Disable networking by default on firstboot. We can drop this once cosa stops
 # defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
+
+ostree-format: "oci"


cgwalters:
```
 --- FAIL: coreos.ignition.systemd.enable-service (61.19s)
        cluster.go:206: cmd systemctl status nfs-server.service did not output inactive 

```

Exciting, this must be a regression from https://github.com/coreos/coreos-assembler/pull/2418

cgwalters:
https://github.com/coreos/coreos-assembler/pull/2419

miabbott:
`Cannot download Packages/container-selinux-2.167.0-1.module%2bel8.5.0%2b12397%2bbf23b712__noarch/container-selinux-2.167.0-1.module%2bel8.5.0%2b12397%2bbf23b712.noarch.rpm`

This is a result of #616 and it looks like that package isn't tagged for 4.10; I'll have to internally request it is tagged again

cgwalters:
OK let's give this another try?

cgwalters:
/hold
for 4.9 branching

cgwalters:
/hold cancel
since I think we're already tracking `release-4.9` and this can go in master.

PR #621, opened 2021-09-10
Revert "manifest: force container-selinux from OSE repo"

miabbott:
This reverts commit cb6bfa845cc9d0d8ed01323a7719d118e67515c1.

This should allow us to build RHCOS from `master` (i.e. 4.10), but
causes us to lose the fix for RHBZ#1999245.

The version of `container-selinux` that includes the fix for the BZ
has been requested to be tagged for 4.10, so we can revert this revert
once that happens.


bgilbert (approved):
LGTM

PR #622, opened 2021-09-13
Bug 2003743: Revert "overaly: Keep /boot RW when kdump is enabled"

travier:
We can now revert this workaround now that the following bugs are fixed:
- https://bugzilla.redhat.com/show_bug.cgi?id=1976252
- https://bugzilla.redhat.com/show_bug.cgi?id=1976260

This reverts commit 6a71f474eb795c774f91df259a92b887f8819277.

### DIFF
--- a/overlay.d/12kdump
+++ b/overlay.d/12kdump
@@ -1,1 +1,0 @@
-../fedora-coreos-config/overlay.d/12kdump
--- a/tests/kola/kdump/test.sh
+++ b/tests/kola/kdump/test.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+set -xeuo pipefail
+# https://docs.fedoraproject.org/en-US/fedora-coreos/debugging-kernel-crashes/
+# Only run on QEMU for now:
+# https://github.com/coreos/fedora-coreos-tracker/issues/860
+# kola: {"platforms": "qemu-unpriv", "minMemory": 4096, "tags": "skip-base-checks"}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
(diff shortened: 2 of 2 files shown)


travier:
See also: https://github.com/coreos/fedora-coreos-config/pull/1212

This will need manual testing (and then backports) on RHCOS 4.7+.

miabbott:
> See also: [coreos/fedora-coreos-config#1212](https://github.com/coreos/fedora-coreos-config/pull/1212)
> 
> This will need manual testing (and then backports) on RHCOS 4.7+.

Friendly reminder that we will need BZs for any backports

miabbott (approved):
I confirmed that the recent RHCOS 4.9 builds have `kexec-tools-2.0.20-46.el8_4.2` and `selinux-policy-3.14.3-67.el8_4.1` so this should be safe to merge.

travier:
@miabbott Can you set the target release & other fields in the BZ? Getting it in 4.9+ would be the minimum and 4.7+ if possible depending on the impact on update (will need testing). Thanks!

travier:
> Is the test the same between here and f-c-c? You can now simply symlink it so that the source of truth stays in f-c-c.

I went that route first but unfortunately the test is hard disabled directly in f-c-c right now and not using a kola-denylist entry.
We can also not enable it on FCOS for now until kexec-tools is fixed.

travier:
See https://github.com/coreos/fedora-coreos-config/pull/1212/files#diff-d13fe2004e2f7382d2d603a5e5790445a80fef844471b47a1f7ecad2243fcd57

miabbott:
Tests passing locally and in CI  :tada: 

/lgtm

PR #629, opened 2021-09-17
FAQ: Note that we expose some key package versions via metadata

cgwalters:
Motivated by private chat asking about the kernel version.


saqibali-2k (approved):

miabbott (approved):
/lgtm

PR #632, opened 2021-09-22
Bug 2003743: tests/kdump: Only run on x86_64 & aarch64

travier:
The test is currently failing on ppc64le & s390x. Limit to x86_64 &
aarch64 only to unblock the pipelines while we figure out the issue.


miabbott (approved):
/lgtm

PR #635, opened 2021-09-23
test: add test to check rhaos package versions

mike-nguyen:
Add test to verify that rhaos package versions match the OpenShift
version.

### DIFF
--- a/tests/kola/misc-ro/misc-ro.sh
+++ b/tests/kola/misc-ro/misc-ro.sh
@@ -161,3 +161,9 @@ if test -f /usr/lib/systemd/system/console-login-helper-messages-issuegen.servic
   fi
 fi
 echo "ok console-login-helper-messages presets present"
+
+# Check that rhaos packages do not match the OpenShift version
+if [[ $(rpm -qa | grep rhaos | grep -v $OPENSHIFT_VERSION) ]]; then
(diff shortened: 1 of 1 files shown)


miabbott:
```Sep 23 17:52:47 qemu0 kola-runext-misc-ro.sh[1600]: + [[ -n cri-o-1.21.3-2.rhaos4.8.gitff0b7fe.el8.x86_64 ]]
Sep 23 17:52:47 qemu0 kola-runext-misc-ro.sh[1600]: + fatal 'Error: rhaos packages do not match OpenShift version'
Sep 23 17:52:47 qemu0 kola-runext-misc-ro.sh[1600]: + echo 'Error: rhaos packages do not match OpenShift version'
```

Already providing value!  `cri-o` should be 1.22 for OCP 4.10

miabbott:
`cri-o-1.22.0-67.rhaos4.10.git6becad2` landed in the downstream plashets; let's see if it is on the mirrors

/retest

cgwalters:
This also seems like something again we could enforce at *build* time, actually it's really something ART should probably be enforcing on their side I think.  That all said, clearly this is still a simple and cheap check, so it makes sense to me to ship.  Thanks for doing this!

/lgtm

PR #641, opened 2021-09-30
Bug 2009465: Revert "Revert "Revert "manifest: force container-selinux from OSE repo"""

miabbott:
There was an async release of the `container-tools` module that has
the necessary `container-selinux` RPM we want, so we can start
consuming this from RHEL proper.

https://access.redhat.com/errata/RHBA-2021:3661

This reverts commit d5bb0ecb4e565514869575d6d69ce5bb23007c3c.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -275,10 +275,6 @@ repo-packages:
       # eventually, we want the one from the container-tools module, but we're
       # not there yet
       - toolbox
-      # we tagged a new version of container-selinux so that we can include the fix
-      # for RHBZ#1999245 in RHCOS 4.9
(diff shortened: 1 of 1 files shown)


miabbott:
Ugh, this needs a backport to 4.9...

_/me cries in backport BZs_

miabbott:
The `container-selinux` that we were using from the RHAOS repo has been untagged, so this is preventing downstream builds from succeeding.

cgwalters:
/lgtm

I think there's a source-of-truth problem here - we are trying to keep the rpm-md repos and this manifest in sync, but it'd be better for one to be truth, and the other follow that.

I think probably the best fix really is lockfiles, which will require ART to maintain multiple versions in the rpm repos.

PR #643, opened 2021-10-01
Bug 2009840: extensions: make virt:av module and advanced virt repo specific to kata

Prashanth684:
these are not needed for P/Z/ARM. With https://github.com/coreos/rpm-ostree/pull/3152 landing, the repo and module can be made specific to the extension

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -3,13 +3,7 @@
 # and https://github.com/coreos/fedora-coreos-tracker/issues/401
 
 repos:
-  - rhel-8-advanced-virt
   - rhel-8-nfv
-
-modules:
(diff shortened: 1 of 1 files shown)


Prashanth684:
hmm..the test is still using the old coreos-assembler commit, not the latest

miabbott:
> hmm..the test is still using the old coreos-assembler commit, not the latest

I think I was a little quick on the trigger...the Quay rebuild didn't complete until about 30 min ago.  Should be enough time now to get it updated in Prow.  Let's hope :crossed_fingers: 

/retest

miabbott (approved):
/lgtm

PR #652, opened 2021-10-14
manifest: pull runc from the OCP repo

miabbott:
The current version of `runc` getting pulled into RHCOS is an older
version coming from the `rhel-8-appstream` repo. We want to start
using correctly versioned packages from the `container-tools` stack
and this is the first step.

See: https://bugzilla.redhat.com/show_bug.cgi?id=2013822

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -274,6 +274,9 @@ repo-packages:
       # eventually, we want the one from the container-tools module, but we're
       # not there yet
       - toolbox
+      # we are going to start pulling the whole container-tools stack from the
+      # OCP repo
+      - runc
 
 modules:
   enable:


miabbott:
The OSE repo has an up-to-date version of `runc` available, so the CI jobs should start passing again.

Side-effect of this change is that we get `crun` pulled in as well (as desired by the Container Tools team).

cgwalters:
/lgtm
But it still really seems to me that forking the container module by default for OCP is a long term bad strategy.
I mean, it already has to be a large burden on the container team to have multiple (traditional) RHEL streams for !OCP cases, on top of an OCP version.  I guess the flip side is we don't actually really *use* `podman` except for bootstrapping, just `runc`.

PR #655, opened 2021-10-19
Bug 1990506: Add 63-scsi-sg3_symlink.rules for udev

ravanelli:
- BZ1990506: Add missing symlinks at boot time


travier:
Is this only for RHCOS and not FCOS too? Anything we should push upstream?

PR #657, opened 2021-10-19
COS-2611: Use deploy-via-container

cgwalters:
Use `deploy-via-container`

This will cause us to run through the ostree-native container
stack when generating the disk images.

Today for RHCOS we're using the "custom origin" stuff which
lets us inject metadata about the built source, but rpm-ostree
doesn't understand it.

With this in the future (particularly after https://github.com/coreos/coreos-assembler/issues/2685)
`rpm-ostree status` will show the booted container and *understand it*;
for example in theory we could have `rpm-ostree upgrade` work (if
we provisioned from a tag instead of a digest).

A side benefit is this should fix https://github.com/coreos/coreos-assembler/issues/2685
because we're now pulling a single large file over 9p instead of lots
of little ones.

---

### DIFF
--- a/image-c9s.yaml
+++ b/image-c9s.yaml
@@ -1,6 +1,9 @@
 # See https://github.com/coreos/coreos-assembler/pull/298
 size: 16
 
+# We default to a container image
+deploy-via-container: true
+
 # Disable networking by default on firstboot. We can drop this once cosa stops
 # defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
--- a/image-rhel-9.2.yaml
+++ b/image-rhel-9.2.yaml
@@ -1,6 +1,9 @@
 # See https://github.com/coreos/coreos-assembler/pull/298
 size: 16
 
+# We default to a container image
+deploy-via-container: true
+
 # Disable networking by default on firstboot. We can drop this once cosa stops
 # defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []


cgwalters:
# Experimenting with ostree-containers in OCP/RHCOS8

First, https://copr.fedorainfracloud.org/coprs/g/CoreOS/experimental/ contains binary RPMs for RHEL8.

Using that with this PR, I've run `coreos-assembler build`:

New format container: `registry.ci.openshift.org/coreos/walters-rhcos-ostreecontainer` - you can use this container via e.g. `FROM registry.ci.openshift.org/coreos/walters-rhcos-ostreecontainer` in a `Dockerfile`

Old (current) format container (archive ostree + extensions RPMs): `registry.ci.openshift.org/coreos/walters-rhcos-ostreecontainer-oldformat:410.84.202110191506-0`

OCP release image: `registry.ci.openshift.org/coreos/walters-ocp410-ostreecontainer:4.10.0-0.nightly-2021-10-16-173656`

cgwalters:
OK moving this to https://github.com/ostreedev/ostree-rs-ext/issues/121

cgwalters:
OK https://copr.fedorainfracloud.org/coprs/g/CoreOS/rpm-ostree-rhel8 is updated with the latest skopeo+rpm-ostree changes.  Previously it was trying to target both Fedora and RHEL8 (via C8S) but since the code is in Fedora I've now changed it to just target RHEL8.  But even doing that suddenly required switching from c8s to epel8 because https://git.centos.org/rpms/json-c/c/aa505d489ccc4ad2e2abfcc61b08b8f8b272c4f4?branch=c8s broke binary backcompat (and libdnf links to json-c).

I also made a few fixes to ensure the stack runs on rhel8, which are rolled in to https://github.com/coreos/rpm-ostree/pull/3249

Next step: update release image etc.

cgwalters:
Ahhhhhh....I finally figured it out:

```
sh-4.4# cat /etc/hosts
# Kubernetes-managed hosts file (host network).
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
172.30.122.47 image-registry.openshift-image-registry.svc image-registry.openshift-image-registry.svc.cluster.local # openshift-generated-node-resolver
```

Something is special casing the registry in `/etc/hosts` - and of course the `host image-registry.openshift-image-registry.svc` command only talks directly to the DNS resolver and so bypasses `/etc/hosts`.  I love that the name of the command is `host` and it doesn't read `/etc/hosts`!

This misled me to somehow think some magic with network namespaces was involved, because I then didn't even try using e.g. `curl` or `podman/oc` to interact with the registry.  I need to retrain my fingers to use `getent hosts` instead.  But it's more typing.

cgwalters:
OK now I'm banging my head against image pull authentication again.  I would swear previously I did

```
export KUBECONFIG=/etc/kubernetes/kubeconfig
oc registry login
```

as root to get a `~/.docker/config.json`, but that isn't working now.

What we really need to drill into here is basically re-using the kubelet credentials I think.

cheesesashimi:
@cgwalters: I most likely misunderstand the context in which you're trying to log into podman, but I was able to pull from the in-cluster registry using the service account credentials for a given namespace (which have pull access to the in-cluster registry by default) by doing the following:

1. Get the image pull credentials for the default service account for a given namespace. We're only interested in the `auth` field, so we have to do a bit of JSON manipulation to just get that field:
```console
$ oc get secret/default-dockercfg-kz5l9 -o json | jq -r '.data[".dockercfg"]' | base64 -d | jq '{"auths": {"image-registry.openshift-image-registry.svc:5000": {"auth": .["image-registry.openshift-image-registry.svc:5000"].auth}}}' > dockercfg.json
```

2. Next, I dropped the resulting config file onto one of the nodes by taking advantage of the fact that the MCD has the root filesystem of a node mounted under `/rootfs`:

```console
$ oc cp -n openshift-machine-config-operator -c machine-config-daemon ./dockercfg.json machine-config-daemon-92bnb:/rootfs/etc/dockercfg.json
```

3. Then, after `oc debug node/nodename` / `chroot /host` into the node I copied the `dockercfg.json` file to:
```console
$ podman pull --authfile=/etc/dockercfg.json 'image-registry.openshift-image-registry.svc:5000/fedora-coreos/my-rhcos:latest'
Trying to pull image-registry.openshift-image-registry.svc:5000/fedora-coreos/my-rhcos:latest...
Getting image source signatures
Copying blob ffd9d46effa0 done
Copying blob a618bdc8dbbe done
Copying blob ed2619ec1e46 done
Copying blob c4f6085c7f89 done
Copying blob ac7d3d7061f7 done
Copying blob 40aed5bb0fab done
Copying config 57b268ba3d done
Writing manifest to image destination
Storing signatures
57b268ba3dda75018ef3449f2d257c911eb8e47ba33babebf70b8f205e16366d
```

I'm not sure if this solves the issue at hand, but I think it may be a step in the right direction.

cgwalters:
Right yep!  Sorry this PR discussion history is a bit of a mess; basically the in-cluster pull thread moved to https://github.com/ostreedev/ostree-rs-ext/issues/121 and then @mkenigs prototyped it out in https://github.com/openshift/machine-config-operator/pull/2886 which actually seems to be passing now :tada:

cheesesashimi:
Ahhh. That makes sense now.

cgwalters:
OK, since we have rpm-ostree v2022.2 now, this PR drops the `microdnf` addition.

cgwalters:
Lifting draft on this; I think we could merge this.  It's not in the blocking path, but it is where we want to go.

One small downside is this will cause a space leak across upgrades until https://github.com/ostreedev/ostree-rs-ext/issues/146 is fixed, but we will be fixing that.

jmarrero (approved):

cgwalters:
Rebased :surfer: for the many changes in the last...wow, has it really been almost *two years* (OK 1.75)...

jmarrero (approved):
/lgtm
/approve

cgwalters:
https://github.com/coreos/coreos-assembler/pull/3573

cgwalters:
OK, we have a green for RHEL now.  However, this one could still use some more manual testing
/hold
because I've realized now it may intersect with the bug surfaced by https://github.com/openshift/machine-config-operator/pull/3857
Additionally, it just looks...well, worse unfortuantely until we finally publish official pull specs.

cgwalters:
> because I've realized now it may intersect with the bug surfaced by https://github.com/openshift/machine-config-operator/pull/3857

I was asked about the status of this issue; I think it's likely to be fixed in the MCO now, but the simplest way to test is basically:

- launch or access a test worker node (or a whole cluster if that's easier)
- `skopeo copy docker://<pull spec> oci:/var/tmp/someoci`
- `rpm-ostree rebase ostree-unverified-image:oci:/var/tmp/someocidir`
- Kill and restart the MCD pod and see how it errors out

If it errors out with just "unexpected osimageurl", that's basically OK.  If it errors in a more fatal way, then the osimageurl parsing still needs work most likely.

The most real test is to scale up a new worker node (or a whole cluster) from a disk image with this change.

> Additionally, it just looks...well, worse unfortuantely until we finally publish official pull specs.

jmarrero:
- I built an rhcos image with this change, pushed it to quay.
- I started a cluster with clusterbot on gcp
- ran the above steps, rebooted the node and killed all machine-config-daemon pods.
- The MachineConfigPool details shows:

```
NodeDegraded
True
Jan 17, 2024, 1:49 PM
1 nodes are reporting degraded status on sync
Node ci-ln-1xk0c2t-72292-9lww5-worker-a-2vc5k is reporting: "unexpected on-disk state validating against rendered-worker-7e34c96a1c0e79f4e4ffd254c0d167cd: expected target osImageURL \"quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:fa632104c396a0dae5ac25814f791eb8650cd862bcb37dbe9eab107c30f56699\", have \"/var/tmp/rhcosoci\" (\"82aebaf74dfbbb748e4d4367c8cba7bb99d9c20537d03ec48681a1f7a0179c92\")"
```

cgwalters:
Yeah that's the error I'd expect; which is generally good.  It means we didn't fail to parse it at least.

cgwalters:
> I built an rhcos image with this change, pushed it to quay.

Note this change has no effect really on the generated container image, only on disk images.

jmarrero:
@cgwalters any reason not to let CI run on this and merge if all looks good then?

I guess the pull spec looking as a directory is not that awful. Not sure if we are tracking official pull specs anywhere else.

cgwalters:
Unfortunately CI in this repository still doesn't cover running OpenShift, which is where this change would matter in general.  Most of our kola unit tests don't care.

dustymabe:
/retest

> The most real test is to scale up a new worker node (or a whole cluster) from a disk image with this change.

so this is what still needs to be tested I think. It's been a while since I've done this but @jmarrero I think you can you create an image in GCP (or an AMI in AWS if that's easier) and then modify the machine pool of a test cluster in that cloud to specify a different AMI ID or GCP image name for workers to start, then scale it up.

jmarrero:
Tested on a cluster by:
1) uploading the ami:
```
cosa buildextend-aws --upload --credentials-file=awscreds --bucket=s3://mybucket --region=us-east-1 --grant-user=####
```
2) On a cluster created by cluster bot I modify the workers machineset to use the ami I just uploaded.
3) Delete one of the machines that are part of the machine.
4) Verify that the new machine is using the ami I uploaded and becomes healthy.

dustymabe:
Thanks @jmarrero - also https://github.com/coreos/coreos-assembler/pull/3700 just merged which should give us a less hacky looking imgref when running `rpm-ostree status`.

I'd say this is unblocked now?

jmarrero (approved):

PR #668, opened 2021-11-11
Force container-selinux from OCP repo

yselkowitz:


ravanelli (approved):
/lgtm

I just tested it, I confirm I can see the  `container-selinux-2.170.0-2.rhaos4.10.el8.noarch` package now

PR #670, opened 2021-11-15
manifest.yaml: Add content_manifest folder to add <nvr>.json for build info

gursewak1997:
Add content_manifests dir to have json files for build_info

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -164,6 +164,29 @@ postprocess:
      # NB: we don't use -f here so we break when this is no longer needed
      rm -v /etc/iscsi/initiatorname.iscsi
 
+  # Add content_manifests dir to have json files for build info
+  # See https://issues.redhat.com/browse/GRPA-3731
+  - |
+     #!/usr/bin/env bash
+     set -euo pipefail
+     
+     REPOS=`jq .repos /usr/share/rpm-ostree/treefile.json`
+     ARCH=`jq .basearch /usr/share/rpm-ostree/treefile.json | tr -d '"'`
+     
+     . /etc/os-release
+     mkdir -p /usr/share/buildinfo/content_manifests && cd "$_" && touch $ID-$OSTREE_VERSION.json
+     # Make an empty json and add all the relevant fields
+     echo "$(var=$REPOS; jq --argjson repos "$var" '. += 
+      {
+       "metadata": {
+         "icm_version": 1,
+         "icm_spec": "https://raw.githubusercontent.com/containerbuildsystem/atomic-reactor/master/atomic_reactor/schemas/content_manifest.json",
+         "image_layer_index": 1
+        },
+        "content_sets": $repos,
+        "image_contents": []
+      }' $ID-$OSTREE_VERSION.json)" > $ID-$OSTREE_VERSION.json
+
 etc-group-members:
   - wheel
   - sudo
--- a/overlay.d/05rhcos/usr/lib/tmpfiles.d/buildinfo.conf
+++ b/overlay.d/05rhcos/usr/lib/tmpfiles.d/buildinfo.conf
@@ -1,0 +1,10 @@
+#  This file is part of systemd.
+#
+#  systemd is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+# See tmpfiles.d(5) for details
+
+L /var/roothome/buildinfo - - - - /usr/share/buildinfo


manifest.yaml
```diff
@@ -164,6 +164,18 @@ postprocess:
...
+  # See https://issues.redhat.com/browse/GRPA-3731
+  - |
+     #!/usr/bin/env bash
+     REPOS=`jq .repos /usr/share/rpm-ostree/treefile.json`
+     OSTREE_VERSION=$(grep -oP '(?<=^OSTREE_VERSION=).+' /etc/os-release | tr -d "'")
+     ARCH=`jq .basearch /usr/share/rpm-ostree/treefile.json | tr -d '"'`
+     RHEL_VERSION=$(grep -oP '(?<=^RHEL_VERSION=).+' /etc/os-release | tr -d '"')
+     OS_NAME=$(grep -oP '(?<=^ID=).+' /etc/os-release | tr -d '"')
+     
+     mkdir -p /var/roothome/buildinfo/content_manifests && cd "$_" && touch $OS_NAME-$OSTREE_VERSION.json
```
jlebon, line 177: We can't write to `/var` at compose time. Maybe `/usr/share/buildinfo/content_manifests`? I recall we talked about adding a `tmpfiles.d` dropin if we really need it available under `/root`.

gursewak1997: My bad, I assumed we could write to` /var`and not to `/root`. I will go with `/usr/share/buildinfo/content_manifests` for now. Thanks:)

cgwalters: Where can we ask these security scanners to change?  Using `/root` for this just conflicts with ostree today, and actually more generally would be problematic for many "image based updates" models.

Even having traditional `yum/rpm` write to `/root` is pretty ugly IMO.

miabbott: > Where can we ask these security scanners to change? Using `/root` for this just conflicts with ostree today, and actually more generally would be problematic for many "image based updates" models.
> 
> Even having traditional `yum/rpm` write to `/root` is pretty ugly IMO.

I made the assumption that it was a security scanner, but in the original ticket there is a reference to just `scanners`, which I guess could be anything doing any kind of scanning.

Since this request came from RHT ProdSec, we'd have to follow-up with them in https://issues.redhat.com/browse/GRPA-3731

miabbott: Ah, actually it is security scanners - `There is significant interest in getting this feature added to RHCOS from our security scanning partners.`

miabbott: I think it is fine to start working to change the security scanners, though I don't know how fast we can make that happen. Let's work with ProdSec to understand which scanners are looking for this file and see if we can get them modified.

Additionally, @ashcrow pointed out that the location of the file comes from OSBS [1][2], so we could work with OSBS to modify the location of the JSON file. Or suggest a secondary location if the first location isn't available.

But given the nature of this request, I think we have to work within the constraints of the existing tools, so we should pursue the use of `tmpfiles.d` to land the file in the location desired.

[1] https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/constants.py#L200
[2] https://osbs.readthedocs.io/en/latest/users.html#image-content-manifests

cgwalters: There's a reason though that ostree designs the constraints it does around `/usr` and `/var` etc.  We can certainly use `tmpfiles.d` or a custom systemd unit or whatever to write this content in `/var`.

But the thing is - that means that `rpm-ostree rollback` will mean the data in `/var` is out of sync with `/usr`.  

This doesn't matter a huge amount for containers because the larger container ecosystem doesn't support in-place offline updates - it's all about either running `yum update` (in place, online), or just tearing down the whole container and starting a new one (Kubernetes native style).

As I commented on the ticket, because rpm-ostree natively sticks this data in the commit object today, it is versioned alongside the deployment, which means that one can easily do e.g. diffs between them, and `rpm-ostree rollback` will always have the right data.

(Also, data in `/var` is mutable by default, whereas clearly this should be immutable)

jlebon: To clarify, the proposal here is to make e.g. `/root/buildinfo` be a symlink to `/usr/share/buildinfo`, not to copy.
Agree though that ideally we get scanners and other producers to use the data that's already available in commit metadata.

miabbott: Thanks for commenting on the Jira ticket with your input. I think we all completely missed the use of `rpmostree.rpmmd-repos` as a solution to this problem.

However, since the current tooling is looking for the information in a specific location, I still think we have to provide the data there...somehow.

Perhaps this is hacky, but might provide a solution given the constraints:

1. postprocess script that writes out the desired JSON file at `/usr/share/buildinfo/content_manifests/<nvr>.json`
2. a systemd unit that runs at boot which ensures there is a `/var/roothome/buildinfo/content_manifests/` directory
3. same systemd unit creates a symlink from  `/usr/share/buildinfo/content_manifests/<nvr>.json` to `/var/roothome/buildinfo/content_manifests/<nvr>.json`

WDYT?

jasinner: I think the location of the content_manifests/<nvr>.json does not have to be in /root/buildinfo, and /usr/share is fine. We can talk to our security scanning partners about the location once the content is ready.

Also, I don't think we can use the rpmostree.rpmmd-repos as is. We need to translate those repositories names to something in pulp, like the the repositories listed [here](https://access.redhat.com/security/data/metrics/repository-to-cpe.json), or directly to CPE (from Errata tool).

manifest.yaml
```diff
@@ -164,6 +164,29 @@ postprocess:
...
+     #!/usr/bin/env bash
+     set -euo pipefail
+     
+     REPOS=`jq .repos /usr/share/rpm-ostree/treefile.json`
+     ARCH=`jq .basearch /usr/share/rpm-ostree/treefile.json | tr -d '"'`
+     
+     . /etc/os-release
+     mkdir -p /usr/share/buildinfo/content_manifests && cd "$_" && touch $ID-$OSTREE_VERSION.json
+     # Make an empty json and add all the relevant fields
+     echo "$(var=$REPOS; jq --argjson repos "$var" '. +=
```
jlebon, line 179: I don't entirely follow the `jq` syntax here, though one thing to note is that the repo names need to be translated to those in https://access.redhat.com/security/data/metrics/repository-to-cpe.json which I think is missing here.

The actual set of repos we use doesn't really change that often, so for this first iteration, we could just hardcode the fact that e.g. `rhel-8-baseos` maps to `rhel-8-for-$ARCH-baseos-rpms`, `rhel-8-appstream` to `rhel-8-for-$ARCH-appstream-rpms`, etc... and error out if we get a repo name we don't know.

Maybe cleaner eventually is to make that mapping part of the repo files themselves and add a treefile option to teach rpm-ostree to extract it and generate the JSON file and potentially even validate against https://access.redhat.com/security/data/metrics/repository-to-cpe.json.

overlay.d/05rhcos/usr/lib/tmpfiles.d/buildinfo.conf
```diff
@@ -0,0 +1,10 @@
+#  This file is part of systemd.
```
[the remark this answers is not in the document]
gursewak1997: True but @miabbott suggested on going through with the creation of .conf file and symlinking the JSON file to /roothome.

manifest.yaml
gursewak1997: Mapping the repo names to their pulp repo IDs (ie, "[rhel-8-baseos]" to "[rhel-8-for-x86_64-baseos-eus-rpms__8_DOT_4]") would be a bit more complex task to put in postprocess so we decided to not do that here. I have just added the given repo names for now so that we can access them when mapping them to [repository-to-cpe.json](https://access.redhat.com/security/data/metrics/repository-to-cpe.json)
So far, we are just landing the changes to create the relevant DIR/JSON file and symlinking it to /root.

jlebon: Had a chat with @gursewak1997 about this. I think in the end doing this in a postprocess script is probably not the right approach. We could hardcode equivalencies here, though the second we change the repo definition files in the internal redhat-coreos repo, it'll become incorrect data. Because of this, I think this data belongs best sitting alongside the repo files.

Apart from the rpm-ostree suggestion in the comment above, another easier option is a YAML file similar to the official `content_sets.yml` currently in use ([internal example](https://url.corp.redhat.com/177f730)):

```yaml
# content_sets.yaml
repo_mapping:
  rhel-8-baseos:
    name: rhel-8-for-$ARCH-baseos-eus-rpms__8_DOT_4
  rhel-8-appstream:
    name: rhel-8-for-$ARCH-appstream-eus-rpms__8_DOT_4
  ...
```

(This would live in the `redhat-coreos` repo so it could be maintained atomically with the repo definitions and would be similarly copied into the workdir before builds.)

And then `cosa build` would check for the existence of that file and auto-generate the JSON file to inject in the OS.

miabbott: My understanding was that Gursewak was going to work on the tooling to generate the `<nvr>.json` file and we would handle the problem of the repo names separately. The last working idea was just to rename all the repos to match what is in the `repository-to-cpe.json` file after the tooling was in place.

The YAML mapping file would also work; either way we are going to incur some amount of manual cost to maintain that the repo names are valid according to `repository-to-cpe.json`.

miabbott: I discussed this with @jlebon today and we discussed various approaches to this, but ultimately settled on the proposal that Jonathan made in his last comment:

- a YAML doc that lives alongside the repo files which maps repo names to names in `repository-to-cpe.json`
  - additionally should include the `baseurl` so we can detect when that changes
- changes to `cosa build` that can consume the mapping YAML doc and produce the `<nvr>.json` file
- (future project) CI job on `openshift/release` that enforces changing the mapping YAML doc when changing the repo files

PR #671, opened 2021-11-15
OWNERS: add gursewak1997

miabbott:
Gursewak missed out on our on-boarding checklist, so he was missing
from the OWNERS file.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -19,3 +19,4 @@ approvers:
   - aaradhak
   - ravanelli
   - peberanek
+  - gursewak1997


miabbott:
> > User is not a member of the org.
> 
> Looks like that needs to be fixed first.

Yup, I've reached out to Gursewak to get that sorted out.

miabbott:
Gursewak should be part of the org now.  If someone wants to drop a `/lgtm` that would be great.

miabbott:
> /verify-owners

Of course there is a command for that...

PR #673, opened 2021-11-16
Bug 2023836: gitmodules: change f-c-c branch to rhcos-4.9

miabbott:
I pushed an `rhcos-4.9` branch to fedora-coreos-config (based on
coreos/fedora-coreos-config@7d3c177f43af646bac634bd0e8fbcedab27e887e),
so we should start using that branch for 4.9 builds.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "fedora-coreos-config"]
 	path = fedora-coreos-config
 	url = https://github.com/coreos/fedora-coreos-config
-	branch = testing-devel
+	branch = rhcos-4.9


jlebon (approved):

travier (approved):

cgwalters:
Tangentially related, longer term it's probably better to do something like use a git submodule for `main/master` to ensure we keep exactly in sync, but since backports almost always create desynchronization, we could switch from a submodule to just copying.  Then for backports there wouldn't be an awkward need for two PRs for each change, *and* we wouldn't pollute the fedora-coreos-config git repo with rhcos branches.

Or, we could try out using a [subtree merge](https://docs.github.com/en/get-started/using-git/about-git-subtree-merges) in both cases.

PR #676, opened 2021-11-16
ci: enable use of buildfetch

miabbott:
This changes the CI job to fetch the last previous build metadata available
from the redirector URL, so that we are able to have some ability to
see what kind of RPM changes (if any) are introduced by changes in the upstream
repos/environment.

### DIFF
--- a/ci/prow-build-test-qemu.sh
+++ b/ci/prow-build-test-qemu.sh
@@ -4,6 +4,8 @@ set -xeuo pipefail
 # entrypoint for PRs to this repo, as well as for PRs on other repos,
 # mainly coreos-assembler.  It assumes that `cosa init` has been run.
 
+REDIRECTOR_URL="https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/"
+
 # record information about cosa + rpm-ostree
 if test -d /cosa; then
     jq . < /cosa/coreos-assembler-git.json
@@ -26,7 +28,9 @@ fi
 # Grab the raw value of `mutate-os-release` and use sed to convert the value
 # to X-Y format
 ocpver=$(rpm-ostree compose tree --print-only src/config/manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|')
+prev_build_url=${REDIRECTOR_URL}/rhcos-${ocpver}/
 curl -L http://base-"${ocpver}"-rhel8.ocp.svc.cluster.local > src/config/ocp.repo
+cosa buildfetch --url=${prev_build_url}
 cosa fetch
 cosa build
 cosa buildextend-extensions


ci/prow-build-test-qemu.sh
```diff
@@ -4,6 +4,8 @@ set -xeuo pipefail
 # entrypoint for PRs to this repo, as well as for PRs on other repos,
 # mainly coreos-assembler.  It assumes that `cosa init` has been run.
 
+REDIRECTOR_URL="https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/"
```
cgwalters, line 7: I would probably have done this by adding e.g. `cosa buildprep --distro=rhcos` or so in cosa, then use it here.  

That said, this actually leads into a whole other huge detail, which is around the split between bootimages and the oscontainer update that is prominent in RHCOS.

In some cases, if we're intending to do a bootimage update, starting CI from the previous bootimage would make sense.  And that's a separate URL.

There's also the long running huge problem that we can (and often have because brew) push an oscontainer without updating the redirector/S3, so the data there can be incorrect, and what we'd really want to do is fetch the previous oscontainer here.

Anyways, none of that should block this, which is a good improvement.  Thanks for doing this!

cgwalters (approved):
Nice, thanks for doing this!

I probably would have added `cosa buildfetch -b rhcos` or so.  This also relates strongly to debate in

ci/prow-build-test-qemu.sh
miabbott: > I would probably have done this by adding e.g. `cosa buildprep --distro=rhcos` or so in cosa, then use it here.

I'll pursue this as a follow-up.

> That said, this actually leads into a whole other huge detail, which is around the split between bootimages and the oscontainer update that is prominent in RHCOS.
> 
> In some cases, if we're intending to do a bootimage update, starting CI from the previous bootimage would make sense. And that's a separate URL.

I think in the case here, where we are just testing changes to this repo, starting from the last version found on the redirector is sane enough. We can improve the CI around bootimage bumps to `openshift/installer` by building out CI jobs that verify cluster installs (https://issues.redhat.com/browse/GRPA-4447).

Additionally (something that I think I'll pursue in the near future), we should expand the CI job here to build the metal + ISO artifacts so we can run the `testiso` suite.  Getting us closer to having more confidence of upstream changes and downstream becomes a rebuild.

> There's also the long running huge problem that we can (and often have because brew) push an oscontainer without updating the redirector/S3, so the data there can be incorrect, and what we'd really want to do is fetch the previous oscontainer here.
> 
> Anyways, none of that should block this, which is a good improvement. Thanks for doing this!

We should be in a better position to avoid that scenario of pushed oscontainer/no S3 update since we've removed the Brew stage from the downstream job. I'll investigate pulling the oscontainer as part of this CI job.

miabbott:
I think `build02` is still unhealthy, but let's check...

/retest

Issue #674, opened 2021-11-16
`crio` tests failing in downstream tests for RHCOS 4.10


miabbott:
```
--- FAIL: crio.base (1225.37s)
    --- PASS: crio.base/crio-info (0.16s)
    --- FAIL: crio.base/pod-continues-during-service-restart (600.00s)
            harness.go:413: TIMEOUT[10m0s]
    --- FAIL: crio.base/networks-reliably (600.00s)
            harness.go:413: TIMEOUT[10m0s]
```

Seems to be correlated to introduction of `crio-1.23.0-23.rhaos4.10.git407d866.el8`

```
ostree diff commit from: a654ec98dfbd0f1309c921465afe2953f284257bbdd98eae85996e3486a0da24
ostree diff commit to:   5f64085e365ed08fcbad9f0ccfa9caeefe1476b5dcd0df2e646475596dda82ce
Upgraded:
  cri-o 1.23.0-12.rhaos4.10.git6ee64e9.el8 -> 1.23.0-23.rhaos4.10.git407d866.el8
  openshift-clients 4.10.0-202111092341.p0.git.a253d9d.assembly.stream.el8 -> 4.10.0-202111161511.p0.g96e0fad.assembly.stream.el8

New image input checksum: f86f660fa1c02152f8858478382a2c4e88ae864ca024636c8fba14a1864e553a
```

Changes in `cri-o` corresponding to the git commits:

```
$ git shortlog --no-merges 6ee64e9..407d866
Maximilian Ehlers (1):
      Updates kubeadm.md: The cgroup property is removed in [kubeadm-config.v1beta3](https://kubernetes.io/docs/reference/config-api/kubeadm-config.v1beta3/)

Peter Hunt (6):
      oci: make some channels buffered
      oci: always close chControl
      oci: fix deadlock in container stop code
      release-notes: update to main
      gh-actions: add sed for kube e2e
      server: fix segfault when using cgroupv2

Qi Wang (2):
      Bug 2012838: fix override storage options from storage.conf
      Specify runtime table format in  the error message

dependabot[bot] (6):
      build(deps): bump go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
      build(deps): bump google.golang.org/grpc from 1.41.0 to 1.42.0
      build(deps): bump go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
      build(deps): bump github.com/onsi/gomega from 1.16.0 to 1.17.0
[...]

PR #677, opened 2021-11-17
denylist: add crio.base

miabbott:
This is failing on all arches right now and no resolution in sight;
let's skip this test for now.

See: https://github.com/openshift/os/issues/674

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,3 +7,5 @@
    - s390x
 - pattern: coreos.ignition.journald-log
   tracker: https://github.com/coreos/coreos-assembler/issues/1173
+- pattern: crio.base
+  tracker: https://github.com/openshift/os/issues/674


cgwalters:
The interesting thing here is whether this crio works in OpenShift right?  Do we know the answer to that yet?

But it is true that some of these failures in the past have been specific to our hacked-up "stand up crio outside of openshift" bits.

Since upstream crio is gating on OpenShift tests *and* we have e2e gating on both the installer and promote job we could perhaps just drop our crio testing and rely on that.  So...

/lgtm

miabbott:
> The interesting thing here is whether this crio works in OpenShift right? Do we know the answer to that yet?
> 
> But it is true that some of these failures in the past have been specific to our hacked-up "stand up crio outside of openshift" bits.
> 
> Since upstream crio is gating on OpenShift tests _and_ we have e2e gating on both the installer and promote job we could perhaps just drop our crio testing and rely on that. So...
> 
> /lgtm /override ci/prow/build-test-qemu

Thanks for the override.  This will at least unblock builds and get newer versions of crio/kubelet to folks who need them.

I'm like 87% in favor of dropping our `crio` tests, but the few times we have failed did highlght a regression (or at least change in defaults) in `crio` behavior.  Though the main use case of `crio` in OCP has continued to be successful.

PR #678, opened 2021-11-18
start sharing FCOS external tests more broadly

miabbott:
This bumps the `fedora-coreos-config` submodule to the latest, which pulls in the changes to categorize the external tests more concretely. Additionally, this also changes the RHCOS to mount `/boot` by UUID as we do in FCOS.

As part of the change, the individually symlinked FCOS tests are replaced by a single `shared` directory where all the FCOS external tests are contained.


jlebon (approved):
Nice work on this! The boot UUID change in particular is significant, though I don't foresee any issues. It's good that we have upgrade tests in RHCOS too now.

miabbott:
> Nice work on this! The boot UUID change in particular is significant, though I don't foresee any issues. It's good that we have upgrade tests in RHCOS too now.

I ran the usual gamut of tests against metal, metal4k, live ISO, and the upgrade path.  Everything came back happy.

jlebon:
AFAICT, there are no RHCOS-specific tests in this repo, so I think we can delete all the copies in `tests/kola` and have just the `tests/kola/shared` symlink, no?

miabbott:
> AFAICT, there are no RHCOS-specific tests in this repo, so I think we can delete all the copies in `tests/kola` and have just the `tests/kola/shared` symlink, no?

I think the only lingering test that I'd like to keep around is the `misc-ro` test; that can be split up into separate tests and deduplicated as a follow-up.

I think the other thing that should be done is the untagging of tests in FCOS that are checking for mounting `/boot` by UUID, since we will get that in RHCOS as part of the submodule update.

miabbott:
/hold

Let's wait for coreos/fedora-coreos-config#1344

travier (approved):
I did not test this change but this looks good.

miabbott:
/hold cancel

The changes to fedora-coreos-config seem to always be happening, so let's just draw a line and pull in the latest.

See the individual commits for details, but the newest change from f-c-c is the inclusion of `crun` in RHCOS.

travier (approved):
/lgtm

mike-nguyen (approved):
LGTM

PR #680, opened 2021-11-23
Break up misc-ro into separated cases

HuijingHei:
These are RHCOS only cases.

Test with local qcow2 image and results are PASSED
```
--- PASS: non-exclusive-tests (64.55s)
    --- PASS: non-exclusive-tests/ext.os.version.rhel-matches-rhcos-build (1.65s)
    --- PASS: non-exclusive-tests/ext.os.files.usrlocal-fixup (0.68s)
    --- PASS: non-exclusive-tests/ext.os.version.rhaos-pkgs-match-openshift (1.66s)
    --- PASS: non-exclusive-tests/ext.os.console.console-login-helper-service (1.64s)
    --- PASS: non-exclusive-tests/ext.os.disk.iscsi-initiatorname-service (1.64s)
    --- PASS: non-exclusive-tests/ext.os.files.conntrack-tools (1.66s)
    --- PASS: non-exclusive-tests/ext.os.files.check-qemu-ga (0.72s)
    --- PASS: non-exclusive-tests/ext.os.afterburn.sshkey-service (1.71s)
```


travier (commented):
This should probably also remove the existing misc-ro test at the same time?

HuijingHei:
> This should probably also remove the existing misc-ro test at the same time?

Yes, thanks for your remind, I will remove it.

HuijingHei:
Done

tests/kola/console/console-login-helper-service
```diff
@@ -0,0 +1,27 @@
...
+}
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# Check that we have the proper presets for console-login-helper-messages in RHCOS.
+# Versions of CLHM prior to v0.21 have `issuegen`-related systemd units.
+if test -f /usr/lib/systemd/system/console-login-helper-messages-issuegen.service; then
```
travier, line 19: This probably should always exist. Noting here for reference. Can be in a follow up PR.

travier (commented):
OK, the diff here is not 1 to 1 as some of the tests are moved here: https://github.com/coreos/fedora-coreos-config/pull/1350.
Also, this needs https://github.com/openshift/os/pull/678 and careful review to make sure we do not lose tests here.
A list of all tests that were in this repo before and were they went would help. I did a manual check and things look good so far.

travier:
Initial set of misc fixes in https://github.com/coreos/fedora-coreos-config/pull/1351

HuijingHei:
@travier , thanks for your review! 
I updated the script according to your comments, could you help to review? PR #678 will remove the duplicated cases.

miabbott:
Looks good Huijing!  I'd really like to land #678 first and then we can rebase this PR on top.

HuijingHei:
> Looks good Huijing! I'd really like to land #678 first and then we can rebase this PR on top.

Sure, thanks @miabbott

miabbott (approved):
/lgtm

PR #681, opened 2021-11-25
Get more container-tools packages from OCP

yselkowitz:
These just showed up as newer in OCP than in RHEL.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -271,14 +271,16 @@ repo-packages:
       - nss-altfiles
   - repo: rhel-8-server-ose
     packages:
-      # eventually, we want the one from the container-tools module, but we're
-      # not there yet
-      - toolbox
-      # we are going to start pulling the whole container-tools stack from the
-      # OCP repo
+      # packages which exist in containers-tools:rhel8 but newer versions exist
+      # in OCP repo for the purposes of RHCOS
+      - conmon
+      - containernetworking-plugins
+      - containers-common
       - container-selinux
       - crun
       - runc
+      - skopeo
+      - toolbox
 
 modules:
   enable:


jlebon:
> These just showed up as newer in OCP than in RHEL.

Hmm, did you mean the other way?

If we currently don't want anything from the container-tools module, can't we just remove `container-tools:rhel8` from `modules.enable` instead?

yselkowitz:
No, they're newer in OCP now but they're not getting into RHCOS because of modular vs. non-modular conflicts.  #668 was a recent such case.

I can't speak to whether RHCOS is 100% ready to drop the container-tools:rhel8 module, or whether this is a permanent or temporary change, just that right now RHCOS isn't getting the intended builds of these packages for this reason.

jlebon:
CI is failing on

```
 Problem: conflicting requests
  - nothing provides containers-common >= 2:1-2 needed by skopeo-2:1.5.1-3.rhaos4.10.el8.x86_64 
```

Probably need to add `containers-common` to the list?

jlebon:
Thanks!
/lgtm

> I can't speak to whether RHCOS is 100% ready to drop the container-tools:rhel8 module, or whether this is a permanent or temporary change, just that right now RHCOS isn't getting the intended builds of these packages for this reason.

Ack. Let's discuss this next week.

yselkowitz:
> > I can't speak to whether RHCOS is 100% ready to drop the container-tools:rhel8 module, or whether this is a permanent or temporary change, just that right now RHCOS isn't getting the intended builds of these packages for this reason.
> 
> Ack. Let's discuss this next week.

Given that there are 18 packages in the container-tools module, and only 6 are overrides in RHCOS, it seems we're not quite at that point yet?

In the meantime, I still can't get this dependency to resolve. :-(

yselkowitz:
The problem here is that the non-modular RHCOS/OCP skopeo build has a Requires on the (currently only) modular RHEL containers-common.  Unless I'm doing something wrong here (and I've tried a few different things) without the enable_hotfixes hammer, I'm not sure if that's possible.  Possible solutions include:

* building containers-common for RHCOS/OCP (but keeping it compatible with the modular version, as buildah and podman both use it too);
* cross-tagging the latest RHEL 8.4.z version into OCP, but then you'll probably have to keep it updated
* completing the transition away from container-tools:rhel8 module to RHCOS/OCP builds of all required components (if that's indeed the intention)
* last resort, `enable_hotfixes=1`
* any other ideas?

jlebon:
> The problem here is that the non-modular RHCOS/OCP skopeo build has a Requires on the (currently only) modular RHEL containers-common.

Hmm, looking at this in Brew, I can't actually find a containers-common modular build for 8.4.0. Do you see one? There are `containers-common` RPMs in Appstream, but none of them are epoch >= 2. I think they're all from when it was part of skopeo. So even with `module_hotfixes=1` this wouldn't work I think. (OK yup, just tested this.)

> building containers-common for RHCOS/OCP

:+1: Sounds like the simplest to me short-term.

>  (but keeping it compatible with the modular version, as buildah and podman both use it too);

Shouldn't the module have its own containers-common RPM?

> Given that there are 18 packages in the container-tools module, and only 6 are overrides in RHCOS, it seems we're not quite at that point yet?

My concern is more the mix-and-match approach we have right now of only picking some packages from the container-tools module and rebuilding others for OCP. The modular RPMs are built and shipped together and move at their own pace. So things like this are going to continue happening unless there's clear procedures in place.

Half-baked idea: maybe instead we should add new `rhaos-4.$x` streams of the `container-tools` module in the RHAOS plashets and in RHCOS we just ship that? Then, the containers team can just keep dealing in modularity terms instead of straddling both worlds. And for RHCOS, all the container-related packages come to us as a single unit.

yselkowitz:
> Hmm, looking at this in Brew, I can't actually find a containers-common modular build for 8.4.0. Do you see one? There are `containers-common` RPMs in Appstream, but none of them are epoch >= 2. I think they're all from when it was part of skopeo. So even with `module_hotfixes=1` this wouldn't work I think. (OK yup, just tested this.)

Good catch!
 
> > building containers-common for RHCOS/OCP
> 
> +1 Sounds like the simplest to me short-term.

If this was split out of skopeo and you're overriding RHEL's skopeo with your own build, this would be a necessity.

> > (but keeping it compatible with the modular version, as buildah and podman both use it too);
> 
> Shouldn't the module have its own containers-common RPM?

Maybe, but as you point out, not until 8.5+.  I'm still not sure if that would work though.

> > Given that there are 18 packages in the container-tools module, and only 6 are overrides in RHCOS, it seems we're not quite at that point yet?
> 
> My concern is more the mix-and-match approach we have right now of only picking some packages from the container-tools module and rebuilding others for OCP. The modular RPMs are built and shipped together and move at their own pace. So things like this are going to continue happening unless there's clear procedures in place.
> 
> Half-baked idea: maybe instead we should add new `rhaos-4.$x` streams of the `container-tools` module in the RHAOS plashets and in RHCOS we just ship that? Then, the containers team can just keep dealing in modularity terms instead of straddling both worlds. And for RHCOS, all the container-related packages come to us as a single unit.

I'm not sure if plashets are module-aware tbh, this would need some discussion with the ART team.

yselkowitz:
Repushed, but that will fail until there's an OCP/RHCOS build of containers-common.

yselkowitz:
in the meantime
/hold

miabbott:
> Half-baked idea: maybe instead we should add new `rhaos-4.$x` streams of the `container-tools` module in the RHAOS plashets and in RHCOS we just ship that? Then, the containers team can just keep dealing in modularity terms instead of straddling both worlds. And for RHCOS, all the container-related packages come to us as a single unit.

I've had numerous conversations with the Containers team and we decided on OCP specific branches for all of the `container-tools` content. This is still a work in progress, but would explain why @yselkowitz was seeing newer versions of the packages in the OCP plashets. The intent is that we start pulling all the `containers-tool` content from the plashets and not use the versions that will be shipped as part of the module(s) in RHEL 8.4. This way we keep getting newer versions of the `container-tools` content, but retain the stable RHEL 8.4 base.

I'll follow-up on this next week when I am back from PTO, but wanted to toss in some context in the meantime.  Apologies if this plan wasn't well communicated.

miabbott:
> I'll follow-up on this next week when I am back from PTO, but wanted to toss in some context in the meantime. Apologies if this plan wasn't well communicated.

I'm working with the packager of the `container-tools` content to create versions of all the needed RPMs that will appear in the OCP plashets. If all goes well, we'll have the bits we need in the places we want sometime this week.

yselkowitz:
Any update here?

miabbott:
> Any update here?

Been mired in Jira tickets waiting for EXD to configure branches + add packages to Brew tag listings.  I *think* most of that is completed and we can start this effort again.

miabbott:
> Been mired in Jira tickets waiting for EXD to configure branches + add packages to Brew tag listings. I think most of that is completed and we can start this effort again.

Latest is we need a dependency added to the buildroot of `rhaos-4.10-rhel-8`; there's a ticket with EXD about this

cgwalters:
All this relates to https://github.com/openshift/aos-cd-jobs/pull/2849#issuecomment-990154749 BTW - since these builds are targeted for OCP, they should be gated on OCP CI and ideally work in a more OCP way where we'd e.g. have github.com/openshift/skopeo or so.  (But, I know that's a radical proposal)

The more middle ground aspect might be a new github.com/openshift/rhcos-components that is like dist-git - and *acts as the source of truth*.  We'd also then have an obvious place to hook up "test this skopeo with OCP in CI".

Or, another option going forward with RHEL9/C9S is to have a github repo that is just a *list*, but pulls components from e.g. an
`openshift-4.11` branch on https://gitlab.com/redhat/centos-stream/rpms/skopeo/

Then we'd have to hook up CI on gitlab that talks to Prow somehow.  But *that* would have huge potential benefits.

yselkowitz:
Post-break ping

miabbott:
> Post-break ping

Still waiting on EXD to properly configure buildroots

miabbott:
TL;DR - OCP/RHCOS 4.10 will continue to get **most** of the `container-tools` content from the RHEL 8.4.Z EUS `container-tools:rhel8` module stream.  (See caveats at the bottom of this comment)

---

We've recognized that users/customers want to have most of the OS remain stable, but would like to be able to use newer pieces of the container stack, i.e. `podman`, `crun`, etc.

Towards that goal, some folks from CoreOS have had multiple discussions with the Container team on how best to achieve it.

At first, we thought it would be as simple as "build the newest `container-tools` content in the OCP 4.10 tag and make it available via the RHOSE repo".  We tried to include the latest version of `container-tools:rhel8` that was targeting RHEL 8.6.

This ran into a number of snags that included:
  - conflicting dependencies on `protobuf-c` in the OCP 4.10 build root when building `criu`
  - install dependencies on unreleased RHEL 8.6 content for things like `podman`

We didn't want to be in the position of having to ask permission to ship unreleased RHEL 8.6 content as part of OCP/RHCOS 4.10, so we punted on this original plan.

We regrouped and decided to try just making the latest version of `container-tools:rhel8` that is available to RHEL 8.5 customers.  This would still get us newer versions of RPMs, but shouldn't require us to ship unreleased content.

Unfortunately, that decision was made a bit too late towards code freeze and we ran into additional issues trying to make that work:
- `container-selinux` isn't building  due to a `selinux-policy` dependency problem
- `criu` is still not building due to conflicting `protobuf-c`dependencies in the buildroot

So for OCP/RHCOS 4.10 GA, we've abandoned the plan to get newer `container-tools` content into RHCOS.  We do plan to revisit this idea as part of 4.11 (and hopefully 4.10.z).

For now the versions of RPMs from the `container-tools` stack that will be available in RHCOS 4.10 will look like this:

```
conmon-2.0.29-1.module+el8.4.0+11822+6cc1e7d7.x86_64
container-selinux-2.173.1-0.rhaos4.10.el8.6.noarch *
containernetworking-plugins-0.9.1-1.module+el8.4.0+11822+6cc1e7d7.x86_64
containers-common-1.3.1-5.module+el8.4.0+11990+22932769.x86_64
criu-3.15-1.module+el8.4.0+11822+6cc1e7d7.x86_64
crun-1.2-3.rhaos4.10.el8.x86_64 **
fuse-overlayfs-1.6-1.module+el8.4.0+11822+6cc1e7d7.x86_64
libslirp-4.3.1-1.module+el8.4.0+11822+6cc1e7d7.x86_64
podman-3.2.3-0.11.module+el8.4.0+12050+ef972f71.x86_64
runc-1.0.1-3.rhaos4.10.el8.x86_64 ***
skopeo-1.3.1-5.module+el8.4.0+11990+22932769.x86_64
slirp4netns-1.1.8-1.module+el8.4.0+11822+6cc1e7d7.x86_64
```

* `container-selinux` is newer than what is included in RHEL 8.4.Z EUS, but addresses some BZs that customers were encountering
** `crun` is newer than what is included in RHEL 8.4.Z EUS, because the k8s folks wanted to start testing with 1.x versions of `crun`
*** `runc` is slightly newer than what is included in RHEL 8.4.Z EUS, because we had previously shipped a newer version in OCP/RHCOS 4.9 and had to preserve the upgrade path

jlebon:
Thanks Micah for the breakdown. In that case, let's document this a bit in the manifest: https://github.com/openshift/os/pull/696.

PR #684, opened 2021-12-02
kola-denylist: temporarily exclude coreos.ignition.mount.* on s390x

julianwiedmann:
This is the companion PR to https://github.com/coreos/coreos-assembler/pull/2538. For this we need to exclude the relevant tests on s390x, until a fixed gdisk has also landed in RHCOS (see the tracker BZ).


jlebon (approved):
/lgtm

Thanks!

saqibali-2k (approved):
Looks good! Just an FYI, there is also the option of snoozing a test to a particular date as show in the example [here](https://coreos.github.io/coreos-assembler/kola/#kola-run).

PR #685, opened 2021-12-09
25rhcos-azure-udev-rules: add 50-azure-ptp.rules

jlebon:
This will create the `/dev/ptp_hyperv` symlink. We'll need this once our
chrony generator uses it:

https://github.com/coreos/fedora-coreos-config/pull/1355

### DIFF
--- a/overlay.d/25rhcos-azure-udev-rules/usr/lib/udev/rules.d/50-azure-ptp.rules
+++ b/overlay.d/25rhcos-azure-udev-rules/usr/lib/udev/rules.d/50-azure-ptp.rules
@@ -1,0 +1,5 @@
+# Manual backport of:
+# https://github.com/systemd/systemd/commit/32e868f058da8b90add00b2958c516241c532b70
+# Can drop once we get to RHEL 8.6:
+# https://bugzilla.redhat.com/show_bug.cgi?id=1991834
+SUBSYSTEM=="ptp", ATTR{clock_name}=="hyperv", SYMLINK += "ptp_hyperv"
--- a/tests/kola/version/drop-azure-ptp
+++ b/tests/kola/version/drop-azure-ptp
@@ -1,0 +1,13 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ -e /usr/lib/udev/rules.d/50-azure-ptp.rules ] && \
+    grep ptp_hyperv /usr/lib/udev/rules.d/50-udev-default.rules; then
+  fatal "50-udev-default.rules includes ptp_hyperv symlink; drop 50-azure-ptp.rules"
+fi


overlay.d/25rhcos-azure-udev-rules/usr/lib/udev/rules.d/50-azure-ptp.rules
```diff
@@ -0,0 +1,5 @@
+# Manual backport of:
+# https://github.com/systemd/systemd/commit/32e868f058da8b90add00b2958c516241c532b70
+# Can drop once we get to RHEL 8.6:
```
dustymabe, line 3: > # Can drop once we get to RHEL 8.6:

Is there anything we can put in place now to force us to do that?

dustymabe: seems like a ext test that said `if RHEL 8.6 && -f  50-azure-ptp.rules; then fail`

dustymabe:
tested in azure with a local dev build:

```
[bound] -bash-4.4$ cat /etc/os-release  | head -n 2 
NAME="Red Hat Enterprise Linux CoreOS"
VERSION="410.84.202112161449-0"
[bound] -bash-4.4$ ls -l /dev/ptp_hyperv 
lrwxrwxrwx. 1 root root 4 Dec 16 15:25 /dev/ptp_hyperv -> ptp0
[bound] -bash-4.4$ chronyc sources
210 Number of sources = 1
MS Name/IP address         Stratum Poll Reach LastRx Last sample               
===============================================================================
#* PHC0                          0   3   377     6  -2651ns[-2919ns] +/- 5551ns
```

jlebon:
Thanks a lot @dustymabe!

tests/kola/version/drop-azure-ptp
```diff
@@ -0,0 +1,13 @@
...
+set -xeuo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ -e /usr/lib/udev/rules.d/50-azure-ptp.rules ] && \
+    grep ptp_hyperv /usr/lib/udev/rules.d/50-udev-default.rules; then
+  fatal "50-udev-default.rules includes ptp_hyperv symlink; drop 50-azure-ptp.rules"
```
dustymabe, line 12: ❤️

dustymabe (approved):
LGTM

overlay.d/25rhcos-azure-udev-rules/usr/lib/udev/rules.d/50-azure-ptp.rules
jlebon: I'm thinking we should extend kola to handle these kinds of checks so they're non-fatal but still loud. Some ideas:
- have kola check test output for `KOLAWARNING: ...` and reprint the warning on the console.
- add a `warn-on-rc: N` field to kola test JSON where if a test exits with the given code, kola prints `WARN: $testname` instead of `FAIL: $testname`.

We can then have kola e.g. touch a file or print a report of warnings and the Slack notifier in the pipelines can check for this file and tweak the notification.

miabbott:
I think the status reporting is wedged; going to close/open the PR to try to kick it

overlay.d/25rhcos-azure-udev-rules/usr/lib/udev/rules.d/50-azure-ptp.rules
dustymabe: carried forward discussion in https://github.com/coreos/coreos-assembler/issues/2620

PR #686, opened 2021-12-10
manifest: get crun from the RHAOS repos

miabbott:
We're starting to build all of the `container-tools` packages from the
`rhaos-4.X-rhel8` branches in Brew, so that we have more up=to-date
versions of those packages.

The `crun` package has been hotly requested, so let's start with that
package since it is built and ready.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -276,8 +276,9 @@ repo-packages:
       - toolbox
       # we are going to start pulling the whole container-tools stack from the
       # OCP repo
-      - runc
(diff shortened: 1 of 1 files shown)


jlebon (approved):

saqibali-2k (approved):
/lgtm

travier:
This fails on runc missing from `rhel-8-server-ose` in a local build but does not fail in CI for an unknown reason 🤔

PR #688, opened 2021-12-16
scripts/upload-ami: add support for making Aliyun images public

miabbott:
This allows us to configure Aliyun images as publicly available,
similar to what we do with AMIs.

I refactored a good chunk of the code to be a bit more reusable across
clouds.

Ideally, we'd have more of this in `coreos-assembler`, but this is
(hopefully) a short-term bandaid.

(See coreos/coreos-assembler#2604 where the `aliyun` CLI was added.)

### DIFF
--- a/scripts/upload-ami
+++ b/scripts/upload-ami
@@ -14,6 +14,11 @@ OPENSHIFT_CI_ACCOUNT = "460538899914"
 OPENSHIFT_DEV_ACCOUNT = "269733383066"
 OPENSHIFT_MARKETPLACE_TEST_ACCOUNT = "125666959065"
 
+# See https://issues.redhat.com/browse/SPLAT-193 for account values
+OPENSHIFT_ALIYUN_CI = "5920522175273455"
+OPENSHIFT_ALIYUN_DEV = "5807403157081522"
+OPENSHIFT_ALIYUN_QE = "5724326381648897"
+
 # Since we run all tests on QEMU, and as of today don't have really
 # any tests that are specific to a cloud provider, we just run
 # basic sanity checks.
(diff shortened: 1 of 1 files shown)


scripts/upload-ami
```diff
@@ -26,6 +31,39 @@ ALIYUN = 'aliyun'
...
+    with open(f"builds/{build_id}/{arch}/meta.json") as meta_f:
+        meta = json.load(meta_f)
+    image_id = None
+
+    if cloud_id == AWS:
+        for region in meta['amis']:
+            if region['name'] == region_id:
+                image_id = region['hvm']
+
+    if cloud_id == ALIYUN:
```
cgwalters, line 57: Minor nit, I think this should be `else if` since the cases are mutually exclusive.

scripts/upload-ami
```diff
@@ -61,6 +99,40 @@ def upload_to_cloud(buildid, region, bucket, suffix=None, public=False):
...
+        arch = subprocess.check_output(["cosa", "basearch"], text='UTF-8').strip()
+        image_id = find_image_id(cloud, region, buildid, arch)
+        share_cmd = ['aliyun', '--endpoint', get_aliyun_endpoint(region), 'ecs',
+                      'ModifyImageSharePermission', '--region', region,
+                      '--ImageId', image_id]
+
+        # The aliyun CLI will return an error if you try to share an image with
+        # your own account, so this is a hack to prevent that from happening
+        # in a RHCOS devel pipeline scenario
+        aliyun_accts = [OPENSHIFT_ALIYUN_CI, OPENSHIFT_ALIYUN_DEV, OPENSHIFT_ALIYUN_QE]
```
cgwalters, line 114: I think it'd be more obvious to have this be a constant at the top; otherwise someone adding a new account e.g. `_STAGE` will also need to remember to hunt down and append to this list in the middle of the code.

scripts/upload-ami
```diff
@@ -61,6 +99,40 @@ def upload_to_cloud(buildid, region, bucket, suffix=None, public=False):
...
+        whoami_cmd = ['aliyun', 'sts', 'GetCallerIdentity']
+        out = subprocess.check_output(whoami_cmd, text='UTF-8')
+        whoami_dict = json.loads(out)
```
cgwalters, line 117:
```suggestion
        json.loads(subprocess.check_output(whoami_cmd, text='UTF-8'))
```

scripts/upload-ami
```diff
@@ -61,6 +99,40 @@ def upload_to_cloud(buildid, region, bucket, suffix=None, public=False):
...
+            if acct != whoami_id:
+                account_num += 1
+                share_cmd.extend(['--AddAccount.{}'.format(account_num), acct])
+
+        try:
+            print(share_cmd)
+            subprocess.check_call(share_cmd)
+        except subprocess.CalledProcessError as upload_err:
+            print("Failed to grant access to image! Error: {}".format(upload_err))
+            raise
```
cgwalters, line 134: I think *both* printing and re-throwing errors is an anti-pattern.  Instead perhaps:

```suggestion
            raise SystemExit(f"Failed to grant access to image: {upload_err}")
```

A bit like https://github.com/coreos/coreos-assembler/blob/6ac16f68b8842ca853a8b1dd41a2e38ee3095dac/src/cmd-build-validate#L50

cgwalters: (here and other places)

cgwalters (commented):
Only some nits, I didn't review the real logic in depth.

> Ideally, we'd have more of this in coreos-assembler, but this is (hopefully) a short-term bandaid.

I'm really skeptical of that...did https://github.com/openshift/os/pull/690 to try to explain this more

cgwalters:
/approve
though to be clear

miabbott:
> Only some nits, I didn't review the real logic in depth.
> 
> > Ideally, we'd have more of this in coreos-assembler, but this is (hopefully) a short-term bandaid.
> 
> I'm really skeptical of that...did #690 to try to explain this more

If I'm being really honest with myself, I probably should have done the work to land it in `cosa`, but I have a much easier time working with Python, so I went in this direction.

miabbott:
/hold

I'm exploring putting the public functionality into `ore`

miabbott:
/close

See: coreos/coreos-assembler#2648, coreos/coreos-assembler#2662, coreos/coreos-assembler#2670

PR #689, opened 2021-12-17
scripts: add utility for marking Aliyun images public

miabbott:
Frustratingly, replicated copies of a public image in Aliyun do not
inherit the publicly available properties of the original image. So we
have to go back and mark the images as public after they have been
replicated.
    
I opted for a smaller utility script for this, rather than trying to
overload the already ballooning `upload-ami` script.
    
Requires openshift/os#688

### DIFF
--- a/scripts/make_aliyun_public
+++ b/scripts/make_aliyun_public
@@ -1,0 +1,69 @@
+#!/usr/bin/python3 -u
+
+# This is a utility script to mark all the Aliyun images that have been
+# replicated across regions as publicly available images. It assumes that
+# meta.json has been updated with replicated regions + image IDs. It also
+# assumes that you want to make all images in all regions public.
+
+import argparse
+import json
+import subprocess
+import sys
+
+from upload_ami import get_aliyun_endpoint
+
+ALIYUN = 'aliyun'
+cloud = ALIYUN
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


scripts/make_aliyun_public
```diff
@@ -0,0 +1,69 @@
...
+
+ALIYUN = 'aliyun'
+cloud = ALIYUN
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--build', help="Build ID to use")
+    args = parser.parse_args()
+
+    arch = subprocess.check_output(["cosa", "basearch"], text='UTF-8').strip()
```
cgwalters, line 23: But since this is a new script that depends on coreos-assembler, we could just put it in cosa right now right?  `cosa aliyun-make-public` or so?

miabbott: Yeah! That makes sense, I can move this functionality over there.

miabbott:
/hold

Going to pursue moving this to `cosa`

miabbott:
/close

See:  coreos/coreos-assembler#2648,  coreos/coreos-assembler#2662,  coreos/coreos-assembler#2670

miabbott:
/close

See: coreos/coreos-assembler#2648, coreos/coreos-assembler#2662, coreos/coreos-assembler#2670

PR #690, opened 2021-12-17
upload-ami: Extend blurb about how things should go into coreos-assembler

cgwalters:
Motivated by recent Aliyun additions; adding stuff like that here
makes it harder to do Aliyun stuff with FCOS *and* breaks the model
that this repo is the definition of the OS, not the pipeline to
build the OS.

### DIFF
--- a/scripts/upload-ami
+++ b/scripts/upload-ami
@@ -1,8 +1,13 @@
 #!/usr/bin/python3 -u
 
-# Wraps coreos-assembler buildextend-aws which itself wraps `ore aws upload`,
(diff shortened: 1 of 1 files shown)


cgwalters:
The other big thing we actually should do I think is only make AMIs or equivalent public *at the time we go to bump the installer bootimage*.  That would drop the need for "make things public" logic in this script and the RHCOS pipeline by default.  (And a "make all the AMIs public" logic could go in coreos-assembler too, factored out from plume if it's doing that)

jlebon:
/lgtm

Folding this whole script into cosa could be a nice project for someone wanting to get familiar with it.

PR #692, opened 2022-01-14
Use new `overlay.d/cosa-no-autolayer` option

jlebon:
Tell coreos-assembler to not automatically add the overlays to the final
manifest. Add references to the auto-generated OSTree layers in our
manifest.

This isn't strictly necessary for RHCOS like it is for FCOS because the
branching model is simpler here. But doing this will allow us to not
support it both ways in cosa and make this the default behaviour so we
can drop the stamp file.

Requires: https://github.com/coreos/coreos-assembler/pull/2641

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -14,6 +14,17 @@ include:
   # RHCOS owned packages
   - rhcos-packages.yaml
 
+ostree-layers:
+  - overlay/01fcos
+  - overlay/02fcos-nouveau
+  - overlay/05rhcos
+  - overlay/06gcp-routes
+  - overlay/15rhcos-logrotate
+  - overlay/15rhcos-tuned-bits
+  - overlay/20platform-chrony
+  - overlay/21dhcp-chrony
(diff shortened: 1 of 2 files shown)


jlebon:
```
 Problem 1: conflicting requests
  - package crun-1.4.1-0.rhaos4.10.el8.2.aarch64 does not have a compatible architecture
  - nothing provides libdl.so.2(GLIBC_2.17)(64bit) needed by crun-1.4.1-0.rhaos4.10.el8.2.aarch64
  - nothing provides ld-linux-aarch64.so.1()(64bit) needed by crun-1.4.1-0.rhaos4.10.el8.2.aarch64
  - nothing provides ld-linux-aarch64.so.1(GLIBC_2.17)(64bit) needed by crun-1.4.1-0.rhaos4.10.el8.2.aarch64
```

Hmm, interesting. Seems like we might be missing x86_64 cri-o from the repos?

dustymabe (approved):
LGTM

PR #693, opened 2022-01-19
tests/kola/version/toolbox: check we have the correct toolbox version

miabbott:
We want to keep shipping the version of `toolbox` from
https://github.com/coreos/toolbox in RHCOS for the time being. But it
is easy for the RHEL dist-git of `toolbox` to get updated without us
being notified, so let's put a simple check in place to make sure we
are getting the bash script version and not the golang version.

### DIFF
--- a/tests/kola/version/toolbox
+++ b/tests/kola/version/toolbox
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# This is RHCOS only, we ship a different version of toolbox in FCOS
+
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# make it a var so we can change when/if we switch to containers/toolbox
+url="https://github.com/coreos/toolbox"
+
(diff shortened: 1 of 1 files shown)


tests/kola/version/toolbox
```diff
@@ -0,0 +1,15 @@
...
+
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+if [ "$(file -b /usr/bin/toolbox)" != "Bourne-Again shell script, ASCII text executable" ]; then
```
jlebon, line 13: This is fine though if you'd like, another approach that gets to the heart of it is to look at the URL, e.g.:

```sh
if [ "$(rpm -q --qf '%{URL}' toolbox)" != "https://github.com/coreos/toolbox" ]; then
```

jlebon (approved):

tests/kola/version/toolbox
miabbott: Ooh, I like this better.  Thanks Jonathan!

PR #695, opened 2022-01-26
overlay.d: delete `cosa-no-autolayer` knob file

jlebon:
It's the default behaviour now:
https://github.com/coreos/coreos-assembler/pull/2665


jlebon:
This is the final step in https://github.com/coreos/coreos-assembler/pull/2641#issuecomment-1013285699.

jlebon:
Let's wait a day or two on this to be nice because it's another cosa bump required on local developer machines and the stamp file doesn't really hurt there until then.

jlebon:
/retest
/hold cancel

This should be good to go now!

sohankunkerkar (approved):

PR #696, opened 2022-01-26
manifest.yaml: document pinned container-tools packages

jlebon:
For the container stack, we currently default to obtaining everything
from `container-tools:rhel8` but there are some exceptions. Document
them as per:

https://github.com/openshift/os/pull/681#issuecomment-1022443830


jlebon:
If someone has handy links with more information for each exception there, I'll add them.

PR #700, opened 2022-02-02
ci: add same tests as downstream pipeline

gursewak1997:
Expand CI jobs to run same tests as downstream pipeline so we get more feedback early and often.

Jira: https://issues.redhat.com/browse/COS-793

### DIFF
--- a/ci/prow-build-test-qemu.sh
+++ b/ci/prow-build-test-qemu.sh
@@ -38,4 +38,19 @@ cosa buildextend-extensions
 cosa kola --basic-qemu-scenarios
 kola run-upgrade -b rhcos -v --find-parent-image --qemu-image-dir tmp/ --output-dir tmp/kola-upgrade
 cosa kola run --parallel 2
-
+# Build metal + installer now so we can test them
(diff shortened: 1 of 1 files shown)


cgwalters:
I was a bit concerned about the speed hit, but it's not as bad as I thought because our baseline actually appears to be [two hours, looking at a previous PR](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_os/699/pull-ci-openshift-os-master-build-test-qemu/1488624098824163328).  This took 2h49m, so a 40% increase.  A fair amount, but not the end of the world.

That said:

- Having it take nearly 3 hours to merge a single PR is quite slow
- What's a bit more of a problem is that *most* changes here are unlikely to break those tests.  I think what we really want is to do lockfiles, which would mean that e.g. bumping stuff like kernel, coreos-installer, rpm-ostree, ignition would gate on these tests.  And *those* things are much more likely to break our CI.
- CI here is only running on one architecture; extending to aarch64 seems quite tractable but currently conflicts a bit with our GCP nested virt baseline.
-

PR #702, opened 2022-02-02
ci: delete unneeded arch checks

gursewak1997:
Since we run this job for only x86_64, we can delete these checks


cgwalters:
Well...yes, today.  But it's *possible* for us to run this same CI test on aarch64 in theory.

I think if anything I'd argue that it makes sense instead to move the architecture conditionals into coreos-assembler.

Not opposed to this, but I am not sure it's worth the code cleanup to delete the architecture checks.

gursewak1997:
> Not opposed to this, but I am not sure it's worth the code cleanup to delete the architecture checks.

Right; There's a few tests(`kola run-upgrade` for instance) in this file which would need arch checks if we do decide to run it fort aarch64 as an example so I assumed I'd just clean up till we make all those changes.

I'm okay with keeping the checks if you think it's not worth it:)

PR #705, opened 2022-02-08
packages: Install NetworkManager-config-server

qinqon:
There is a bug [1] related to kubernetes-nmstate since nmstate does not
support multipath. This change add the mentioned RPM to fix the issue.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=2048988#c15

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -227,6 +227,7 @@ packages:
  - openvswitch2.15
  - dnsmasq
  - NetworkManager-ovs
+ - NetworkManager-config-server
  # Extra runtime
  - sssd
  # Common tools used by scripts and admins interactively


cgwalters:
First, in general we'd prefer to keep RHCOS in sync with FCOS - so this should instead be a PR to https://github.com/coreos/fedora-coreos-config/

But there's something a lot more profound going on here, which is basically that this would (AFAIK) break the cloud default of doing DHCP, right?

I was involved in the discussions long ago behind the creation of `NetworkManager-config-server`.  The original motivation was precisely that for desktop/laptop systems, one wants plugging in an ethernet cord to do DHCP.  For bare metal servers, it's mixed - many environments are static addressing only (which we also support).

But the cloud case (as well as many IoT scenarios) circle back to the DHCP case.

And CoreOS spans bare metal and cloud - and we try to keep things aligned.

Broadly speaking, I don't think we can make this change now.  But we could recommend doing this for bare metal installs via Ignition/MachineConfig.

cgwalters:
> But we are talking about FCOS/RHCOS they are suppose to be targeting servers isn't it ?

I already covered the "servers that use DHCP" case.

> Do you know what is the proper way to include it there so it's present at the vanilla openshift clusters ?

I think we basically just need to update our bare metal installation documentation to recommend doing this via Ignition/MachineConfig, particularly if one is using static IP addressing.

But there are also bare metal installations that use DHCP - and we'd also break them when upgrading if we applied this change as is!

cgwalters:
```
 === RUN   basic
--- FAIL: basic (601.62s)
        harness.go:101: TIMEOUT[10m0s]: SSH unsuccessful within allotted timeframe for ab810b8f-6005-4ae2-a5ba-c482de1e4635. 
```

Yep, there's our CI successfully failing!  :tada: 

So...closing this because we simply can't make the proposed change.  I think though the team would be happy to help with any PRs to the FCOS and OCP documentation.  I think if we added some sort of high level sugar for networking to e.g. [butane](https://github.com/coreos/butane/) this would clearly be nice as a flag like:

```
networking:
  no-auto-default: true
```
or so.

PR #714, opened 2022-02-10
Bump fedora-coreos-config

cgwalters:
Specifically to pull in https://github.com/coreos/fedora-coreos-config/pull/1511

```
Colin Walters (1):
      tests/default-network-behavior-change: Adapt to RHCOS 4.11

CoreOS Bot (1):
      lockfiles: bump to latest

Dusty Mabe (2):
      denylist: snooze ext.config.chrony.dhcp-propagation in rawhide
      denylist: apply snoozes to the branched stream as well

Huijing Hei (1):
      denylist: enable coreos.boot-mirror.luks test
```


cgwalters:
Ahhh right, hah deadlock - this one depends on the 8.5 bump which is only internal.

Hmm.  We need to have $rhelminor source-of-truth be this repo.

cgwalters:
Ohhh right.  The problem is actually that we need to explicitly separate the RHEL from OCP version, but the repo files in CI merge those together.  

So...we need to add into release-controller e.g. `ocp-4.11-rhel8.5.repo` or so.  And fix things actually so that source-of-truth for the minor version is *this* repo, and not duplicated between this and downstream.

cgwalters:
https://github.com/openshift/release/pull/26193

Issue #498, opened 2021-02-08
proposal: generate "base rhel" container image, build OCP on top


cgwalters:
Now that rpm-ostree is close to supporting "live updates", one thing we could do is move crio/kubelet into a separate `machine-os-kubelet` container or so, and also move `openvswitch` as part of e.g. the SDN container.

But these would still be treated as "first class" bits because they'd still be underneath the readonly bind mount in `/usr` etc.  The MCO would learn to pull down this `machine-os-kubelet` container and apply updates from it too; and we can generalize that to N container images with M RPMs inside (or...perhaps not RPMs at all).

Advantages:

The RHCOS bootimage is basically just RHEL, and this would greatly increase alignment with OKD since we'd use the same approach in both places.

On the bootstrap node, the crio/kubelet in use become exactly the same as the one shipped in cluster.

There wouldn't be anymore "CI -> shipping" gap for kubelet - when a PR merges to that repo it'd get rebuilt and shipped the same way all other containers do and not versioned with RHCOS at all.

Note in this we wouldn't be breaking at all the concept that the cluster owns and manages OS updates; we'd still be testing the OS and kubelet and cluster components all together as a unit in the end.  The goal here is just *internally* split things up more so we can improve the process for CI and building; for example, the RHCOS version number would (mostly) just be a RHEL version number which would greatly increase clarity of how things work.  We can be more agile with kubelet/crio etc.

ashcrow:
I like the idea! It sounds like decoupling into a few classes of streams to make bootstrapping and CI testing easier to manager.

1. boot strap _and_ cluster (requirements to bootstrap and to run a cluster)
2. general cluster (more general OS level)

One question is how would we tie these container images together? For example, if the `machine-os-kubelet` needed to bump and `machine-os-content` didn't change. Would they be combined by container tags or through the payload manifest at the higher level? etc..

cgwalters:
The `machine-os-kubelet` would just be another payload image, it would be built with a `Dockerfile` the same way as everything else in the cluster.  The MCO would know how to pull it down and apply it.  During cluster bootstrap, `bootkube.sh` already downloads the MCO container, so we'd extend that to call the MCO to also extract crio/kubelet.  For upgrades the MCO would also extract and apply it the same way it does `machine-os-content`.

IOW the end goal here is that the lifecycle of this container is logically separate; either container can change independently without caring about the other, we just merge the result.

cgwalters:
[...]

PR #715, opened 2022-02-14
manifest: bump to openvswitch2.16

dcbw:
We want to keep host OVS and container OVS synchronized as much
as possible, plus OVS 2.16 is the first release that can take
advantage of kernel features like per-cpu dispatch (if the kernel
has it).

@trozet

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -224,7 +224,7 @@ packages:
  - cri-o cri-tools
  # Networking
  - nfs-utils
- - openvswitch2.15
+ - openvswitch2.16
  - dnsmasq
  - NetworkManager-ovs
  # Extra runtime


jlebon:
Related: https://github.com/openshift/os/issues/498

Meanwhile, if you have a checklist somewhere for rebasing Open vSwitch in the SDN container, it might be good to add an item for bumping it here too.

/approve

cgwalters (approved):
I'm open to making this switch.  Ooh: better: I'm open to shifting the switch.

dcbw:
> Meanwhile, if you have a checklist somewhere for rebasing Open vSwitch in the SDN container, it might be good to add an item for bumping it here too.

@jlebon no checklist other than (a) has it been through normal RHEL FDP QE? and (b) does it pass CI?

cgwalters:
One thing I'll note here that's important: CI here isn't going to test this in any meaningful way, i.e. we aren't yet today launching a cluster from this build.  Hopefully soon we will be though!

miabbott:
I believe the CI failed due to https://github.com/openshift/os/pull/718 + https://github.com/openshift/release/pull/26305, but I built and tested this locally with success.

PR #724, opened 2022-02-23
docs: add instructions for using cosa in app.ci

miabbott:
I found myself doing this today and figured it was useful for others.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -11,14 +11,15 @@ just like [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config)
 is for FCOS.
 
 For example, you would use
-```
+
+```bash
 $ cosa init https://github.com/openshift/os
 ```
 
 to start.  However, currently you need to configure
 the rpm-md repositories manually:
 
-# Repositories
+## Repositories
 
 The RHEL repositories are only available to Red Hat customers,
 and it's likely that you want to use a mirror.  And further,
@@ -28,7 +29,7 @@ you need to provide the repos of input RPMs.
 
 To do so, create a `rhcos.repo` file that looks like this:
 
-```
+```text
 # RHEL repos
 [rhel-8-baseos]
 baseurl=<url>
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -69,7 +58,68 @@ of FCOS.
...
 
-```
+```bash
 $ (cd fedora-coreos-config && git fetch origin && git reset --hard origin/testing-devel)
 $ git commit -a -m 'Update FCOS'
 ```
+
+## Debugging failures in app.ci
+
+Occasionally it is necesasry to debug test failures on the [OpenShift CI clusters](https://docs.ci.openshift.org/docs/getting-started/useful-links/#clusters)
```
jlebon, line 68:
```suggestion
Occasionally it is necessary to debug test failures on the [OpenShift CI clusters](https://docs.ci.openshift.org/docs/getting-started/useful-links/#clusters)
```

docs/development.md
```diff
@@ -69,7 +58,68 @@ of FCOS.
...
+$ umask 0022
+```
+
+1. Init the repo
+
+`$ cosa init https://github.com/openshift/os`
+
+1. Populate the repo files
+
+Determine which repo the failing CI job is using and replicate it. For example:
```
jlebon, line 121: This step could use more information.

docs/development.md
```diff
@@ -69,7 +58,68 @@ of FCOS.
...
+metadata:
+  # Replace this with e.g. $USER-cosa
+  name: test-cosa
+spec:
+  containers:
+  - args:
+    - shell
+    - sleep
+    - infinity
+    image: quay.io/coreos-assembler/coreos-assembler:latest
```
jlebon, line 88: Heh, the funny thing is that now this is actually built in app.ci itself. This is fine, but it might be better to use that FQIN instead. That reduces the chance of using a different image than what the tests are since Quay.io is delayed.

miabbott:
Thanks for the review, Jonathan!  Pushed an update :arrow_up:

jlebon (approved):
Cool, this is helpful!

PR #524, opened 2021-03-23
overlay: Add 45-rhcos-extensions.preset

cgwalters:
This is part of https://github.com/openshift/enhancements/pull/677

https://bugzilla.redhat.com/show_bug.cgi?id=1941342

PR #725, opened 2022-02-24
Verify kata-osbuilder-generate.service as part of the default presets

HuijingHei:
See:
- https://bugzilla.redhat.com/show_bug.cgi?id=1941342
- https://github.com/openshift/os/pull/524/files


miabbott:
The new test passed and doesn't affect other tests, so I'm fine to override the CI job

/lgtm

PR #726, opened 2022-02-24
Remove FDP puddle for aarch64

andymcc:
Now that FDP is shipped for aarch64 we should not be using the FDP
puddle for Arm.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -28,7 +28,6 @@ repos:
   - rhel-8-baseos
   - rhel-8-appstream
   - rhel-8-fast-datapath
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

yselkowitz (approved):
While openvswitch2.16 will not be shipped until the next batch, RHCOS 4.11 ends up using the version cross-tagged into OCP anyway.  Therefore, this can go in now, and backported to 4.10 and 4.9.
/lgtm

saqibali-2k (approved):

PR #733, opened 2022-03-02
s390x: Temporary fix for cpictl permissions

jschintag:
Fix file permission for cpictl until fix is backported to RHEL 8.6
https://bugzilla.redhat.com/show_bug.cgi?id=2024102

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -97,6 +97,13 @@ postprocess:
      EOF
      fi
 
+     # TEMPORARY: Fix file permission for cpictl until fix is backported to RHEL 8.6
+     # See https://bugzilla.redhat.com/show_bug.cgi?id=2024102
+     if [ "$(uname -m)" == "s390x" ]; then
+     [ "$(stat -c '%a' /usr/lib/s390-tools/cpictl)" == "755" ] && echo "Permission for /usr/lib/s390-tools/cpictl is fixed, remove temporary hack"
+     chmod 755 /usr/lib/s390-tools/cpictl
+     fi
+
      # Nuke network.service from orbit
      # https://github.com/openshift/os/issues/117
      rm -rf /etc/rc.d/init.d/network /etc/rc.d/rc*.d/*network


miabbott:
@jschintag do we have any way of tracking the progress of the backport?

jschintag:
@micah Good point, we should probably have a rhcos issue/bugzilla to track the progress and reference that in the denylist instead. Does it make more sense to do a issue or a bugzilla?

miabbott:
> @micah Good point, we should probably have a rhcos issue/bugzilla to track the progress and reference that in the denylist instead. Does it make more sense to do a issue or a bugzilla?

Following the breadcrumbs, it looks `s390utils` was fixed as part of https://bugzilla.redhat.com/show_bug.cgi?id=2024102

We should request a backport to 8.6 from that BZ and see if it is granted.

cgwalters:
I think I'd be more in favor of just adding a `chmod u-w` in our postprocess scripts for now and a reminder to ourselves to remove it later.

miabbott:
> I think I'd be more in favor of just adding a `chmod u-w` in our postprocess scripts for now and a reminder to ourselves to remove it later.

Great idea!   @jschintag do you want to handle that?

cgwalters:
One thing to bear in mind is that while the original BZ around group write permissions is valid as a general rule, the ostree read-only bind mount would prevent any attempt by users in the `root` group to actually write, and I suspect SELinux would too.

jschintag:
> > I think I'd be more in favor of just adding a `chmod u-w` in our postprocess scripts for now and a reminder to ourselves to remove it later.
> 
> Great idea! @jschintag do you want to handle that?

@miabbott Sure, that sounds like an even better way to unblock s390x for now. I'll work on that today. And reuse this PR. Waste not want not or so.


> One thing to bear in mind is that while the original BZ around group write permissions is valid as a general rule, the ostree read-only bind mount would prevent any attempt by users in the `root` group to actually write, and I suspect SELinux would too.

@cgwalters If the file permissions don't actually matter, i would question why we have a test dedicated to making sure they are okay?

jschintag:
I switched this PR from snoozing to hacking/fixing.

micah:
On 2022-03-02 07:35:46, Jan Schintag wrote:
> @micah Good point, we should probably have a rhcos issue/bugzilla to
> track the progress and reference that in the denylist instead. Does it
> make more sense to do a issue or a bugzilla?

FYI - @micah is not Micah Abbott, who you are trying to get.

jschintag:
Ups, sorry.

cgwalters:
> @cgwalters If the file permissions don't actually matter, i would question why we have a test dedicated to making sure they are okay?

One reason is that these permissions actually matter a lot in a default OpenShift *container*, because for historical reasons the unprivileged user is in the root group.  In that scenario, there's no ostree read-only bind mount and SELinux only applies outside the container.

There are also people using traditional RHEL host systems (i.e. no ostree read only bind mount) and who have disabled SELinux (to claw back 1-2% of performance or other reasons) and then both those protective layers are stripped away.

jschintag:
Thanks for the background information @cgwalters

manifest.yaml
```diff
@@ -97,6 +97,12 @@ postprocess:
      EOF
      fi
 
+     # TEMPORARY: Fix file permission for cpictl until fix is backported to RHEL 8.6
+     # See https://bugzilla.redhat.com/show_bug.cgi?id=2024102
+     if [ "$(uname -m)" == "s390x" ]; then
+     chmod 755 /usr/lib/s390-tools/cpictl
+     fi
+
```
miabbott, line 105: I'd suggest enhancing this to check if the permissions are set correctly before applying this workaround. Maybe emit a message too, so we can know when we can remove the workaround.

jschintag: I added a check for this. I would also like to reserve the right to miss that message in between all the output of a `cosa build` unless i specifically search for it.

PR #735, opened 2022-03-02
honor COSA_DIR env var

cheesesashimi:
To make Prow CI a bit more friendly, this script should honor the `$COSA_DIR` env var so that the dir can be specified from outside of this script.


cgwalters (approved):

cheesesashimi:
Well that was embarrassing! Since we're using `set -u` in this script, `$COSA_DIR` is unbound (unless explicitly set in the environment). I changed this to use a [default expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) instead which has the same effect.

PR #739, opened 2022-03-04
denylist: Skip coreos.boot-mirror and coreos.boot-mirror.luks on ppc64le

ravanelli:
- Looks the denylist works different than I thought/tested.
 It won't skip coreos.boot-mirror.luks/detach-primary and coreos.boot-mirror/detach-primary
 since they are called by coreos.boot-mirror tests, even though it shows a message as it would be skipped,
 they are only skipped if they are called separately. We probably have some area for improvements here.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -25,11 +25,11 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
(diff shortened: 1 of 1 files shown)


jlebon:
The reason why it doesn't work is because `detach-primary` is a subtest so it can't be skipped alone. (The non-exclusive tests work makes this more confusing because tests in that case do become subtests.)

/lgtm

PR #740, opened 2022-03-07
extensions: Remove references to qemu-kiwi and advanced-virt

etrunko:
From RHEL 8.6 onwards, there is no qemu-kiwi package nor advanced
virtualization reposiory. The kata-containers package has been adjusted
to require the qemu-kvm-core pacakge instead.

Signed-off-by: Eduardo Lima (Etrunko) <etrunko@redhat.com>


jlebon:
/ok-to-test
/lgtm

Cool, that's a nice cleanup. Let's also delete the repo file from the internal redhat-coreos repo.

miabbott:
``` 
Problem: conflicting requests
  - nothing provides qemu-kiwi >= 5.1.0-16 needed by kata-containers-2.3.2-2.el8.x86_64
```
Looks like we need an update to the repos defined in `openshift/release` to point to RHEL 8.6

cgwalters:
> Looks like we need an update to the repos defined in openshift/release to point to RHEL 8.6

I think we have a catch-22 here, we can't do that until it's released, right?  I think we need to handle this by branching/versioning.  We need to *add* rhel8.6 as an option, not replace the current version.  And we should make a `main-8.6` branch here, where this patch could merge.

miabbott:
> > Looks like we need an update to the repos defined in openshift/release to point to RHEL 8.6
> 
> I think we have a catch-22 here, we can't do that until it's released, right? I think we need to handle this by branching/versioning. We need to _add_ rhel8.6 as an option, not replace the current version. And we should make a `main-8.6` branch here, where this patch could merge.

Correct, we don't have the ability to use any 8.6 content yet (due to legal reasons), so a new branch sounds like a reasonable compromise.

etrunko:
> > > Looks like we need an update to the repos defined in openshift/release to point to RHEL 8.6
> > 
> > 
> > I think we have a catch-22 here, we can't do that until it's released, right? I think we need to handle this by branching/versioning. We need to _add_ rhel8.6 as an option, not replace the current version. And we should make a `main-8.6` branch here, where this patch could merge.
> 
> Correct, we don't have the ability to use any 8.6 content yet (due to legal reasons), so a new branch sounds like a reasonable compromise.

Alright, I don't have any experience with this, so can you please provide more information on how should I proceed here?

cgwalters:
/hold
until we can set things up so that CI can work and this can merge.

> Alright, I don't have any experience with this, so can you please provide more information on how should I proceed here?

No worries, it's not your fault or task, it's more on the RHCOS team to figure out.  This is just the first PR that requires 8.6.

HuijingHei:
The error logs:
```
Importing rpm-md...done
rpm-md repo 'rhel-8-baseos'; generated: 2022-03-23T18:28:26Z solvables: 1718
rpm-md repo 'rhel-8-appstream'; generated: 2022-03-23T18:29:24Z solvables: 6603
rpm-md repo 'rhel-8-server-ose'; generated: 2022-03-31T01:08:29Z solvables: 491
rpm-md repo 'rhel-8-fast-datapath'; generated: 2022-03-30T16:21:12Z solvables: 705
rpm-md repo 'rhel-8-nfv'; generated: 2022-03-23T18:27:47Z solvables: 20
Resolving dependencies...done
 Problem: package kata-containers-2.3.3-3.el8.x86_64 requires qemu-kvm-core >= 6.2.0, but none of the providers can be installed
  - conflicting requests
  - package qemu-kvm-core-15:6.2.0-8.module+el8.6.0+14324+050a5215.x86_64 is filtered out by modular filtering
```

HuijingHei:
A little confused about the failed logs, qemu-kvm-core is included in 'rhel-8-appstream', does this mean we should add `rhel-8-appstream` here?

miabbott:
I think we need to keep the `virt:av` module enabled here, since `qemu-kvm-core` is part of the `virt` module.

But modules continue to confuse me to this day, so I could be wrong. :shrug:

etrunko:
Yes, it is confusing, the module still exists, as it is the way the virtualization packages are built with RHEL 8. This is not going to change until RHEL 9.

What changed from 8.4 to 8.6 is that the same version of the virt packages are available in both AV and regular RHEL repositories. So we do not require the AV repository to be enabled. Do you know if there is a call to `dnf module enable virt:av` somewhere?

etrunko:
> A little confused about the failed logs, qemu-kvm-core is included in 'rhel-8-appstream', does this mean we should add `rhel-8-appstream` here?

I think this is what we need.

cgwalters:
Hm, did we lose having build-test-qemu be a gating context?

PR #745, opened 2022-03-09
denylist: skip multipath.partition on ppc64le

miabbott:
This test has been rather consistently failing since the latest RHEL
8.4.z update.


miabbott:
/hold

Not sure if we want to use this hammer yet

PR #746, opened 2022-03-14
adds openshift image derivation test

cheesesashimi:
This PR introduces the OpenShift in-cluster image derivation test. What this does is introduce a new test which will test that in-cluster image derivation works as we intend. The test works thusly:

1. We grab the latest RHCOS image which was built just prior to test execution.
2. Build a derived image from the aforementioned RHCOS image.
3. Select a cluster node to apply the image to.
4. Apply the new image using `rpm-ostree`.
5. Reboot the cluster node.
6. Verify that the cluster node came up successfully.
7. Roll back applying the derived image.

This is a deceptively large PR because it introduces a bunch of vendored dependencies. For ease of review, I added the vendored dependencies in a commit by themselves. Additionally, since I wasn't sure about the impact my changes would have upon the `coreos-assembler` image, I opted to multistage the Dockerfile build so I could inject a pre-compiled version of the layering test suite into the `coreos-assembler` image.

The test itself originally came from https://github.com/openshift/machine-config-operator/pull/2886.


tests/layering/layering_test.go
```diff
@@ -0,0 +1,207 @@
...
+	dockerConfigJSON := credentialprovider.DockerConfigJSON{
+		Auths: dockerConfig,
+	}
+	authfileData, err := json.Marshal(dockerConfigJSON)
+	require.Nil(t, err)
+	helpers.ExecCmdOnNode(t, cs, infraNode, "mkdir", "-p", path.Dir(path.Join("/rootfs", authfilePath)))
+	// will get cleaned up on reboot since file is in /run
+	writeToMCDContainer(t, cs, infraNode, path.Join("/rootfs", authfilePath), authfileData)
+
+	// rpm-ostree rebase --experimental ostree-unverified-image:docker://image-registry.openshift-image-registry.svc.cluster.local:5000/openshift-machine-config-operator/test-boot-in-cluster-image-build
```
mkenigs, line 118:
```suggestion
	// rpm-ostree rebase --experimental ostree-unverified-image:image-registry.openshift-image-registry.svc:5000/openshift-machine-config-operator/test-boot-in-cluster-image
```
or maybe just
```suggestion
	// rpm-ostree rebase --experimental imageURL
```
to avoid similar inconsistencies in the future

jlebon:
Cool!

> The test itself originally came from https://github.com/openshift/machine-config-operator/pull/2886.

Is the test exactly the same? Could we instead of adding go vendoring here, have the builder also check out the MCO repo and build the test from there?

tests/layering/layering_test.go
cheesesashimi: No worries! I'm cleaning this up as I go.

cgwalters:
So...this is definitely tricky.  As of right now, this repository is only consumed by coreos-assembler - which already has a lot of Go code for tests.  Adding new Go code here means there's another place to handle vendor bumps etc.

Another balance we try to handle though is that we don't have a hard dependency for "coreos stuff" on OpenShift or even Kubernetes.  (e.g. coreos-assembler can also be run via podman)

WDYT about having this in e.g. https://github.com/openshift/origin/tree/master/test/extended instead?  I think we can wire things up so that the new test is only run on this repo to start - but it would mean we could also more easily run it as a periodic *outside* of PRs to this repo?

That said, on the other hand there *are* advantages to having tests live with the repository.  A big part of the reason [coreos external tests](https://github.com/coreos/coreos-assembler/blob/main/docs/kola/external-tests.md) were created is that the tests can live inside the config repo (like this one) that also (for FCOS) also versions the content.

I think I lean on balance towards adding this to the openshift tests, but it's by no means a strongly held opinion.  Another counterbalance could be adding *more* tests in Go code here.

cgwalters:
Or I guess to say this another way what has me hesitating a bit is that today the output from a build of this repository is a "cosa build", which is the oscontainer image and disk images.

Now we *do* have a Dockerfile, but that's mainly to "fake out" Prow and try to get "a build" of stuff together - we don't actually ship the result of that build.

But if we're building a test, I think there are definitely cases where we'd want to build that test *outside* of building the repo - which is how the origin tests work.

If we were to go this path, perhaps actually it should be [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) that has a Go testing framework, and this adds RHCOS tests to it?  That would be a new middle ground between "external tests are in shell script" and "cosa tests which can be Go, but are lifecycled with cosa, not the OS".


Anyone else have opinions?

jlebon:
Personally, I don't really mind Go code being used for tests in this repo or in f-c-c. I think  we've discussed in the past re. how if we want to keep draining kola of tests into the config repos, we'd probably need "cosa build"-level testing there to make `kola testiso` work in that model, and I can see Go being good for this for sharing with cosa.

I really dislike the Go vendoring model though. Maybe if we go that way, we can find some way to use the vendoring from cosa and not ship any vendoring in the config repos (which might be what you're implying already with "lifecycled with cosa").

But re. this test in specific, since this is a cluster-level test (which unlike kola tests, isn't really something developers are expected to easily run locally, right?) and OpenShift CI falls over anyway if GitHub is down, could we just ship `go.sum` and `go.mod`? I think that'd make this much easier to swallow.

cheesesashimi:
I'm torn between this belonging here or in openshift/origin since this is a cluster-level test that tests interaction with RHCOS and various parts of OpenShift. I see benefits and drawbacks for both of those repos and I'm not particularly opinionated with it going in either of those places.

However, I'm not sure if it belongs with coreos-assembler or even fedora-coreos-config since those seem like they're more OS-specific tests that exercise and validate the various aspects of CoreOS. But I do see the value of vendoring the dependencies for this test within either of those repos. That said, I share @jlebon's dislike of the Go vendor model and think we can get away with only the `go.mod` / `go.sum` files in this repo.

jlebon:
Also good with this as is.
/approve

Minor suggestion to layout: add a `tests/go` subdir and in there have the `go.mod` and `go.sum` files and the `layering/` subdir.

cheesesashimi:
This is now ready for a more in-depth review.

tests/layering/layering_test.go
```diff
@@ -0,0 +1,148 @@
...
+	// ostreeUnverifiedRegistry means no GPG or container signatures are used.
+	// Right now we're usually relying on digested pulls. See
+	// https://github.com/openshift/machine-config-operator/blob/master/docs/OSUpgrades.md#questions-and-answers around integrity.
+	// See https://docs.rs/ostree-ext/0.5.1/ostree_ext/container/struct.OstreeImageReference.html
+	ostreeUnverifiedRegistry = "ostree-unverified-registry"
+	imageRegistry            = "image-registry.openshift-image-registry.svc:5000"
+	// If this moves from /run, make sure files get cleaned up
+	authfilePath   = "/run/ostree/auth.json"
+	mcoNamespace   = "openshift-machine-config-operator"
+	imagePullSpec  = "registry.ci.openshift.org/rhcos-devel/rhel-coreos:%s"
```
kikisdeliveryservice, line 30: q: why %s here?

tests/layering/node_cmd_test.go
```diff
@@ -0,0 +1,436 @@
...
+		}
+
+		// This is nil, meaning we haven't started yet
+		if status.Started == nil {
+			return false
+		}
+
+		// We haven't started the container yet
+		if *status.Started != true {
+			return false
```
kikisdeliveryservice, line 425: q: probably obvious but why a pointer?

tests/layering/node_cmd_test.go
```diff
@@ -0,0 +1,436 @@
...
+
+		if condition.Type == corev1.PodScheduled && condition.Status != "True" {
+			return false
+		}
+
+		if condition.Type == corev1.PodReady && condition.Status != "True" {
+			return false
+		}
+
+		if condition.Type == corev1.ContainersReady && condition.Status != "True" {
```
kikisdeliveryservice, line 396: q: if containers are ready and condition.status == true can the below ever be valid checks? (I think I'm asking: is containersReady a proxy for the check below??)

tests/layering/layering_test.go
```diff
@@ -0,0 +1,148 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var deleteBuild = flag.Bool("delete-build", false, "Delete the derived build at the end of the test.")
+var deleteMachine = flag.Bool("delete-machine", false, "Delete the target machine after test run.")
+var streamBuild = flag.Bool("stream-build", false, "Stream the derived image build to stdout")
+var buildLogFile = flag.String("build-log", "", "Filename to write the derived image build log to")
+
+const (
+	imageStreamName = "test-boot-in-cluster-image"
```
kikisdeliveryservice, line 19: nit: could we alphabetize these so they are grouped together a little more

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+
+	// Wait for the build to complete
+	return waitForBuildToRun(t, cs, build)
+}
+
+// Puts the pull secret for the MCD service account on the underlying node so rpm-ostree can use it.
+// get ImagePullSecret for the MCD service account and save to authfilePath on the node
+// eventually we should use rest.InClusterConfig() instead of cs with kubeadmin
+func putPullSecretOnNode(t *testing.T, cs *framework.ClientSet, node *corev1.Node) func() {
+	t.Logf("placing image pull secret on node")
```
kikisdeliveryservice, line 135:
```suggestion
	t.Log("placing image pull secret on node")
```

tests/layering/node_cmd_test.go
cheesesashimi: Honestly, I wasn't sure what I needed to check and went a bit overboard. Just this check alone might be sufficient.

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+		Auths: dockerConfig,
+	}
+	authfileData, err := json.Marshal(dockerConfigJSON)
+	require.Nil(t, err)
+
+	require.Nil(t, writeFileToNode(t, cs, node, authfileData, authfilePath))
+
+	return func() {
+		// Clean up the pull secret from the node if the node has not rebooted and we do not want to delete it.
+		t.Logf("cleaning up pull secret from node")
```
kikisdeliveryservice, line 161: logf->log or alternatively add "node: %s, node.name" (or something)

tests/layering/node_cmd_test.go
cheesesashimi: The `Started` field is a pointer and one has to deference it in order to check if its true or false. I probably could've done something like:

```golang
if status.Started != nil && *status.Started != true {
  return false
}
```

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+
+	t.Logf("new OS image placed in %v", cmdResult.Duration)
+
+	return nil
+}
+
+// Rolls back to the original OS image that was on the node.
+func rollbackToOriginalImage(t *testing.T, cs *framework.ClientSet, node *corev1.Node) error {
+	// Apply the rollback to the previous image
+	t.Logf("rolling back to previous OS image")
```
kikisdeliveryservice, line 224: :)

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+
+	return false
+}
+
+// RebootAndWait reboots a node, waits until it's status is ready.
+func rebootAndWait(t *testing.T, cs *framework.ClientSet, node *corev1.Node) error {
+	t.Logf("rebooting %s", node.Name)
+
+	// Get an updated node object since this one may potentially be out-of-date
+	updatedNode, err := cs.Nodes().Get(context.TODO(), node.ObjectMeta.Name, metav1.GetOptions{})
```
kikisdeliveryservice, line 312:
```suggestion
	updatedNode, err := cs.Nodes().Get(context.TODO(), node.Name, metav1.GetOptions{})
```
You shouldn't need ObectMeta to get name for node of corev1.Node type for example in our repo: https://github.com/openshift/machine-config-operator/blob/9c6c2bfd7ed498bfbc296d530d1839bd6a177b0b/pkg/daemon/node.go#L24

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+
+	prevBootID := updatedNode.Status.NodeInfo.BootID
+
+	if _, err := ExecCmdOnNode(cs, node, []string{"systemctl", "reboot"}); err != nil {
+		return fmt.Errorf("could not reboot node: %w", err)
+	}
+
+	startTime := time.Now()
+	err = wait.Poll(2*time.Second, 10*time.Minute, func() (bool, error) {
+		n, err := cs.Nodes().Get(context.TODO(), node.ObjectMeta.Name, metav1.GetOptions{})
```
kikisdeliveryservice, line 325: same as above re:objectmeta

tests/layering/layering_test.go
cheesesashimi: This value gets replaced in with either `latest` or the OCP release (e.g., `4.11`) using https://github.com/openshift/os/pull/746/files#diff-58ffaa9005e1e047a4897d0a0d8a414238a5334c730449caa62fbd8fca7cb34dR36.

This occurs here: https://github.com/openshift/os/pull/746/files#diff-58ffaa9005e1e047a4897d0a0d8a414238a5334c730449caa62fbd8fca7cb34dR75.

My reasoning was that this could run on all of the release branches as well as `master`, so it should pull the appropriate image for that branch.

tests/layering/node_cmd_test.go
kikisdeliveryservice: i was thinking the same thing on that one :)

tests/layering/node_cmd_test.go
```diff
@@ -0,0 +1,436 @@
+package e2e_test
```
cheesesashimi, line 1: I wanted to mention that I may want to eventually port this library to the MCO's testing library. With this present, we could reduce our need to shell out to `oc` to execute arbitrary commands on a node.

tests/layering/helpers_test.go
cheesesashimi: I made that mistake a lot; sorry! I'll get them all fixed and make a note to tune my linter.

tests/layering/node_cmd_test.go
kikisdeliveryservice: good idea, along those lines, can you add an owners file to this tests/layering?

tests/layering/layering_test.go
kikisdeliveryservice: for some reason link not working what line is it (just for my own info when you get a chance)

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+	runOpts := NodeCmdOpts{
+		Command: []string{"systemctl", "show", "--no-page", "rpm-ostreed"},
+		Retries: 5,
+		RetryCheckFunc: func(attempt int, cr *CmdResult, err error) bool {
+			// We ran into an error trying to get the rpm-ostreed status, retry.
+			if err != nil {
+				return false
+			}
+
+			combined := string(cr.Stdout) + string(cr.Stderr)
```
kikisdeliveryservice, line 361: do we need a space or colon or something here ie ...+ " " + string(...)

cheesesashimi: I don't think we do, but it couldn't hurt!

tests/layering/layering_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+type Deployments struct {
+	Booted                  bool   `json:"booted"`
+	ContainerImageReference string `json:"container-image-reference"`
+}
+
+type Status struct {
+	Deployments []Deployments `json:"deployments"`
+}
+
```
mkenigs, line 43: We really need to make this a library somewhere :flushed: 
https://github.com/jkyros/machine-config-operator/pull/1#discussion_r819106902

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+
+	_, err := cs.ImageStreams(mcoNamespace).Create(context.TODO(), imageStreamConfig, metav1.CreateOptions{})
+	if err != nil && !k8sErrors.IsAlreadyExists(err) {
+		// If we have an error and it is not an already exists error, something is wrong.
+		return err
+	}
+
+	if k8sErrors.IsAlreadyExists(err) {
+		// We've already got an imagestream matching this name
+		imagestream, err := cs.ImageStreams(mcoNamespace).Get(context.TODO(), imageStreamName, metav1.GetOptions{})
```
mkenigs, line 63: Not sure how much it matters but I've been asked before to do `ctx := context.TODO()` and then use `ctx` when I'm calling `context.TODO()` repeatedly. You have that here and a few other places

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+// a previously-built image, if one is available.
+func maybeBuildDerivedOSImage(t *testing.T, cs *framework.ClientSet) error {
+	imageStreamConfig := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      imageStreamName,
+			Namespace: mcoNamespace,
+		},
+	}
+
+	_, err := cs.ImageStreams(mcoNamespace).Create(context.TODO(), imageStreamConfig, metav1.CreateOptions{})
```
mkenigs, line 55: Might be my imagination but it feels like a slightly more common pattern to call `Get` first and then `Create` second if it doesn't exist

tests/layering/layering_test.go
mkenigs: tests/layering/helpers_test.go:75

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+		}
+
+		// Lets see if it has an image we can use
+		if len(imagestream.Status.Tags) != 0 {
+			t.Logf("image already built, reusing!")
+			return nil
+		}
+	}
+
+	baseImageBuildArg := fmt.Sprintf(imagePullSpec, getBaseImageTag())
```
mkenigs, line 75: How come you're grabbing this from an env var?

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+	if err != nil {
+		return fmt.Errorf("could not create build: %w", err)
+	}
+
+	t.Log("base image derivation build started")
+
+	build, err := cs.BuildV1Interface.Builds(mcoNamespace).Get(context.TODO(), buildConfig.Name, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("could not get build %s: %w", build.Name, err)
+	}
```
mkenigs, line 125: Does the `Create` call already return the build?

tests/layering/helpers_test.go
```diff
@@ -0,0 +1,611 @@
...
+	dockerConfigJSON := credentialprovider.DockerConfigJSON{
+		Auths: dockerConfig,
+	}
+	authfileData, err := json.Marshal(dockerConfigJSON)
+	require.Nil(t, err)
+
+	require.Nil(t, writeFileToNode(t, cs, node, authfileData, authfilePath))
+
+	return func() {
+		// Clean up the pull secret from the node if the node has not rebooted and we do not want to delete it.
```
mkenigs, line 160:
```suggestion
		// Allow cleaning up the pull secret from the node if the node will not be rebooted or deleted
```

tests/layering/layering_test.go
```diff
@@ -0,0 +1,148 @@
...
+		},
+		{
+			name: "Boots Into Derived Image",
+			testFunc: func(t *testing.T) {
+				nodeRebooted := false
+				deletePullSecret := putPullSecretOnNode(t, cs, targetNode)
+				defer func() {
+					// The pull secret should be cleared from the node upon reboot since
+					// it is placed into the /run directory. However, if the node does
+					// not reboot and we're not deleting it, we should clean it up.
```
mkenigs, line 103: I got confused about whether it was the pull secret or the node, maybe
```suggestion
					// not reboot or get deleted, we should clean it up.
```

tests/layering/helpers_test.go
cheesesashimi: What you suggested is good practice and something I should have been doing. Will fix.

tests/layering/helpers_test.go
cheesesashimi: I could see that being the case. If that's what's more common, I'll adopt that pattern instead.

tests/layering/helpers_test.go
cheesesashimi: Sort of. I don't think it's a complete build object. In other words, if you didn't supply some metadata, what it returns may be missing that metadata. However, I think that may be fine for what I'm doing with it.

tests/layering/helpers_test.go
cheesesashimi: Yeah, I'll go through and fix all of those. I'm not sure why I did that ¯\_(ツ)_/¯

tests/layering/helpers_test.go
cheesesashimi: I have a slightly better idea to do that, somewhat inspired by how `os/exec` does it.

tests/layering/helpers_test.go
mkenigs: I dunno if it actually is though, I bet @kikisdeliveryservice would know?

tests/layering/layering_test.go
mkenigs: We also have a duplicated `getRPMOStreeStatus`

tests/layering/helpers_test.go
kikisdeliveryservice: I'm a little confused bc I can't find this line anymore - but it depends on what you're doing (which I can't tell bc I can't find the line)

cheesesashimi:
After a bunch of refactoring, I think this is ready for a final review.

cgwalters (approved):
Oops I started a review but forgot to click submit.
Nothing blocking though!

cgwalters:
Is it correct the state of things that this just adds the test, but nothing executes it today?

Are you thinking we'd run the test in the periodic?

Nothing stops us from having an opt-in
`/test derivation`
here too?

OK but actually, don't we need to wire this test up to "launch a cluster" bits?

cgwalters:
From my PoV though I'm good to `/lgtm` and iterate further after merge.  Will leave open for a bit in case anyone else has comments.

cheesesashimi:
> Is it correct the state of things that this just adds the test, but nothing executes it today?

Yup. This just adds the test for now.

> Are you thinking we'd run the test in the periodic?

That was my plan. I think it would also be good to run on PRs as you mention below this.

> OK but actually, don't we need to wire this test up to "launch a cluster" bits?

Correct. That's being worked on in https://github.com/openshift/release/pull/26912. The "magic" bit over there is the `workflow: ipi-gcp` line which causes the OpenShift CI system to provision a cluster.

kikisdeliveryservice:
I can give this a final pass this afternoon

kikisdeliveryservice:
This looks great, Zack. 

/lgtm

PR #748, opened 2022-03-17
image: specify VMware OS and HW versions

bgilbert:
Update to hardware version 15 and switch the OS ID from RHEL 7 to 64-bit RHEL 8.

For https://github.com/coreos/coreos-assembler/pull/2762.

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -13,3 +13,8 @@ squashfs-compression: gzip
 ignition-network-kcmdline: []
 
 ostree-format: "oci"
+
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

miabbott:
I know this will address Bug 2050452, but it needs to be delivered as part of a bootimage update.

The BZ tooling will move this to MODIFIED once it merges, so maybe drop the `Bug 2050452` to short-circuit the tooling?

Afterwards, I'll give it the `/lgtm`

Issue #749, opened 2022-03-17
RHCOS 4.11 + RHEL 8.6 x86: Failed to insert 'ipmi_si': No such device


jschintag:
When runnning RHCOS 4.11 based on RHEL 8.6 on x86, the `systemd-modules-load.service` fails to load the `ipmi_si` module.
```
# systemctl status systemd-modules-load
● systemd-modules-load.service - Load Kernel Modules
   Loaded: loaded (/usr/lib/systemd/system/systemd-modules-load.service; static; vendor preset: disabled)
   Active: active (exited) since Thu 2022-03-17 12:01:09 UTC; 8min ago
     Docs: man:systemd-modules-load.service(8)
           man:modules-load.d(5)
  Process: 928 ExecStart=/usr/lib/systemd/systemd-modules-load (code=exited, status=0/SUCCESS)
 Main PID: 928 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 5555)
   Memory: 0B
   CGroup: /system.slice/systemd-modules-load.service

Mar 17 12:01:09 localhost systemd-modules-load[928]: Module 'msr' is builtin
Mar 17 12:01:09 localhost systemd[1]: Started Load Kernel Modules.

# dmesg | grep ipmi
[    8.582521] ipmi device interface
[    8.633522] ipmi_si: IPMI System Interface driver
[    8.640952] ipmi_si: Unable to find any System Interface(s)
[    8.659581] systemd-modules-load[384]: Failed to insert 'ipmi_si': No such device
```
Relate bug https://bugzilla.redhat.com/show_bug.cgi?id=2037294

HuijingHei:
Check on 4.11 (based on rhel8.5) and 4.10 (based on rhel8.4), no such issue. And no ipmi related modules found

jschintag:
Looking at the BZ and comparing it to s390x, i think we have to different problems here depending on platform:
x86:
- missing device for `ipmi_si` kernel module
- potentially fixed in the BZ
s390x:
- missing `ipmi` modules (not installed by rpm)
```
[core@cosa-devsh char]$ rpm -ql kernel-core | grep ipmi
[core@cosa-devsh char]$ rpm -ql kernel-modules | grep ipmi
```

HuijingHei:
Upgrade to fwupd-1.7.1-1.el8.x86_64, result is the same as https://bugzilla.redhat.com/show_bug.cgi?id=2037294#c8
```
# systemctl status systemd-modules-load
[...]

PR #757, opened 2022-03-24
manifest.yaml: add workaround for fwupd

HuijingHei:
TEMPORARY: Remove `ipmi-si` in /usr/lib/modules-load.d/fwupd-redfish.conf
until fix is landed to RHEL 8.6, fixed version is fwupd-1.7.4-1.el8
See:
- https://github.com/openshift/os/issues/749
- https://bugzilla.redhat.com/show_bug.cgi?id=2037294
- https://github.com/fwupd/fwupd/pull/4162

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -117,6 +117,15 @@ postprocess:
      grep -q '# RHEL-only: Disable /tmp on tmpfs' /usr/lib/systemd/system/basic.target
      echo '# RHCOS-only: we follow the Fedora/upstream default' >> /usr/lib/systemd/system/basic.target
      echo 'Wants=tmp.mount' >> /usr/lib/systemd/system/basic.target
+
+     # TEMPORARY: Remove `ipmi-si` in /usr/lib/modules-load.d/fwupd-redfish.conf until fix is landed to RHEL 8.6
+     # Fixed version is fwupd-1.7.4-1.el8
+     # See https://github.com/openshift/os/issues/749
+     #     https://bugzilla.redhat.com/show_bug.cgi?id=2037294
+     #     https://github.com/fwupd/fwupd/pull/4162
+     [ ! -f /usr/lib/modules-load.d/fwupd-redfish.conf ] && exit 0
+     grep -q ipmi-si /usr/lib/modules-load.d/fwupd-redfish.conf || exit 100
+     sed -i '/ipmi-si/d' /usr/lib/modules-load.d/fwupd-redfish.conf
   - |
      #!/usr/bin/env bash
      set -xeo pipefail


manifest.yaml
```diff
@@ -117,6 +117,18 @@ postprocess:
...
      echo '# RHCOS-only: we follow the Fedora/upstream default' >> /usr/lib/systemd/system/basic.target
      echo 'Wants=tmp.mount' >> /usr/lib/systemd/system/basic.target
+
+     # TEMPORARY: Remove `ipmi-si` in /usr/lib/modules-load.d/fwupd-redfish.conf until fix is landed to RHEL 8.6
+     # Fixed version is fwupd-1.7.4-1.el8
+     # See https://github.com/openshift/os/issues/749
+     #     https://bugzilla.redhat.com/show_bug.cgi?id=2037294
+     #     https://github.com/fwupd/fwupd/pull/4162
+     . /etc/os-release
+     if [ "$(uname -m)" == "x86_64" ]; then
```
travier, line 127: Let's skip this and do it for all arches.

manifest.yaml
```diff
@@ -117,6 +117,18 @@ postprocess:
...
+     [ "${ID}" == "rhel" ] && [ ${VERSION_ID//.} -ne 86 ] && exit 0
```
travier, line 128: I don't think we need this one if we do the one below

travier:
No need for ALL CAPS too 🙂

manifest.yaml
HuijingHei: You are correct, I will remove this

manifest.yaml
HuijingHei: For this, I just do not want to make sure this does not affect rhel8.5
Check on rhel8.5, no found /usr/lib/modules-load.d/fwupd-redfish.conf

HuijingHei: Maybe I should check the file before grep

travier (approved):

HuijingHei:
Current 8.6 beta includes fwupd-1.7.1-1.el8.x86_64.rpm ( without fixed patch ), need to include this patch.

manifest.yaml
```diff
@@ -117,6 +117,15 @@ postprocess:
...
      echo '# RHCOS-only: we follow the Fedora/upstream default' >> /usr/lib/systemd/system/basic.target
      echo 'Wants=tmp.mount' >> /usr/lib/systemd/system/basic.target
+
+     # TEMPORARY: Remove `ipmi-si` in /usr/lib/modules-load.d/fwupd-redfish.conf until fix is landed to RHEL 8.6
+     # Fixed version is fwupd-1.7.4-1.el8
+     # See https://github.com/openshift/os/issues/749
+     #     https://bugzilla.redhat.com/show_bug.cgi?id=2037294
+     #     https://github.com/fwupd/fwupd/pull/4162
+     [ ! -f /usr/lib/modules-load.d/fwupd-redfish.conf ] && exit 0
+     grep -q ipmi-si /usr/lib/modules-load.d/fwupd-redfish.conf || exit 100
```
cgwalters, line 127: I'm guessing someone else asked for this but IMO, the "hard fail the build when the workaround is no longer needed" causes more pain than it solves.  I think it's nicer to ourselves to do "tech debt reminders", e.g. scheduling a reminder for 3 months later to check and see if a workaround is still needed.

(But, this is fine as is)

HuijingHei: Good suggestion. I should add more text to clarify when to remove this temporary workaround in the future, for this maybe "Should remove this workaround if the version of fwupd >= 1.7.4-1.el8"

miabbott: I added a reminder on our internal Slack instance to remove this when 8.6 hits GA; presumably that is when the fixed `fwupd` will land

PR #763, opened 2022-03-31
MCO-280: extensions: Add container to serve repository with extensions

jmarrero:
depends on: https://github.com/coreos/rpm-ostree/pull/3562

This container will create a small service that will host the extensions rpms that we will be shipping. The idea is that this new container will be shipped alongside the machine-os and would provide a way for users layering extensions a way to do so without depending of connections outside of the cluster.

https://issues.redhat.com/browse/MCO-288

### DIFF
--- a/ci/get-ocp-repo.sh
+++ b/ci/get-ocp-repo.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# TODO: share this with ci/prow-entrypoint.sh
+ocpver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["mutate-os-release"]')
+ocpver_mut=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|')
+rhelver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["automatic-version-prefix"]' | cut -f2 -d.)
+
+curl -L "http://base-${ocpver_mut}-rhel${rhelver}.ocp.svc.cluster.local" -o "ocp.repo"
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -1,0 +1,48 @@
+ARG RHCOS_VERSION=latest
+#Build a newer rpm-ostree
+FROM quay.io/centos/centos:stream8 as rpm-ostree
+RUN sed -i -e 's,enabled=0,enabled=1,' /etc/yum.repos.d/CentOS-Stream-PowerTools.repo
+RUN yum -y group install "Development Tools"
+
+RUN git clone https://github.com/coreos/rpm-ostree.git
+WORKDIR rpm-ostree
+RUN git checkout rhel8
+RUN ./ci/installdeps.sh
+RUN PATH=$PATH:/rpm-ostree/target/cxxbridge/bin/
+RUN git submodule update --init
+RUN env CFLAGS='-ggdb -Og' CXXFLAGS='-ggdb -Og' ./autogen.sh --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc
+RUN make
+
+## Downloads the extensions given the extensions.yaml
+FROM registry.ci.openshift.org/rhcos-devel/rhel-coreos:${RHCOS_VERSION} as os
+
+# Install new rpm-ostree
+COPY --from=rpm-ostree /rpm-ostree/target/debug/rpm-ostree /usr/bin/rpm-ostree
+
+# Expects os to be cloned and this build run from the top level dir like:
+# podman build -f extensions/Dockerfile .
+# also expects submodules to be initialized
+RUN mkdir /os
+WORKDIR /os
+ADD . .
+RUN if [ ! -f ocp.repo  ]; then ci/get-ocp-repo.sh ; fi
+
+RUN rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}.yaml
+
+## Creates the repo metadata for the extensions & builds the go binary
+FROM quay.io/centos/centos:stream8 as builder
+COPY --from=os /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
+RUN dnf install -y createrepo_c golang
+ADD extensions/repo-server/main.go .
+RUN mkdir /build
+RUN go build -o /build/webserver main.go
+
+RUN createrepo_c /usr/share/rpm-ostree/extensions/
+
+## Final container that has the extensions and webserver
+FROM registry.access.redhat.com/ubi8/ubi:latest
+COPY --from=builder /build/webserver /usr/bin/webserver
+COPY --from=builder /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
+
+CMD ["./usr/bin/webserver"]
+EXPOSE 9091/tcp
--- a/extensions/test/Dockerfile
+++ b/extensions/test/Dockerfile
@@ -1,0 +1,7 @@
+FROM registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
+#We want to test it's only using the extensions repo
+RUN rm -rf /etc/yum.repos.d/*
+ADD ext.repo /etc/yum.repos.d/extension.repo
+#Install usbguard as provided by the repo
+RUN rpm-ostree install usbguard
+RUN usbguard
--- a/extensions/test/ext.repo
+++ b/extensions/test/ext.repo
@@ -1,0 +1,5 @@
+[ext]
+name=extensions
+baseurl=http://localhost:9091/
+enabled=1
+gpgcheck=1
--- a/extensions/repo-server/main.go
+++ b/extensions/repo-server/main.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"net/http"
+)
+
+var (
+	// flagPort is the open port the application listens on
+	flagPort = flag.String("port", "9091", "Port to listen on")
+)
+
+func main() {
+	extensions := http.FileServer(http.Dir("/usr/share/rpm-ostree/extensions/"))
+	flag.Parse()
+	mux := http.NewServeMux()
+	mux.Handle("/", extensions)
+
+	log.Printf("listening on port %s", *flagPort)
+	log.Fatal(http.ListenAndServe(":"+*flagPort, mux))
+}


jmarrero:
> Let's put all these files in an `extensions/` subdirectory?
That makes sense, I'll push that change today.

extensions/Dockerfile
```diff
@@ -0,0 +1,25 @@
+## Builds the webserver for the repo service.
+FROM golang as gobuild
+ADD repo-server/main.go .
+RUN mkdir /build
+RUN go build -o /build/webserver main.go
+
+## Downloads the extensions given the extensions.yaml
+ARG RHCOS_VERSION=latest
```
jmarrero, line 8: something interesting is that buildah bud can deal with the ARG before the FROM line but podman build can't. It looks like it's a design/scope decision when using podman build: https://github.com/containers/podman/issues/9392

cgwalters (changes requested):

extensions/Dockerfile
```diff
@@ -0,0 +1,25 @@
+## Builds the webserver for the repo service.
+FROM golang as gobuild
```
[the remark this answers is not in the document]
jmarrero: I am not sure if I am missing some go magic but when using that image or a newer one to build I get:
```
Trying to pull registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.18-openshift-4.11...
Getting image source signatures
Copying blob 47aa3ed2034c skipped: already exists  
Copying blob eac1b95df832 skipped: already exists  
Copying blob 1cfc2c7a4c70 done  
Copying blob 455a09169fee done  
Copying config 3122dc0047 done  
Writing manifest to image destination
Storing signatures
[1/4] STEP 2/5: ADD repo-server/main.go .
[1/4] STEP 3/5: RUN mkdir /build
[1/4] STEP 4/5: RUN go build -o /build/webserver main.go
panic: internal error: rs.graph is unexpectedly nil with -mod=vendor

goroutine 1 [running]:
cmd/go/internal/modload.updatePrunedRoots({0xadb6d8, 0xc0000280a0}, 0x40fc1b?, 0xc000020280, {0xc000300400, 0x79, 0x0?}, {0x0, 0x0, 0x0}, ...)
        /usr/lib/golang/src/cmd/go/internal/modload/buildlist.go:955 +0xb6f
cmd/go/internal/modload.updateRoots({0xadb6d8?, 0xc0000280a0?}, 0x40ecc5?, 0x7f3d53146a68?, {0xc000300400?, 0x400?, 0xc0000cd1b8?}, {0x0, 0x0, 0x0}, ...)
        /usr/lib/golang/src/cmd/go/internal/modload/buildlist.go:653 +0xb7
cmd/go/internal/modload.(*loader).updateRequirements(0xc00012b5f0, {0xadb6d8, 0xc0000280a0})
        /usr/lib/golang/src/cmd/go/internal/modload/load.go:1361 +0x2a5
cmd/go/internal/modload.loadFromRoots({0xadb6d8, 0xc0000280a0}, {{{0x0, 0x0}, 0xc00012c450, 0x0, {0x0, 0x0}, 0x0, 0x1, ...}, ...})
        /usr/lib/golang/src/cmd/go/internal/modload/load.go:1096 +0x70f
cmd/go/internal/modload.ImportFromFiles({0xadb6d8, 0xc0000280a0}, {0xc0000200e0, 0x1, 0x1})
        /usr/lib/golang/src/cmd/go/internal/modload/load.go:704 +0x234
cmd/go/internal/load.GoFilesPackage({0xadb6d8, 0xc0000280a0}, {0x38?, 0x3?, 0xa4?}, {0xc0000200e0?, 0x1, 0x1})
        /usr/lib/golang/src/cmd/go/internal/load/pkg.go:2925 +0x22f
cmd/go/internal/load.PackagesAndErrors({0xadb6d8?, 0xc0000280a0?}, {0xb?, 0x0?, 0x0?}, {0xc0000200e0, 0x1, 0x1})
        /usr/lib/golang/src/cmd/go/internal/load/pkg.go:2673 +0xbf3
cmd/go/internal/work.runBuild({0xadb6d8, 0xc0000280a0}, 0xc0000243f0?, {0xc0000200e0, 0x1, 0x1})
        /usr/lib/golang/src/cmd/go/internal/work/build.go:382 +0xa7
main.invoke(0xdb62e0, {0xc0000200b0, 0x4, 0x4})
        /usr/lib/golang/src/cmd/go/main.go:218 +0x2ee
main.main()
        /usr/lib/golang/src/cmd/go/main.go:175 +0x78e
error building at STEP "RUN go build -o /build/webserver main.go": error while running runtime: exit status 2
ERRO[0045] exit status 2   ```

I changed back to a `golang` base image using go 1.17 and 1.18 instead of 1.18 and they still work. I also changed my Dockerfile to use Centos8 stream to build too not just create the repo and that works as well. I updated the PR with that last version.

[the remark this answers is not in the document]
Comment (jmarrero) [reply extensions/Dockerfile]: Humm based on: https://maelvls.dev/go111module-everywhere/ and https://github.com/golang/go/issues/16562 I think [this](https://github.com/openshift/machine-config-operator/blob/5516bd8911ce3e53cfc20924609c129d2eabc604/Makefile#L17) is manipulating the default value on the image on the MCO build.  I will give that a try. But... @cgwalters does it make sense to build and create the repo on the same centos container either way?

**Comment by cgwalters** [comment]
Cross-linking to https://issues.redhat.com/browse/MCO-288

I've been thinking about this one more, particularly in light of the discovery in https://github.com/coreos/coreos-assembler/issues/2685#issuecomment-1149050545
Alongside that, a general problem we knew with Dockerfile approach here is that we're giving up on good "change detection".

I think combining those two facts pulls this back towards being something like `cosa buildextend-extensions-container`.

**Comment by jmarrero** [comment]
> Cross-linking to https://issues.redhat.com/browse/MCO-288
> 
> I've been thinking about this one more, particularly in light of the discovery in [coreos/coreos-assembler#2685 (comment)](https://github.com/coreos/coreos-assembler/issues/2685#issuecomment-1149050545) Alongside that, a general problem we knew with Dockerfile approach here is that we're giving up on good "change detection".
> 
> I think combining those two facts pulls this back towards being something like `cosa buildextend-extensions-container`.

Closing this PR in favor of the above.

**Comment by cgwalters** [comment]
> Closing this PR in favor of the above.

Ideally though, we can reuse some or most of the code/approach here.  We still need the webserver, etc.  This was definitely not a lost effort!

I might have said we should keep the PR open as a reference and then rework it, but either way the code isn't lost.

Or alternatively...we could keep this all as is, but have a flow that updates meta.json *after* the "main" RHCOS build job.  (Though, this gets into races unless we have a coordinator...needs some design)

**Comment by cgwalters** [comment]
Joseph and I had a live chat about this.  We're thinking we can:

- Keep the code here 
- Add `cosa buildextend-extensions-container` that knows to build the `Dockerfile` in `extensions/`.  This flow would either need to use the [nested container stuff](https://github.com/coreos/coreos-assembler/blob/1d83765e132b65915f674a296b6dda6df325bcf2/src/oscontainer.py#L116) or (I think I like this better) do the build via KVM in the qemu image which we already have.  Though this path will be implicitly testing podman.
- This build must substitute the base RHCOS in Dockerfile to pull the base ostree container, and not from a registry.  
- This new extensions container needs to be appear in `meta.json` in the same way as the main one, see https://github.com/coreos/coreos-assembler/pull/2907.  Probably change `push-container` to add `cosa push-container --extensions` or so?

File: extensions/Dockerfile (hunk 2 above)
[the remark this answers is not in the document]
Comment (cgwalters) [reply extensions/Dockerfile]: Let's figure this one out before merging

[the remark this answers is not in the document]
Comment (jmarrero) [reply extensions/Dockerfile]: I just used stream8 to also build the go binary which gives one one less stage in my latest push.

**Comment by jmarrero** [comment]
This depends on https://github.com/coreos/rpm-ostree/issues/3819

File: extensions/Dockerfile [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+    ocpver_mut=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|') && \
+    rhelver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["automatic-version-prefix"]' | cut -f2 -d.) && \
+    curl -L "http://base-${ocpver_mut}-rhel${rhelver}.ocp.svc.cluster.local" -o "ocp.repo"
+
+RUN rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}.yaml
+
+## Creates the repo metadata for the extensions & builds the go binary
+FROM quay.io/centos/centos:stream8 as builder
+COPY --from=os /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
+RUN dnf install -y createrepo_c golang
```
jmarrero, line 38: @cgwalters no createrepo_c or createrepo on ubi :(

cheesesashimi:
> This flow would either need to use the [nested container stuff](https://github.com/coreos/coreos-assembler/blob/1d83765e132b65915f674a296b6dda6df325bcf2/src/oscontainer.py#L116) or (I think I like this better) do the build via KVM in the qemu image which we already have

For simplicity, I'd rather point Prow at `extensions/Dockerfile` and let it build the container image as opposed to needing a registry roundtrip / extraction step like we currently have for the RHCOS image.

ci/get-ocp-repo.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+echo test!
```
cgwalters, line 2: We can drop this bit right?

extensions/test/ext.repo
```diff
@@ -0,0 +1,5 @@
+[ext]
+name=extensions
+baseurl=http://localhost:9091/
+enabled=1
+gpgcheck=0
```
cgwalters, line 5: Not a blocker, but in theory we can enable this right?

extensions/test/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
+#We want to test it's only using the extensions repo
+RUN rm -rf /etc/yum.repos.d/*
+ADD ext.repo /etc/yum.repos.d/extention.repo
```
cgwalters, line 4: s/extention/extension/

extensions/test/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+#Install usbguard as provided by the repo
+RUN rpm-ostree install usbguard
+RUN usbguard
```
cgwalters, line 7: Maybe better as `usbguard --help`?  Oh weird...just invoking `usbguard` prints the usage but also exits successfully, but `usbguard --help` exits with code 1.  And annoyingly it doesn't have a no-op success command like `usbguard --version`.

Oh well, OK as is I guess!

extensions/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ADD extensions/repo-server/main.go .
+RUN mkdir /build
+RUN go build -o /build/webserver main.go
+
```
cgwalters, line 39: Any reason to keep this blank line?  It seems like a leftover from doing two separate container builds maybe?

ci/get-ocp-repo.sh
```diff
@@ -0,0 +1,7 @@
...
+ocpver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["mutate-os-release"]')
+ocpver_mut=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|')
+rhelver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["automatic-version-prefix"]' | cut -f2 -d.)
+
+curl -L "http://base-${ocpver_mut}-rhel${rhelver}.ocp.svc.cluster.local" -o "ocp.repo"
```
cgwalters, line 7: This is totally OK for now but maybe a quick:
`# TODO: share this with ci/prow-entrypoint.sh` ?

cgwalters (commented):
Overall looking good to me!  Just a few nits

ci/get-ocp-repo.sh
jmarrero: TODO added. I can update the scripts when doing the COSA work as will probably need to change the Dockerfile then too.

extensions/Dockerfile
jmarrero: removed.

extensions/test/Dockerfile
jmarrero: fixed

ci/get-ocp-repo.sh
jmarrero: yes sir.

extensions/test/ext.repo
jmarrero: yeah, enabled in latest push

PR #766, opened 2022-04-04
tests/kola/version rhel-major-version: ensure packages from rhel major version are included

RyanJenkins99:
I added a test to ensure packages from the RHEL major version are included, this allows us to make sure we're not mixing up versions while also staying up to date

### DIFF
--- a/tests/kola/version/rhel-major-version
+++ b/tests/kola/version/rhel-major-version
@@ -1,0 +1,18 @@
+#!/bin/bash
+# kola: { "exclusive": false }
+set -xeuo pipefail
+
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# checks to ensure that the only packages from the RHEL major version are included
+
+source /etc/os-release 
+var=$(echo "${RHEL_VERSION}" | cut -d. -f1)
+for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+'); do
+    if [[ "$var" -ne $((x)) ]]; then 
+        fatal "Error RHEL packages do not match current version"
+    fi
+done


tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,15 @@
...
+    echo "$@" >&2
+    exit 1
+}
+
+#checks to ensure that the only packages from the RHEL major version are included
+source /etc/os-release
+
+if [ rpm -qa --queryformat='%{VERSION}\n' systemd ]; then
+    fatal "Error RHEL packages do not match current version"
+fi
```
miabbott, line 15: I don't think this is doing what you think it is doing.  It's mixing up a two query operations: `-qa` is used to fetch *all* the RPMs on a system and providing `systemd` you are asking for just `systemd`

This `if` statement will always succeed (i.e. return `0`) if `systemd` is installed in the OS...which unless RHCOS/RHEL changes drastically in the near future, will always be the case.

You need to set a variable to the RHEL major version, probably derived from the `RHEL_VERSION` in `/etc/os-release`.  (In this case the RHEL major version should be `8`)

Then you will want to query all the RPMs on the system to make sure they are for that RHEL major version.

Start by looking at `rpm -qa --queryformat='%{RELEASE}\n'` as an example of how to get the release value of all the installed packages.  Then sort out a way to compare the output to ensure that all the packages are targeting the correct RHEL major version.

ashcrow:
@RyanJenkins99 what syntax error are you seeing?

tests/kola/version/rhel-major-version
```diff
@@ -12,8 +12,11 @@ fatal() {
 source /etc/os-release
 
 
+RHEL_VERSION="8"
```
ravanelli, line 15: You used source and then replaced the value for 8. Remove this line or it will always be 8

ravanelli (changes requested):

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,22 @@
...
+# checks to ensure that the only packages from the RHEL major version are included
+
+source /etc/os-release
+
+
+RHEL_VERSION="8"   
+
+for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+')
+    do
+        if "${RHEL_VERSION} != "${x}""; then
```
miabbott, line 19: This branch is failing when run locally.  

I did `cosa run` to get an RHCOS 4.11 system up, downloaded this script, and ran it manually:

```
[core@cosa-devsh ~]$ rpm-ostree status
State: idle
Deployments:
* b97794784366ba8e34978aada1a32007dcd5cc4256b507696f2857c3f6e88a8c
                   Version: 411.86.202204131300-0 (2022-04-13T13:03:08Z)
[core@cosa-devsh ~]$ curl -LO https://raw.githubusercontent.com/openshift/os/53ac06f6402cf2cf494a04078f7988b087c51997/tests/kola/version/rhel-major-version
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   450  100   450    0     0   7377      0 --:--:-- --:--:-- --:--:--  7377
[core@cosa-devsh ~]$ chmod +x rhel-major-version 
[core@cosa-devsh ~]$ ./rhel-major-version                                                                                                                                                                                                                                                                                     
+ source /etc/os-release                                                  
++ NAME='Red Hat Enterprise Linux CoreOS'                                 
++ ID=rhcos                                                               
++ ID_LIKE='rhel fedora'                                                  
++ VERSION=411.86.202204131300-0                                          
++ VERSION_ID=4.11                                                        
++ PLATFORM_ID=platform:el8                                               
++ PRETTY_NAME='Red Hat Enterprise Linux CoreOS 411.86.202204131300-0 (Ootpa)'
++ ANSI_COLOR='0;31'                                                      
++ CPE_NAME=cpe:/o:redhat:enterprise_linux:8::coreos
++ HOME_URL=https://www.redhat.com/
++ DOCUMENTATION_URL=https://docs.openshift.com/container-platform/4.11/
++ BUG_REPORT_URL=https://bugzilla.redhat.com/
++ REDHAT_BUGZILLA_PRODUCT='OpenShift Container Platform'
++ REDHAT_BUGZILLA_PRODUCT_VERSION=4.11
++ REDHAT_SUPPORT_PRODUCT='OpenShift Container Platform'
++ REDHAT_SUPPORT_PRODUCT_VERSION=4.11
++ OPENSHIFT_VERSION=4.11
++ RHEL_VERSION=8.6
++ OSTREE_VERSION=411.86.202204131300-0
+ RHEL_VERSION=8
++ rpm -qa '--queryformat=%{RELEASE}\n'
++ grep -oP 'el\s*\K\d+'
+ for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+')
+ '8 != 8'
./rhel-major-version: line 19: 8 != 8: command not found
+ for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+')
+ '8 != 8'
...
[core@cosa-devsh ~]$ echo $?
0
```

Since the `if` statement fails, it never drops into the code where failure would be reported and the whole script exits successfully.

This is giving you a false positive result.

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,22 @@
...
+            fail "the expected version does not equal what we found."
```
miabbott, line 20: `fail` is not a valid command here; I think you meant `fatal`

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,22 @@
...
+fatal() {
+    echo "$@" >&2
+    exit 1
+}
+
+# checks to ensure that the only packages from the RHEL major version are included
+
+source /etc/os-release
+
+
```
miabbott, line 14: too many blank lines; one is sufficient between lines of code

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,22 @@
...
+RHEL_VERSION="8"   
+
+for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+')
+    do
+        if "${RHEL_VERSION} != "${x}""; then
+            fail "the expected version does not equal what we found."
+        fi
+    done
```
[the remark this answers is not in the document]
miabbott: eh, maybe not., based on how the `for` loop is written

i think we usually have bash `for` loops that look like this:

```bash
for x in <some range>; do
   <more code>
done
```

travier:
Your commit message first line should not be longer that 50 characters.
Then a blank line and then a short text that explain your change (lines no longer than 72 characters).
You should also update this PR name and description once you have fixed your commit message.

travier:
Tests should also follow the following format: https://github.com/coreos/fedora-coreos-config#tests-layout

travier:
You should install ShellCheck and use it to check your script: https://github.com/koalaman/shellcheck
You can also use your editor to automatically indent your code properly.

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,20 @@
...
+    echo "$@" >&2
+    exit 1
+}
+
+# checks to ensure that the only packages from the RHEL major version are included
+
+source /etc/os-release 
+var=$(echo "${RHEL_VERSION}" | cut -d. -f1)
+    for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+');
+        do
```
travier, line 15: You can place this `do` in the same line as the `for` after the `;`

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,20 @@
...
+            if [ "$var" -ne $((x)) ]
```
travier, line 16: Double brackets `[[` are preferred. You can also place the `then` on the same line after a `;` here.

tests/kola/version/rhel-major-version
```diff
@@ -0,0 +1,18 @@
...
+
+# checks to ensure that the only packages from the RHEL major version are included
+
+source /etc/os-release 
+var=$(echo "${RHEL_VERSION}" | cut -d. -f1)
+    for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+'); do
+            if [[ "$var" -ne $((x)) ]]; then 
+                        fatal "Error RHEL packages do not match current version"
+            fi
+done
```
miabbott, line 18: Let's clean up the indentation like so:

```suggestion
var=$(echo "${RHEL_VERSION}" | cut -d. -f1)
for x in $(rpm -qa --queryformat='%{RELEASE}\n' | grep -oP 'el\s*\K\d+'); do
    if [[ "$var" -ne $((x)) ]]; then 
        fatal "Error RHEL packages do not match current version"
    fi
done
```

miabbott:
Looks like the CI job is hitting the test failure reported here https://github.com/coreos/fedora-coreos-tracker/issues/1184

miabbott (approved):
Worked for me locally; LGTM!

PR #769, opened 2022-04-05
adds the derived image test scripts for prow

cheesesashimi:
Instead of having complex Bash scripts embedded in YAML structures, it would be better if the scripts that were written and refined in https://github.com/openshift/release/pull/26912 lived in this repo. This allows additional benefits such as static analysis, reuse, etc.

### DIFF
--- a/ci/derived-image-test.sh
+++ b/ci/derived-image-test.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+# This script performs the following tasks:
+#
+# 1. Logs into the rhcos-devel service account to get the image push credential
+# so we can push to the registry.ci.openshift.org/rhcos-devel namespace.
+#
+# 2. Kicks off the build-test-qemu.sh script to build and test an RHCOS image
+# with Kola tests.
+#
+# 3. Pushes the resulting image to the image registry using the Prow build ID
+# as the tag since we need to have the image in this registry so our derived
+# image build test can use it.
+#
+# 3. Kicks off the derived OS image testing binary which pulls the newly built
+# image, builds a derived image from it, applies it to an underlying OpenShift
+# cluster node, verifies that it was successfully applied, then rolls the node
+# back.
+
+set -euo pipefail
+
+# Note: The oc binary will be injected by the Prow CI process. It
+# is not present in ci/Dockerfile.
+oc login https://api.ci.l2s4.p1.openshiftapps.com:6443 --token="$(cat /service-account-token/image-pusher-service-account-token)"
+oc registry login --registry=registry.ci.openshift.org --to="$SHARED_DIR/dockercfg.json"
+
+export COSA_DIR="/tmp/cosa"
+mkdir -p "$COSA_DIR"
+
+# Run the cosa build / test
+/src/ci/build-test-qemu.sh
+
+export REGISTRY_AUTH_FILE="$SHARED_DIR/dockercfg.json"
+
+# Ensure we're in the designated cosa directory so the push-container commands work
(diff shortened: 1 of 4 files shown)


ci/derived-image-test.sh
```diff
@@ -0,0 +1,49 @@
...
+
+# Run the cosa build / test
+/src/ci/build-test-qemu.sh
+
+export REGISTRY_AUTH_FILE="$SHARED_DIR/dockercfg.json"
+
+# Ensure we're in the designated cosa directory so the push-container commands work
+cd "$COSA_DIR"
+
+# Tags with the cosa build ID / arch - unique to this specific build
```
cgwalters, line 38: Will anything prune these tags by default?  From a quick look at openshift/release I see [image pruners](https://docs.openshift.com/container-platform/4.10/applications/pruning-objects.html) set up on the CI clusters but not the central CI registry?

ci/derived-image-test.sh
```diff
@@ -0,0 +1,49 @@
...
+cosa push-container registry.ci.openshift.org/rhcos-devel/rhel-coreos
+
+# Tag with the Prow Build ID because we don't want to overwrite our well-known
+# tags yet, but our test cluster needs the image to be pushed someplace so we
+# can ingest it. We use the BUILD_ID value because its unique to each job so
```
cgwalters, line 43: (Doesn't need to be in this PR but I think we should add a TODO here about how we actually do want to push to the "CI namespace" that prow creates that is naturally lifecycle bound to the PR, but we just can't do that right now because Prow wants to own the build process)

ci/derived-image-test.sh
```diff
@@ -0,0 +1,49 @@
...
+# they won't stomp on each other if running concurrently.
+cosa push-container "registry.ci.openshift.org/rhcos-devel/rhel-coreos:$BUILD_ID"
+
+# Perform the derived OS image build tests
+export BASE_IMAGE_TAG="$BUILD_ID"
```
cgwalters, line 48: Hmm, I think we should inject the full pull spec in the future - this would give us more flexibility.  (We'll hopefully be pushing the official images outside of `rhcos-devel` in the near future)

cgwalters (approved):
Overall LGTM, some things that can be fixed now or in a followup!

ci/derived-image-test.sh
cheesesashimi: I was thinking about that and only pushing to the external registry on success since that would eliminate the need to prune the build ID tags. However, there are two things I need to figure out:

1. How to get those creds within a test since the `KUBECONFIG` env var points to ones ephemeral test cluster, not the ephemeral CI registry?
2. Is the ephemeral test cluster is created with the perms to pull from the CI namespace registry? I have to assume that it is, but I could be wrong.

ci/derived-image-test.sh
cheesesashimi: This is consumed by the derived OS test, but I agree. I'll get a PR open to fix that.

ci/derived-image-test.sh
cheesesashimi: I'm not sure to be honest. It would be nice to push to the ephemeral CI registry first so one doesn't have to worry about pruning these tags.

ci/derived-image-test.sh
cheesesashimi: I updated the Golang test to search for `BASE_IMAGE_PULLSPEC` and default to `registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest` if that env var isn't set.

Issue #767, opened 2022-04-04
RHCOS4.11/s390x: cri-o causes kernel panic during exec


jschintag:
Since the rebase from RHEL 8.5 to RHEL 8.6, the `crio.base/networks-reliably` test is failing because of a kernel panic
```
=== RUN   crio.base
=== RUN   crio.base/crio-info
=== RUN   crio.base/pod-continues-during-service-restart
=== RUN   crio.base/networks-reliably
--- FAIL: crio.base (122.27s)
    --- PASS: crio.base/crio-info (0.15s)
    --- PASS: crio.base/pod-continues-during-service-restart (60.32s)
    --- FAIL: crio.base/networks-reliably (22.69s)
            cluster.go:184: "sudo crictl exec dd20dead325dcfea1332a5416772561fb8d00882721d6852ab1aa60a73049704 ping -i 0.2 10.85.0.1 -w 1 >/dev/null && echo PASS || echo FAIL" failed: output , status wait: remote command exited without exit status or exit signal
        harness.go:1355: Found kernel panic (Fatal exception: panic_on_oops) on machine cd817c09-5e65-477e-bbf7-74925313a4fc console
```
Nightly: rhcos-411.86.202204041235-0-qemu.s390x.qcow2
cri-o: cri-o-1.24.0-10.rhaos4.11.git7735eb6.el8.s390x (upgraded from cri-o-1.24.0-5.rhaos4.11.gitd020fdb.el8.s390x)
Kernel Panic:
```

[  119.992630] usercopy: Kernel memory exposure attempt detected from SLUB object 'mm_struct' (offset 1152, size 480)!

[  119.992663] ------------[ cut here ]------------

[  119.992664] kernel BUG at mm/usercopy.c:103!

[  119.992786] monitor event: 0040 ilc:2 [#1] SMP 

[...]

PR #770, opened 2022-04-05
manifest.yaml: pin sssd to a rhel-8.5 version

saqibali-2k:
`sssd` is crashing and restarting in a loop. Lets pin it to an older
version for now.

PR #771, opened 2022-04-06
s390x: Use RHEL 8.5 Kernel until 8.6 has a new kernel

jschintag:
Currently the RHEL 8.6 kernel causes a kernel panic when using exec in containers.
See: https://bugzilla.redhat.com/show_bug.cgi?id=2047242
Switch on s390x to an older kernel until the fix lands in RHEL 8.6

Fixes: https://github.com/openshift/os/issues/767

### DIFF
--- a/kernel-rhel-85.yaml
+++ b/kernel-rhel-85.yaml
@@ -1,0 +1,6 @@
+# Temporarily use RHEL 8.5 Kernel for s390x, until RHEL 8.6 gets a new kernel (Build 370+)
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=2047242
+repo-packages:
+  - repo: rhel-85-baseos
+    packages:
+      - kernel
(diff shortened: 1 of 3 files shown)


cgwalters:
But wouldn't the kernel panics trigger when the OS build is used in a cluster?  If so this would just shift *where* failures are happening.  I think we need to either revert the kernel (across the board or just for s390x) or cherry pick a newer one (if available, again where sub-options include across the board or just for s390x).

miabbott:
> But wouldn't the kernel panics trigger when the OS build is used in a cluster? If so this would just shift _where_ failures are happening. I think we need to either revert the kernel (across the board or just for s390x) or cherry pick a newer one (if available, again where sub-options include across the board or just for s390x).

Reverting to the 8.5 kernel would be in line with what we did for the `sssd` issue (#770), though I think doing it for only `s390x` seems to be the best option.  The trouble is that we won't have access to a newer 8.6 kernel until 8.6 GA, so we'll have this strange "two kernel" scenario for a while.

jschintag:
I changed the PR to downgrade the kernel for s390x

HuijingHei (approved):
LGTM

PR #776, opened 2022-04-13
separate build from test steps

cheesesashimi:
This provides a separate build script for RHCOS images which can be used within an image build. They should be separate since it does not make sense to run the test scripts within an image build. Rather, we should use the test steps for that instead.

### DIFF
--- a/ci/prow-build.sh
+++ b/ci/prow-build.sh
@@ -1,0 +1,50 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# Prow jobs don't support adding emptydir today
+export COSA_SKIP_OVERLAY=1
+# Create a temporary cosa workdir if COSA_DIR is not set.
+cosa_dir="${COSA_DIR:-$(mktemp -d)}"
+echo "Using $cosa_dir for build"
+cd "$cosa_dir"
+cosa init /src
+
+# This script is called via build.sh which is the main Prow
+# entrypoint for PRs to this repo, as well as for PRs on other repos,
+# mainly coreos-assembler.  It assumes that `cosa init` has been run.
+
+REDIRECTOR_URL="https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/"
+
+# record information about cosa + rpm-ostree
+if test -d /cosa; then
+    jq . < /cosa/coreos-assembler-git.json
+fi
+rpm-ostree --version
+
+# We generate .repo files which write to the source, but
+# we captured the source as part of the Docker build.
+# In OpenShift default SCC we'll run as non-root, so we need
+# to make a new copy of the source.  TODO fix cosa to be happy
+# if src/config already exists instead of wanting to reference
+# it or clone it.  Or we could write our .repo files to a separate
+# place.
+if test '!' -w src/config; then
+    git clone --recurse src/config src/config.writable
+    rm src/config -rf
+    mv src/config.writable src/config
+fi
+
+# Grab the raw value of `mutate-os-release` and use sed to convert the value
+# to X-Y format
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh I see you're thinking the release job would be changed to invoke them separately, but we can't do that in this PR because we'd have to ratchet it in?

Hmm the need to "ratchet" CI changes when the CI configuration lives in a separate repo is annoying :frowning:

cheesesashimi:
Not quite ratcheting. What I'm thinking about for the tests is possibly being able to shard them across multiple test steps that run concurrently and have separate `/test` targets, e.g.:

```yaml
- as: validate
  commands: ./ci/validate.sh
  container:
    from: src
  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
- as: test-qemu
  commands: |
    #!/bin/bash
    set -xeuo
    cd "${COSA_DIR}"
    cosa kola run --qemu-nvme=true basic
    cosa kola run --qemu-firmware=uefi basic
    kola run-upgrade -b rhcos -v --find-parent-image --qemu-image-dir tmp/ --output-dir tmp/kola-upgrade
    cosa kola run --parallel 2
  container:
    from: cosa-build
  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
- as: test-qemu-metal
  commands: |
    #!/bin/bash
    set -xeuo
    cd "${COSA_DIR}"
    # Build metal + installer now so we can test them
    cosa buildextend-metal && cosa buildextend-metal4k && cosa buildextend-live
    # compress the metal and metal4k images now so we're testing
    # installs with the image format we ship
    cosa compress --artifact=metal --artifact=metal4k
    # Running testiso scenarios on metal artifact
    # Skip the following scenarios: iso-install,iso-offline-install,iso-live-login,iso-as-disk
    # See: https://github.com/openshift/os/issues/666
    kola testiso -S --scenarios pxe-install,pxe-offline-install --output-dir tmp/kola-metal
  container:
    from: cosa-build
  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
- as: validate-image-build
  commands: cat /etc/os-release
  container:
    from: rhcos-image
  skip_if_only_changed: ^docs/|\.md$|^(?:.*/)?(?:\.gitignore|OWNERS|PROJECT|LICENSE)$
```

cheesesashimi:
Whereas, I intend for this new trimmed-down script to be run inside the image build context.

miabbott:
I like the proposal here https://github.com/openshift/os/pull/776#issuecomment-1098276986, makes a lot of sense!

/approve

ci/prow-build.sh
```diff
@@ -0,0 +1,56 @@
...
+# https://bugzilla.redhat.com/show_bug.cgi?id=2072050
+curl -L http://base-"${ocpver_mut}"-rhel86.ocp.svc.cluster.local > src/config/ocp.repo
+curl -L http://base-"${ocpver_mut}"-rhel85.ocp.svc.cluster.local > src/config/ocp85.repo
+sed -i -e 's,\[rhel-8-,\[rhel-85-,' src/config/ocp85.repo
+cosa buildfetch --url=${prev_build_url}
+cosa fetch
+cosa build
+cosa buildextend-extensions
+
+# Give the newly-built OCI archive a predictable filename to make OCI archive extraction simpler
```
cgwalters, line 52: Yeah, this intersects with https://github.com/coreos/coreos-assembler/issues/2685 which would lead towards `cosa build ostree` *only* generating an oci dir/archive.

PR #779, opened 2022-04-18
make the prow-build script executable

cheesesashimi:


sohankunkerkar (approved):
/lgtm

sohankunkerkar:
I think we'd need to add a check for non-executable scripts in this repo

PR #781, opened 2022-04-19
refactors the prow build scripts

cheesesashimi:
After a bit of experimenting, it was discovered that:

- Kola tests expect the OCI archive to still have the build number attached to it. This means we'll need to modify the OCI archive extraction / ingestion process slightly.
- Because OpenShift runs a container with a random user ID, we need to ensure that the `builder` user can be looked up by the random OpenShift user ID as well as to ensure that it has the necessary permissions to read and write files within the COSA_DIR directory.

### DIFF
--- a/ci/prow-build.sh
+++ b/ci/prow-build.sh
@@ -48,9 +48,3 @@ cosa buildfetch --url=${prev_build_url}
 cosa fetch
 cosa build
 cosa buildextend-extensions
-
-# Give the newly-built OCI archive a predictable filename to make OCI archive extraction simpler
-arch="x86_64"
-cosa_build_id="$(cat "${COSA_DIR}/builds/builds.json" | jq -r '.builds[0].id')"
-current_build_dir="${COSA_DIR}/builds/latest/${arch}"
-mv "${current_build_dir}/rhcos-${cosa_build_id}-ostree.${arch}.ociarchive" "${current_build_dir}/rhcos.${arch}.ociarchive"
--- a/ci/set-openshift-user.sh
+++ b/ci/set-openshift-user.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+# This script is used to update the /etc/passwd file within the COSA container
+# at test-time. The need for this comes from the fact that OpenShift will run a
+# container with a randomized user ID by default to enhance security. Because
+# COSA runs with an unprivileged user ("builder") instead of (container) root,
+# this presents special challenges for file and disk permissions. This particular
+# pattern was inspired by:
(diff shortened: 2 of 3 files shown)


cgwalters:
> Kola tests expect the OCI archive to still have the build number attached to it. This means we'll need to modify the OCI archive extraction / ingestion process slightly.

Hmm yes.  I think we're potentially losing *all* labels (metadata) with the magic step to import the ociarchive?  But are you actually fixing that in this PR?  If so I don't understand how?

That said
/approve

cheesesashimi:
> Hmm yes. I think we're potentially losing all labels (metadata) with the magic step to import the ociarchive? But are you actually fixing that in this PR? If so I don't understand how?

That's not what this fixes. This separates placing the OCI archive into a static path from the cosa-build process since that is only required by the import process and breaks the Kola tests. I don't think we're losing the labels, but I'll do a check on my next CI run to confirm / deny that.

cheesesashimi:
Just an update: We're not losing the labels. If anything, we're gaining labels from the OpenShift Image Build (which are empty here because this is a rehearsal build or from an inline Dockerfile; not sure which). Here are the labels from an imported image from a previous CI run:

```json
{
  "coreos-assembler.image-config-checksum": "a63a225142788970b32bb6571c3b120811785ac63152b900e588df3db9e0f997",
  "coreos-assembler.image-input-checksum": "78599191ea1a0824158910c172bbbd5229e24b3e757bd47788787c160243bb9a",
  "io.buildah.version": "1.22.4",
  "io.openshift.build.commit.author": "",
  "io.openshift.build.commit.date": "",
  "io.openshift.build.commit.id": "",
  "io.openshift.build.commit.message": "",
  "io.openshift.build.commit.ref": "",
  "io.openshift.build.name": "",
  "io.openshift.build.namespace": "",
  "io.openshift.build.source-context-dir": "",
  "io.openshift.build.source-location": "",
  "org.opencontainers.image.revision": "v0.0.2-574-g196356217e1b2ca4e482268b7d4e97d9474ba38b",
  "org.opencontainers.image.source": "/tmp/cosa/src/config",
  "ostree.bootable": "true",
  "ostree.commit": "e930a0b6d432e48988be0dcca323af71ffec5cf0fddeb13c7f0d8b53c9d7cedc",
  "ostree.diffid": "sha256:e1fc41840b113e551fbaf75bd87accbbff2460c4c84b7a35635a5bcdb6c9cad7",
  "ostree.linux": "4.18.0-369.el8.x86_64",
  "rpmostree.inputhash": "7b81ae96332ece9fb19aefbd1d9761b0c8ead18d8bea7f07299e30ad2ce8fa9c",
  "vcs-ref": "",
  "vcs-type": "",
  "vcs-url": "",
  "version": "411.86.202204191810-0"
}
```

PR #785, opened 2022-04-22
add 25rhcos-azure-udev overlay; drop 25rhcos-azure-udev-rules overlay

dustymabe:
We no longer need the 25rhcos-azure-udev-rules overlay because
the WALinuxAgent-udev package exists in el8. We do still need
the glue code that copies the udev files into the initramfs
via a dracut module, thought. Can drop that in el9.


dustymabe:
For this we'll wait for two things.

1. The move to 8.6 to complete
2. https://github.com/coreos/fedora-coreos-config/pull/1686 to merge so we can bump the f-c-c submodule.

miabbott:
> For this we'll wait for two things.
> 
>     1. The move to 8.6 to complete
> 
>     2. [Revert "overlay: add new module for installing Azure udev rules" coreos/fedora-coreos-config#1686](https://github.com/coreos/fedora-coreos-config/pull/1686) to merge so we can bump the f-c-c submodule.

@dustymabe both of these have completed; I think you can proceed with this PR

travier:
I merged this one into https://github.com/openshift/os/pull/803

bgilbert:
This addresses a 4.11 regression tracked in https://bugzilla.redhat.com/show_bug.cgi?id=2092966.

travier:
/close
Closing as it has been included in https://github.com/openshift/os/pull/815

PR #786, opened 2022-04-22
Bug 2078016: drop multipathd socket workaround

miabbott:
f-c-c submodule bump to drop the workaround:

```
$ (cd fedora-coreos-config && git shortlog 9700b247b8b2d7424beea27df1e7163e619d7d64..67566822f1f957e6484cc3e3960101717dc3b14b)
Micah Abbott (1):
      remove multipathd socket workaround

```


miabbott:
```
--- FAIL: rhcos.upgrade.from-ocp-rhcos (2.57s)
        upgrade.go:214: exit status 1
```

I spent some time debugging this on `build02` by spinning up a `cosa` pod and running this test with a bunch of `fmt.Println()` statements scattered through the code.  (The test fails with no log output otherwise)

Something may have changed in the cluster, as the test appeared to be failing during [the setup of the `oc` client](https://github.com/coreos/coreos-assembler/blob/e7e5dfbebe3ae3e9f48bd8d296190ee569f646bb/mantle/kola/tests/rhcos/upgrade.go#L357-L363).  I suspected it was the `sudo` call to unpack the tarball to `/usr/bin`, so I hacked up the code to explode the tarball into a temp dir and run the `oc` binary from there. Once the `oc` binary was usable, the test seemed to proceed normally.

I suspect that the pod I was using is not identical to the pod being created by Prow, so the `sudo` error might be a red herring.  Though, it looks like the history of the job on PRs to `master` show that `rhcos.upgrade.from-ocp-rhcos` is getting skipped, so perhaps not.

Regardless, I was able to successfully get the upgrade test to run and I'm going to override the test result for now.

/override ci/prow/built-test-qemu

miabbott:
@mike-nguyen FYI on that upgrade test failure; since you wrote the original code for that test, you might be in position to do some additional digging

miabbott:
> * expected dependent [Bugzilla bug 2078013](https://bugzilla.redhat.com/show_bug.cgi?id=2078013) to be in one of the following states: VERIFIED, RELEASE_PENDING, CLOSED (ERRATA), CLOSED (CURRENTRELEASE), but it is ASSIGNED instead

We are aware of the state of the parent BZ for 4.11 and it is agreed that we will remove the workaround in `master` via https://github.com/coreos/fedora-coreos-config/pull/1689 once the 8.5.z update ships this week

PR #789, opened 2022-04-26
faq: Add how to for Kernel change

ravanelli:
- Describe how to change Kernel packages in RHCOS

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -242,3 +242,17 @@ RHCOS inherits the majority of its configuration from Fedora CoreOS, so we aim t
 If the package makes sense to include in Fedora CoreOS, it will ultimately be included in RHCOS in the future when the [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) submodule is updated in this repo.
 
 If the package is **not** included in Fedora CoreOS, you may submit a PR to this repo asking for the inclusion of the package with the reasoning for including it.
+
+## Q: How do I replace the current Kernel with Kernel-RT or a new Kernel version in RHCOS?
+
+The process for Kernel and Kernel-RT is almost the same. For Kernel-RT you need to `override remove` and `install`, and use
+the kernel-rt-kvm package instead of kernel-rt package. As following:
+
+```
+rpm-ostree override remove kernel kernel-core kernel-modules kernel-modules-extra --install kernel-rt-core-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-kvm-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-modules-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-modules-extra-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm
+```
+For the normal Kernel only `override replace` is enough:
+
+```
+rpm-ostree override replace kernel-{,modules-,modules-extra-,core-}4.18.0-305.34.2.107.el8_4.x86_64.rpm
+```


docs/faq.md
```diff
@@ -242,3 +242,12 @@ RHCOS inherits the majority of its configuration from Fedora CoreOS, so we aim t
 If the package makes sense to include in Fedora CoreOS, it will ultimately be included in RHCOS in the future when the [fedora-coreos-config](https://github.com/coreos/fedora-coreos-config) submodule is updated in this repo.
 
 If the package is **not** included in Fedora CoreOS, you may submit a PR to this repo asking for the inclusion of the package with the reasoning for including it.
+
+## Q: How do I replace the current Kernel with Kernel-RT or a new Kernel version in RHCOS?
```
jlebon, line 246: I think the instructions below are only for the first case (moving from kernel to kernel-rt). For changing kernel versions (but not type), `rpm-ostree override replace [list of RPMs]` should be enough.

ravanelli: A great! Thanks for reviewing it @jlebon

docs/faq.md
```diff
@@ -242,3 +242,16 @@ RHCOS inherits the majority of its configuration from Fedora CoreOS, so we aim t
...
+
+The process for Kernel and Kernel-RT is almost the same. For Kernel-RT you need to `override remove` and `install`, as use
```
jlebon, line 248:
```suggestion
The process for Kernel and Kernel-RT is almost the same. For Kernel-RT you need to `override remove` and `install`, and use
```

docs/faq.md
```diff
@@ -242,3 +242,16 @@ RHCOS inherits the majority of its configuration from Fedora CoreOS, so we aim t
...
+the kernel-rt-kvm package instead of kernel-rt package. For the normal Kernel only `override replace` is enough. As following:
+
+```
+rpm-ostree override remove kernel kernel-core kernel-modules kernel-modules-extra --install kernel-rt-core-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-kvm-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-modules-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm --install kernel-rt-modules-extra-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm
+```
```
jlebon, line 253: Would it be clearer if we moved the "For the normal kernel only" sentence to here instead?

cgwalters:
Not a blocker, and thanks for writing this!  Just a reminder to ourselves though that rpm-ostree is also used in e.g. RHEL for Edge, so it'd be good to think about in the future things like "how to use kernel-rt with rpm-ostree" could actually go in some sort of shared docs.

ravanelli:
> Not a blocker, and thanks for writing this! Just a reminder to ourselves though that rpm-ostree is also used in e.g. RHEL for Edge, so it'd be good to think about in the future things like "how to use kernel-rt with rpm-ostree" could actually go in some sort of shared docs.
Any suggestions where this shared doc could be? It would be shared between rpm-ostree repo?

cgwalters:
/lgtm
since this is not at all blocking!

Honestly, I don't know for sure.  There is 
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9-beta/html/composing_installing_and_managing_rhel_for_edge_images/introducing-rhel-for-edge-images_composing-installing-managing-rhel-for-edge-images
Should it go there?  Dunno.  

Let's *definitely* continue to use this repository as a low-overhead way to ship things like this.  I'm just trying to plant the seed that we will have more things like this that we want to share with other RHEL efforts using rpm-ostree.

But again, let's get this in :smile:

PR #791, opened 2022-04-27
manifest: bump to openvswitch2.17

dcbw:
We want to keep host OVS and container OVS synchronized as much
as possible. We've bumped to 2.17 in ovn-kubernetes containers,
plus 2.17 has a number of code sanitizer fixes, and 2.17 will be
the new LTS release that OSP also uses.

@trozet @jcaamano @tssurya


gursewak1997 (approved):
/lgtm

Issue #795, opened 2022-05-06
RHCOS 4.11 failing to build due to missing `qemu-kvm-core` dependency for `kata-containers`


miabbott:
This appears to be fallout from reverting back to RHEL 8.5, so wanted to track this in the open.

```
+ cosa buildextend-extensions
Running: rpm-ostree compose extensions --repo=/home/jenkins/agent/workspace/rhcos-art/rhcos-art-rhcos-4.11/tmp/repo --touch-if-changed tmp/extensions/changed --cachedir=/home/jenkins/agent/workspace/rhcos-art/rhcos-art-rhcos-4.11/cache src/config/manifest.yaml src/config/extensions.yaml --base-rev 271460977b476d0fcd3184d106dbf0c11665af4fcaf5cc42ea35bd727118827d --output-dir tmp/extensions/output
Checking out 2714609... done!
warning: Found bdb_ro Packages database while attempting sqlite backend: using bdb_ro backend.
Enabled rpm-md repositories: rhel-8-server-ose rhel-8-fast-datapath rhel-8-baseos rhel-8-appstream rhel-8-nfv
Updating metadata for 'rhel-8-server-ose'...done
Updating metadata for 'rhel-8-fast-datapath'...done
Updating metadata for 'rhel-8-baseos'...done
Updating metadata for 'rhel-8-appstream'...done
Updating metadata for 'rhel-8-nfv'...done
warning: Found bdb_ro Packages database while attempting sqlite backend: using bdb_ro backend.
Importing rpm-md...done
rpm-md repo 'rhel-8-server-ose'; generated: 2022-05-05T16:01:36Z solvables: 426
rpm-md repo 'rhel-8-fast-datapath'; generated: 2022-04-25T14:15:35Z solvables: 736
rpm-md repo 'rhel-8-baseos'; generated: 2022-05-04T07:23:27Z solvables: 11158
rpm-md repo 'rhel-8-appstream'; generated: 2022-05-05T13:46:51Z solvables: 24626
rpm-md repo 'rhel-8-nfv'; generated: 2022-04-26T16:45:33Z solvables: 610
Resolving dependencies...done
[0m[31merror: [0mCould not depsolve transaction; 1 problem detected:
 Problem: conflicting requests
  - nothing provides qemu-kvm-core >= 15:6.2.0 needed by kata-containers-2.4.1-1.el8.x86_64
Traceback (most recent call last):
```

[...]

PR #796, opened 2022-05-06
[rhel8.5 revert] Use RHEL8.6 repo for kata-containers

mike-nguyen:
Reverting back to RHEL 8.5 content caused extension building to fail
because the newer kata-container has a requirement on a qemu-kvm-core
version that is newer than the one in the rhel-8-advanced-virt repo.

Get the kata-containers dependencies from RHEL8.6 appstream.

See: https://github.com/openshift/os/issues/795


miabbott:
Should be able to target `master` with this PR, right?

miabbott:
I think you'll need to change the CI script to `curl` down the 8.6 repos and populate the `ocp.repo` file accordingly

miabbott:
> I think you'll need to change the CI script to `curl` down the 8.6 repos and populate the `ocp.repo` file accordingly

I think I have a one-liner to do this, will push a commit to this PR

mike-nguyen:
Second commit looks good to me

HuijingHei (approved):
LGTM

PR #798, opened 2022-05-11
bump fedora-coreos-config

jschintag:
```
git shortlog --no-merges 812192f5781d6cb39c0edce764d95c01d75dabe7..0856e21a48d389409f55eb2da8fa6ca9715a0991
Benjamin Gilbert (4):
      tests/kola/ignition: port resource fetching tests from mantle
      tests/kola/ignition: add tests for s3:// versioned resource fetch
      tests/kola/ignition: move S3 test fixtures to a new bucket
      tests/kola/ignition: finish moving S3 test fixtures to a new bucket

CoreOS Bot (6):
      lockfiles: bump to latest
      overrides: fast-track rust-afterburn-5.3.0-1.fc35
      lockfiles: bump to latest
      lockfiles: bump timestamp
      lockfiles: bump to latest
      lockfiles: bump to latest

Dusty Mabe (8):
      overrides: fast-track systemd-249.12-3.fc35
      Revert "15fcos: add coreos-fix-systemd-journald.service"
      manifests: add curl to file-transfer manifest
      make kdump work on first boot
      tests: kdump.crash: extend timeout
      tests: kdump.crash: update crashkernel size based on our docs
      manifests/fedora-coreos: fix kdump.crash test for aarch64 AWS
      manifests/fedora-coreos: switch grub2-removals to conditional-include

Jan Schintag (1):
      clhm.ignition-warnings: Improve test to ignore additional warnings

gursewak1997 (1):
      kola/systemd-enable-units: remove journal check Instead of writing to journal, we just check a file the service should have touched and test that it exists.
```


jschintag:
This should fix `clhm.ignition-warnings` for both ppc64le and s390x

jschintag:
Hmm, do we still have workarounds/problems with fwupd on x86?
```
+ '[' '!' -f /usr/lib/modules-load.d/fwupd-redfish.conf ']'
+ grep -q ipmi-si /usr/lib/modules-load.d/fwupd-redfish.conf
+ exit 100
```

jschintag:
Well i can just say that this PR works on s390x, and the error seems to be a general problem for x86 right now.

miabbott:
> ```
> + '[' '!' -f /usr/lib/modules-load.d/fwupd-redfish.conf ']'
> + grep -q ipmi-si /usr/lib/modules-load.d/fwupd-redfish.conf
> + exit 100
> ```

https://github.com/openshift/os/blob/master/manifest.yaml#L123-L126

This indicates that RHCOS is pulling in the `fwupd` package from RHEL 8.6....which means that the CI job is using the 8.6 repos and doesn't match what we are using downstream.

We have to adjust the manifest/repos used to pull the 8.5 content for now, until we return to 8.6

PR #800, opened 2022-05-16
manifest: pull container-tools content from RHAOS repo

miabbott:
We've worked with the Containers team to enable builds of certain
container-tools RPMs from the RHAOS repo. This allows us to land OCP
specific changes to these packages and potentially ship newer versions
of these packages than what would available in the module.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -353,32 +353,24 @@ repo-packages:
       - nss-altfiles
   - repo: rhel-8-server-ose
     packages:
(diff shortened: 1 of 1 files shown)


miabbott:
In draft state until we land 8.6, as `podman` v4 has dependencies that are only available in 8.6

travier:
/close
Closing as it has been included in https://github.com/openshift/os/pull/815

PR #803, opened 2022-05-23
manifests: Rework to allow co-located builds for each RHEL version

travier:
manifests: Split into common & RHEL 8.6 specific manifests

---

Keep RHEL 8.6 based RHCOS as the default

---

ci: Regroup tests in qemu & metal buckets

- Only build the ostree aci container archive in the build step as it's
  the only artifact used for the layering test.
- Group the remaining tests in two buckets: qemu & metal.
- Disable the now uneeded tests.

---

ci: Make RHEL version explicit

Prepare for testing multiple RHEL versions.

---

ci: Prepare test names

Pre-create test names to be able to enable them in Prow before we merge
the corresponding test code.

---

Enable building RHCOS from different RHEL minor releases from the same branch and repo. This will enable us to work on RHEL 8.6 & 9.0 based RHCOS in parallel instead of using different branches or costly reverts.

Each RHEL minor release based RHCOS is now under a subfolder.

To try it out:

```
# RHEL 8.6 based RHCOS
$ mkdir rhcos-rhel-8.6 && cd rhcos-rhel-8.6 && cosa init https://github.com/travier/os -b splitmanifest rhel-8.6

# Upcoming PR: RHEL 9.0 based RHCOS
$ mkdir rhcos-rhel-9.0 && cd rhcos-rhel-9.0 && cosa init https://github.com/travier/os -b splitmanifest rhel-9.0
```


HuijingHei:
The whole looks good.
Just one small question, maybe should based on the latest patch, missed PR https://github.com/openshift/os/pull/791

ci/build-test-qemu-init.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# Create a temporary cosa workdir if COSA_DIR is not set.
+cosa_dir="${COSA_DIR:-$(mktemp -d)}"
+echo "Using $cosa_dir for build"
+cd "$cosa_dir"
+
+# Require that the RHEL version to use to build RHCOS is passed as first
+# argument
+if [[ "${#}" -ne 1 ]]; then
+    echo "No RHEL release selected. Can not build RHCOS!"
```
cgwalters, line 17: Maybe this is is a bit clearer?

```suggestion
    echo "Missing required version parameter, e.g.: $0 rhel-8.5" 1>&2
```

rhel-8.5.yaml
```diff
@@ -0,0 +1,177 @@
...
+automatic-version-suffix: "-"
+# Keep this is sync with the version in postprocess
+mutate-os-release: "4.11"
+
+postprocess:
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+
+     # Tweak /usr/lib/os-release
```
cgwalters, line 29: Can't all this stuff be in `common.yaml` too?

travier: It depends on how much we want to change things here for SCOS. Right now it's a copy paste for SCOS but maybe it should not.

travier:
> It looks like the default is still 8.5 for the record?

Yes, I would prefer to land this one on 8.5 and then try the switch to 8.6 again.

> I think one thing we should do pretty quickly after this lands is adjust the CI to cover both cases (at least 8.6 optionally) and also add a periodic that builds the non-default cases.

Agree.

travier:
> The whole looks good. Just one small question, maybe should based on the latest patch, seems missed PR #791

Thanks, fixed.

HuijingHei (approved):

HuijingHei:
See error about this, do we need to add the workaround?
```
+ cosa buildextend-metal
Config commit: fatal: unsafe repository ('/tmp/cosa/src/config' is owned by someone else)
To add an exception for this directory, call:
	git config --global --add safe.directory /tmp/cosa/src/config
```

ci/build-test-qemu-init.sh
```diff
@@ -0,0 +1,23 @@
...
+set -xeuo pipefail
+
+dn=$(dirname $0)
+
+# Prow jobs don't support adding emptydir today
+export COSA_SKIP_OVERLAY=1
+
+# Create a temporary cosa workdir if COSA_DIR is not set.
+cosa_dir="${COSA_DIR:-$(mktemp -d)}"
+echo "Using $cosa_dir for build"
```
HuijingHei, line 11: maybe try add this to see if it works:
`git config --global --add safe.directory ${cosa_dir}`

cgwalters:
I went ahead and pushed https://github.com/openshift/os/pull/810 to this PR.

cgwalters:
I'm OK with this as is, though I think long term it will be cleaner to do the same as fedora-coreos-config instead of having subdirectories.  Which means we will either need to branch as we do there (e.g. with `main-rhel-9` type branch) *or* (my preference) close the last gap and have [coreos-assembler know how to take an optional build type](https://github.com/coreos/coreos-assembler/pull/1459), and use that to e.g. choose which `manifest.yaml` and `image.yaml` to use.

cgwalters:
Are we intending to block on getting this in for 4.11 to do the 8.6 switch?  Or are we going to try to re-revert again?

cgwalters:
Since we know we need to ship 8.6, I think my vote is:

- Do the reverts again to get us back to 8.6 on master *without* landing this PR
- Cherry pick to release-4.11 (also we need to evaluate what *other* things should be cherry picked)
- Look at landing this PR on master

miabbott:
>     * Cherry pick to release-4.11 (also we need to evaluate what _other_ things should be cherry picked)

Looking at the diff between `master` and `release-4.11`, I think we should probably pull in a939a71fe653a64565913108edd24f46894335ec and fe2507957b77a7fc0b5e90308499205a09781003 to 4.11

mike-nguyen:
Building SCOS live ISO, I got
```
$ cosa buildextend-live
Targeting build: 411.91.202206031601-0
  File "/usr/lib/coreos-assembler/cmd-buildextend-live", line 55, in <module>
    raise Exception(f"missing directory {srcdir_prefix}")
Exception: missing directory src/config/live/
```

travier:
> ```
> Exception: missing directory src/config/live/
> ```

Looks like we need more symlinks.

cgwalters:
OK so...since we've implicitly branched for 4.11.  I'd be fine trying to get this merged pretty soon, otherwise it will be a continual conflict fest.

jlebon (commented):
Cool, this looks sane to me overall. Though it's tricky to review the manifest file shuffling to make sure nothing is lost, but one test we could do is to do an RHCOS compose before and after this and then do `ostree diff` to verify nothing unexpected changed.

Also, since we've branched, could we delete the 8.5 stuff now? If we really need to revert for 4.11, we can revert in that branch only.

rhel-8.5/extensions.yaml
```diff
@@ -0,0 +1 @@
+../extensions-rhel-8.5.yaml
```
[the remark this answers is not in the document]
travier: I was heading in the direction that the sub-dirs are essentially "empty" or symlinks only and everything meaningful is in the main directory but we can do it differently.

travier:
Will do a manual rebase soon

rhel-8.5/extensions.yaml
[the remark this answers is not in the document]
travier: If we go this route then we can move the entire version specific manifest will end up in the subdir and include from the common one.

ci/prow-entrypoint.sh
```diff
@@ -154,13 +158,23 @@ main () {
             cosa_init
             cosa_build
             ;;
+<<<<<<< HEAD
```
LorbusChris, line 161: @miabbott it looks like something went wrong here while rebasing

rhel-8.6/manifest.yaml
```diff
@@ -0,0 +1,187 @@
...
+  - rhel-8-baseos
+  - rhel-8-appstream
+  - rhel-8-fast-datapath
+  - rhel-8-server-ose
+
+# https://bugzilla.redhat.com/show_bug.cgi?id=1938928
+rpmdb: bdb
+
+# We include hours/minutes to avoid version number reuse
+automatic-version-prefix: "412.86.<date:%Y%m%d%H%M>"
```
jlebon, line 35: This is bumping the version from 411 to 412. For commits like this which move a lot of pieces around, it'd be better to keep the end state functionally the same and split out this kind of changes in separate commits with their own rationales.

common.yaml
```diff
@@ -313,13 +186,8 @@ packages:
...
  # Extras
  - systemd-journal-gateway
- # RHEL7 compatibility
- - compat-openssl10
- # Make sure we pull in at least clevis 15; it drops the rd.neednet=1 hardcode
- # and has a few other patches we need.
- # https://bugzilla.redhat.com/show_bug.cgi?id=1853651
- - "'clevis >= 15-1.el8' 'clevis-luks >= 15-1.el8' 'clevis-dracut >= 15-1.el8'"
- - cryptsetup-reencrypt tpm2-tools
+ - clevis clevis-luks clevis-dracut
```
jlebon, line 189: The dropping of clevis versioning could be a separate commit.

common.yaml
```diff
@@ -313,13 +186,8 @@ packages:
  - systemd-journal-remote
  # Extras
  - systemd-journal-gateway
- # RHEL7 compatibility
- - compat-openssl10
- # Make sure we pull in at least clevis 15; it drops the rd.neednet=1 hardcode
- # and has a few other patches we need.
- # https://bugzilla.redhat.com/show_bug.cgi?id=1853651
- - "'clevis >= 15-1.el8' 'clevis-luks >= 15-1.el8' 'clevis-dracut >= 15-1.el8'"
- - cryptsetup-reencrypt tpm2-tools
```
jlebon, line 322: This is dropping cryptsetup-reencrypt. Is that on purpose? Would be better to have a separate commit with a rationale.

manifest.yaml
```diff
@@ -0,0 +1,3 @@
+# Default RHEL version used to build RHCOS
```
jlebon, line 1: Are we OK with the concept of a "default manifest"? Might be cleaner to just rip off the bandaid and require users/pipelines being more specific going forward. I understand wanting to minimize disruption though. Though... we did fork already for 4.11, so seems OK if this breaks things and we fix them.

miabbott:
Split up the commits some more per @jlebon feedback.  I sanity checked that this builds with RHEL 8.6 + `kola run` passes.

miabbott:
I fought the black box of Prow for most of today trying to get this PR to work, but keep bumping into problems where the symlinked files to/from the `rhel-8.6` subdir cannot be found during the test:

```
Formatting 'cache2.qcow2.tmp', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=10737418240 lazy_refcounts=off refcount_bits=16
Disabled fsync on cache repositories
error: Can't open file "/tmp/cosa/src/config/common.yaml" for reading: No such file or directory (os error 2) 
```

Interestingly, if I `cosa init` from a git URL where this change lives, things seem to work fine.  But `cosa init` from the copied directory runs into the described problem.

I've not yet uncovered a reasonable workaround for this, which allows for using the new layout in both local use cases and the Prow case.

miabbott:
> The problem likely has something to do with the special cases in the virtio bits related to src/config - see [coreos/coreos-assembler#2774](https://github.com/coreos/coreos-assembler/pull/2774)
> 
> In fact I bet its: "Note: This only works for absolute symlinks today." right?

I've not tested the use of absolute symlinks here, but this sounds likely

miabbott:
I hacked up `cosa init` to work in this situation of local source dir + subdirectory, which seems to get me out of the described problems, but I'm not confident it won't break existing workflows that use the source dir + subdir setup:


```patch
From 62bb26c31a5c797d2414177319aa0c389c59fa81 Mon Sep 17 00:00:00 2001
From: Micah Abbott <miabbott@redhat.com>
Date: Wed, 15 Jun 2022 16:49:22 -0400
Subject: [PATCH] cmd-init: handle local subdir case

When doing `cosa init` where the source is a local directory and the
subdir is specified, the resulting COSA_DIR may not be usable.

Adjust the logic to setup symlinks in this case correctly.
---
 src/cmd-init | 9 +++++++--
 1 file changed, 7 insertions(+), 2 deletions(-)

diff --git a/src/cmd-init b/src/cmd-init
index 654751144..26aa6054e 100755
--- a/src/cmd-init
+++ b/src/cmd-init
@@ -14,7 +14,7 @@ TRANSIENT=0
 print_help() {
     cat 1>&2 <<'EOF'
 Usage: coreos-assembler init --help
-       coreos-assembler init [--force] [--transient] [--branch BRANCH] 
+       coreos-assembler init [--force] [--transient] [--branch BRANCH]
                              [--commit COMMIT] GITCONFIG [SUBDIR]
 
   For example, you can use https://github.com/coreos/fedora-coreos-config
@@ -112,7 +112,12 @@ mkdir -p src
 (cd src
  if ! test -e config; then
      case "${source}" in
-         /*) ln -s "${source}/${subdir}" config;;
+         /*) ln -s "${source}" config
+            if [ -n "${subdir}" ]; then
+                mv config config-git
+                ln -sr config-git/"${subdir}" config
+            fi
+         ;;
          *) git clone ${BRANCH:+--branch=${BRANCH}} --depth=1 --shallow-submodules --recurse-submodules "${source}" config
             # If a commit was specified then we'll fetch and reset
             # the specified branch to that commit. This is useful when
-- 
2.36.1
```

miabbott:
> I hacked up `cosa init` to work in this situation of local source dir + subdirectory, which seems to get me out of the described problems, but I'm not confident it won't break existing workflows that use the source dir + subdir setup:

Hmph, well it seemed to work locally, but not in a pod on build02.  Guess that is a sign of my brain being fried for the day.

cgwalters:
If you want to test what is happening in Prow, you'll need to be sure to
`export FORCE_UNPRIVILEGED=1` to ensure everything goes through qemu/supermin.

miabbott:
I tested two additional commits on this branch that got things working for the `FORCE_UNPRIVILEGED=1` case, but the additional changes feel like the approach is wrong.  See the top commits on my branch - https://github.com/miabbott/os/commits/splitmanifest

My branch also works in the local, unprivileged container case.

@travier WDYT?

miabbott:
> I tested two additional commits on top of this branch that got things working for the `FORCE_UNPRIVILEGED=1` case, but the additional changes feel like the approach is wrong. See the top commits on my branch - https://github.com/miabbott/os/commits/splitmanifest

Maybe we should invert the relationship here, where we just pick a default and symlink `manifest.yaml` and friends at the top-level to the contents in `rhel-8.6`?

I think this is what Jonathan suggested https://github.com/openshift/os/pull/803#discussion_r896852512

miabbott:
> I tested two additional commits on top of this branch that got things working for the `FORCE_UNPRIVILEGED=1` case, but the additional changes feel like the approach is wrong. See the top commits on my branch - https://github.com/miabbott/os/commits/splitmanifest
> 
> My branch also appears to work in the local, unprivileged container case.
> 
> @travier WDYT?

It's ugly but it seems to work - https://github.com/openshift/os/pull/849

travier:
I have another bug with cosa replacing the config -> config-git symlink to an absolute symlink when buiding locally.
I think we should drop the sub dir setup and go back to a simpler per repo symlink setup.

travier:
Prep commits split into https://github.com/openshift/os/pull/850

travier:
Alternative option in https://github.com/openshift/os/pull/852

manifest.yaml
travier: Due to https://github.com/coreos/coreos-assembler/commit/480c2396977921ffd6cbc4ea44b3fcd2c470c954 we currently need a default. We could however change cosa to not require that.

common.yaml
travier: Moved to https://github.com/openshift/os/pull/850

common.yaml
travier: Moved to https://github.com/openshift/os/pull/850

rhel-8.6/manifest.yaml
travier: And indeed, agree that we should not sneak changes in. With the many rebases and tests I did I forgot to split those

jlebon:
https://github.com/coreos/coreos-assembler/pull/2936 should make cosa handle symlinks better.

PR #816, opened 2022-06-03
Bug 2068148: /etc/redhat-release symlink is broken

marmijo:
The symbolic links for /etc/system-release and /etc/redhat-release
are pointing a non-existent file and are absolute sym links which
is breaking some workflows / use cases. Update the links to point
to a valid file and make them relative.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -167,8 +167,8 @@ postprocess:
      ${NAME} release ${VERSION_ID}
      EOF
      rm -f /etc/system-release /etc/redhat-release
-     ln -s /usr/lib/system-release /etc/system-release
-     ln -s /usr/lib/system-release /etc/redhat-release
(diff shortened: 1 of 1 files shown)


marmijo:
https://bugzilla.redhat.com/show_bug.cgi?id=2068148 also references another issue https://github.com/openshift/sriov-network-operator/issues/647 which has already been resolved.

marmijo:
I just noticed [PR 803](https://github.com/openshift/os/pull/803) which may resolve this bug. Should I retract this PR?

miabbott:
> I just noticed [PR 803](https://github.com/openshift/os/pull/803) which may resolve this bug. Should I retract this PR?

It does look like 803 is going to accomplish the same thing; I think it would be safe to close this PR and let's focus on reviewing and merging 803.

PR #821, opened 2022-06-06
refactors cosa session initialization

cheesesashimi:
Right now, the OpenShift CI jobs attached to this repository have the bottleneck of running `$ cosa build` within a container image build context. Because of how the OpenShift CI system is designed, it's not likely that we can use KVM within the container image build process to accelerate things. Despite this limitation, I've noticed that while the `cosa-build` image build itself usually does not time out, the 2+ hours it takes to run leaves us with very little headroom to run our test stages afterward. Despite this, we still need to produce the artifact in `cosa-build` in the way that we do. Here's how I plan to address this:

1. Separate the `$ cosa init` and `$ cosa build` steps in https://github.com/openshift/os/blob/master/ci/prow-build.sh.
2. Add another image build in openshift/release that only does the `$ cosa init` stuff (lets call this image `cosa-init` for now). This ensures we're using the same COSA session across all of the build and test stages.
3. Pass the `cosa-init` image into the current image build flow to do `$ cosa build` (this becomes the input to the `cosa-build` image build). This will still be the image that gets promoted if all the required tests pass.
4. We also pass the `cosa-init` image into the `test-qemu-*` test stages. Each test stage will run its own `$ cosa build` (with KVM acceleration) and execute the kola tests (also with KVM acceleration). These stages will run as soon as the `cosa-init` image is built and will run concurrently with the `cosa-build` image build.

This PR implements point 1) above by consolidating the `$ cosa init` (and the rest of the preparation steps) into the `ci/prow-prepare.sh` script. This also modifies `ci/prow-build-test-qemu.sh` and `ci/prow-build.sh` to make use of the new consolidated initialization flow. Points 2-4 are validated by https://github.com/openshift/release/pull/29187.

Once this PR is merged, we'll need another PR to openshift/release to add the `cosa-init` stage.

### DIFF
--- a/ci/build-test-qemu.sh
+++ b/ci/build-test-qemu.sh
@@ -5,8 +5,6 @@ dn=$(dirname $0)
 # Prow jobs don't support adding emptydir today
 export COSA_SKIP_OVERLAY=1
 # Create a temporary cosa workdir if COSA_DIR is not set.
-cosa_dir="${COSA_DIR:-$(mktemp -d)}"
-echo "Using $cosa_dir for build"
-cd "$cosa_dir"
-cosa init --transient /src
-exec ${dn}/prow-build-test-qemu.sh
+export COSA_DIR="${COSA_DIR:-$(mktemp -d)}"
+cd "$COSA_DIR"
+exec "${dn}/prow-build-test-qemu.sh"
--- a/ci/cosa-build.sh
+++ b/ci/cosa-build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -xeuo pipefail
+
+# This script assumes that `cosa init` has been run.
+cosa build
+cosa buildextend-extensions
--- a/ci/prow-build-test-qemu.sh
+++ b/ci/prow-build-test-qemu.sh
@@ -1,49 +1,12 @@
 #!/bin/bash
 set -xeuo pipefail
-# This script is called via build-test-qemu.sh which is the main Prow
-# entrypoint for PRs to this repo, as well as for PRs on other repos,
-# mainly coreos-assembler.  It assumes that `cosa init` has been run.
 
-REDIRECTOR_URL="https://rhcos-redirector.apps.art.xq1c.p1.openshiftapps.com/art/storage/releases/"
+dn="$(dirname "$0")"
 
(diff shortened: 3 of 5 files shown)


cgwalters:
>  Each test stage will run its own $ cosa build (with KVM acceleration)

Wait is that happening today?  Why?  I'd expect us to have a single built ostree image and a single qemu disk image that we test.

travier:
See https://github.com/openshift/release/pull/29187

cheesesashimi:
> Wait is that happening today? Why? I'd expect us to have a single built ostree image and a single qemu disk image that we test.

What's happening today is that the ostree image is being built within a container image build stage. We're seeing a lot of test timeouts because we can't use KVM in an image build to accelerate the build process (this is a limitation of OpenShift CI). So while the image builds themselves are not timing out, the time they take to build leaves us with very little headroom by the time the test stages run. However, since this is the only reasonable way to get an RHCOS container image into the ImageStream, we still need to do this in an image build.

What this PR sets the stage for is a followup PR to openshift/release. The followup PR will use these scripts to create a COSA session in the image build stage and pass that off to the kola test stages. Each test stage will run `$ cosa build` and its test suite. These will run concurrently with the `cosa-build` image build stage. The reasoning behind this is the test stage can perform the build much faster than the image build stage can; however we still need the RHCOS image produced by the `cosa-build` image build stage.

LorbusChris:
@cheesesashimi it should be configureable to expose KVM on CI build farms, but I'm not sure that's already the case on all of them. If the builds are scheduled on clusters where KVM is not exposed, we should be able to configure them to be scheduled on one where KVM is actually exposed.

cheesesashimi:
@LorbusChris I know that's possible for the test stages and we do make use of that already. However, I don't know if that is possible for the image build stages. If my memory serves correctly, even if the image build pods have KVM exposed, I don't think the image build context can make use of it.

cgwalters:
Looks like this needs the `export COSA_SKIP_OVERLAY=1` bits

cheesesashimi:
I'm thinking we may want to combine this with https://github.com/openshift/os/pull/827.

travier:
Replaced by https://github.com/openshift/os/pull/839
/close

Issue #818, opened 2022-06-06
RHCOS 4.11 + RHEL 8.6 x86_64 - `crio.base/networks-reliably` failing


miabbott:
```
    --- FAIL: crio.base/networks-reliably (30.15s)
            cluster.go:162: E0606 18:11:00.483379    4755 remote_runtime.go:201] "RunPodSandbox from runtime service failed" err="rpc error: code = DeadlineExceeded desc = context deadline exceeded"
            cluster.go:162: time="2022-06-06T18:11:00Z" level=fatal msg="run pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded"
            cluster.go:184: "sudo crictl runp ping7Pod265831711" failed: output , status Process exited with status 1
```

Not clear what's happening, but noting for posterity.

From `journal.txt`

```
Jun  6 18:10:56.460581 sudo[4739]:     core : TTY=unknown ; PWD=/var/home/core ; USER=root ; COMMAND=/bin/crictl runp ping7Pod265831711
Jun  6 18:10:56.462785 sudo[4739]: pam_systemd(sudo:session): Cannot create session: Already running in a session or user slice
Jun  6 18:10:56.463263 sudo[4739]: pam_unix(sudo:session): session opened for user root by (uid=0)
Jun  6 18:10:56.483334 crio[2533]: time="2022-06-06 18:10:56.482523657Z" level=info msg="Running pod sandbox: //POD" id=1a3f3498-7543-408c-b5d4-9fa22f1d670b name=/runtime.v1.RuntimeService/RunPodSandbox
Jun  6 18:10:56.483334 crio[2533]: time="2022-06-06 18:10:56.482853456Z" level=warning msg="Allowed annotations are specified for workload [io.containers.trace-syscall]"
Jun  6 18:10:59.133675 crio[2533]: time="2022-06-06 18:10:59.133640473Z" level=info msg="Got pod network &{Name:rhcos-crio-pod-ping7 Namespace:redhat.test.crio ID:d34a9c6c53ac46b3868a2729e3103ef9d0b441b9b9981dd465e51d32c0a88eb0 UID: NetNS:/var/run/netns/7f55df31-752c-48fa-b817-8604fa8462b8 Networks:[] RuntimeConfig:"
[...]

PR #823, opened 2022-06-06
denylist: Add crio.base

cgwalters:
We have an inbound hacky fix for https://github.com/openshift/os/issues/818
in https://github.com/coreos/coreos-assembler/pull/2906
but...let's not even block on that.

I am not totally sure it makes sense for us to test crio ourselves;
this test has I think found one real bug, but otherwise has
just been noise or bugs in the test.

I think we should refocus on testing crio-via-kubelet-via-OCP in
more production paths.

Anyways for now, denylist it.


travier (commented):
/lgtm

PR #826, opened 2022-06-07
Bug 2094385: revert the rebase to RHEL 8.6  [release-4.11]

miabbott:
This reverts #817

Needs a downstream revert of the repos, too


cgwalters:
OK so we're narrowing in on a recent crio change as the culprit here.  And that change sounds like it was already in the process of being backported to 4.10 even.

Actually another way to say this is - I don't think this is related to RHEL 8.6 at all; crio lives in the OCP 4.11 channel. It just literally happened to land at the same time.

So...I think we should close this.

miabbott:
> OK so we're narrowing in on a recent crio change as the culprit here. And that change sounds like it was already in the process of being backported to 4.10 even.
> 
> Actually another way to say this is - I don't think this is related to RHEL 8.6 at all; crio lives in the OCP 4.11 channel. It just literally happened to land at the same time.
> 
> So...I think we should close this.

The culprit appears to be `conmon`, but the reasoning is the same:  `conmon` comes into RHCOS via OCP repos.  Reverting to an older version of RHEL is going continue to pull in the same broken `conmon`, so it would not accomplish the intended effect.

/close

PR #827, opened 2022-06-07
set COSA_DIR in Dockerfile

cheesesashimi:
With the introduction of https://github.com/openshift/release/pull/29187, the test stages no longer have an explicitly set COSA_DIR value. Instead, they are calling `$ mktemp -d`.

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -8,5 +8,14 @@ RUN go test -v -c -o layering_test ./tests/layering
 
 FROM registry.svc.ci.openshift.org/coreos/coreos-assembler:latest
 WORKDIR /src
+ENV COSA_DIR=/tmp/cosa
 COPY --from=builder /go/src/github.com/openshift/os/layering_test /usr/local/bin/layering_test
 COPY . .
+RUN mkdir -p "${COSA_DIR}"
+# We need to make sure that root can read / write to the COSA_DIR so that
+# when this container is actually run, we have permissions to read and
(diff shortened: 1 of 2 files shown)


miabbott:
` /src/ci/prow-build.sh: line 9: cd: /tmp/cosa: No such file or directory `

Looks like we need to create it in the `prow-build.sh` script?

cheesesashimi:
> Looks like we need to create it in the prow-build.sh script?

We can also create it inside the container image. I ran into the same permissions issues I ran into when I was setting up the CI. I fixed it though.

cheesesashimi:
This is still showing the same failure mode of `qemu-img: /var/tmp/mantle-qemu3810394450/disk2048124430: Image creation needs a size parameter`. Is it possible that the qemu-img version in the coreos-aseembler image changed and now requires an additional parameter?

cgwalters:
```
 + cosa buildextend-metal
fatal: /tmp/cosa must be a volume 
```

yet...we didn't exit out?

cgwalters:
I pushed two followup commits here, hope that's OK!

travier (approved):
/lgtm

cgwalters:
OK cool, so it's just the last bit of `test-in-cluster` which is https://github.com/openshift/os/issues/829

cheesesashimi:
> I pushed two followup commits here, hope that's OK!

I was about to add `ENV COSA_SKIP_OVERLAY=1`, but you beat me to it!

PR #831, opened 2022-06-08
[4.10] ci: Work around git checkout security check

travier:
Manual backport of https://github.com/openshift/os/pull/814

### DIFF
--- a/ci/validate.sh
+++ b/ci/validate.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 set -xeuo pipefail
+
+# Create a temporary copy
+workdir="$(mktemp -d)"
+echo "Using $workdir as working directory"
+cd "$workdir"
+git clone /go/src/github.com/openshift/os os
+cd os
 # First ensure submodules are initialized
 git submodule update --init --recursive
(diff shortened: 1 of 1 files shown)


miabbott:
/lgtm

The `validate` job appears to be using the CI job config for `master`, not `release-4.10

/retest

miabbott:
> /lgtm
> 
> The `validate` job appears to be using the CI job config for `master`, not `release-4.10
> 
> /retest

/lgtm cancel

Ugh, too early in the morning to do this....

The commit is using `cosa init --transient` which isn't on the `rhcos-4.10` branch of `cosa`

miabbott:
@travier we'll need a different workaround for the unsafe `git` checkout problem

miabbott:
```
+ git clone /src
fatal: repository '/src' does not exist
```

Looks like the correct path would be `/go/src/github/openshift/os` based on https://github.com/openshift/os/blob/21ea2730e63af11e963fa77985f4110a58fc91f4/ci/Dockerfile#L14

travier:
OK, looks like this workaround works

PR #838, opened 2022-06-09
Modify CI scripts to use RHEL 8.6 repos

miabbott:
We've switched to using RHEL 8.6 in the manifest + downstream, so update the CI scripts accordingly.

### DIFF
--- a/ci/prow-build-test-qemu.sh
+++ b/ci/prow-build-test-qemu.sh
@@ -34,9 +34,7 @@ fi
 ocpver=$(rpm-ostree compose tree --print-only src/config/manifest.yaml | jq -r '.["mutate-os-release"]')
 ocpver_mut=$(rpm-ostree compose tree --print-only src/config/manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|')
 prev_build_url=${REDIRECTOR_URL}/rhcos-${ocpver}/
-# we want to use RHEL 8.5 for testing until we can start using 8.6
-# see https://github.com/openshift/release/pull/26193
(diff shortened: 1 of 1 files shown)


ci/prow-build-test-qemu.sh
```diff
@@ -34,9 +34,7 @@ fi
 ocpver=$(rpm-ostree compose tree --print-only src/config/manifest.yaml | jq -r '.["mutate-os-release"]')
 ocpver_mut=$(rpm-ostree compose tree --print-only src/config/manifest.yaml | jq -r '.["mutate-os-release"]' | sed 's|\.|-|')
 prev_build_url=${REDIRECTOR_URL}/rhcos-${ocpver}/
-# we want to use RHEL 8.5 for testing until we can start using 8.6
-# see https://github.com/openshift/release/pull/26193
-curl -L http://base-"${ocpver_mut}"-rhel85.ocp.svc.cluster.local > src/config/ocp.repo
+curl -L http://base-"${ocpver_mut}"-rhel86.ocp.svc.cluster.local > src/config/ocp.repo
```
jlebon, line 37: Idea: we can parse the RHEL version from `automatic-version-prefix`.

OK, this works:

```
rhelver=$(rpm-ostree compose tree --print-only manifest.yaml | jq -r '.["automatic-version-prefix"]' | cut -f2 -d.)
```

With https://github.com/openshift/os/pull/803, it becomes a variable in the manifest so it's even more accessible. But sadly rpm-ostree doesn't reserialize variables, so we'd have to parse the YAML directly. We could change that though.

jlebon (approved):
One idea, but LGTM as is too.

travier:
Merging this one as part of https://github.com/openshift/os/pull/839

miabbott:
/close

Preferring #839

PR #840, opened 2022-06-09
moves scripts and Dockerfiles into os repo

cheesesashimi:
Moves the Dockerfiles and test scripts from the openshift/release repo.


miabbott:
Plan is to merge this first, then merge openshift/release#29318 and plow forward.  CI is already broken as is, so we are going override a lot of these jobs.

/lgtm

PR #842, opened 2022-06-09
OWNERS: remove OWNERS file to rely on openshift/release aliases

jmarrero:
On https://github.com/openshift/ci-tools/tree/master/cmd/autoowners states:
`If the upstream does not have an OWNERS file, the utility will ignore syncing it.`

We want to use the ALIASES already defined on:
https://github.com/openshift/release/blob/master/OWNERS_ALIASES#L90

The PR to update the OWNERS on the release repo is: https://github.com/openshift/release/pull/29338


bgilbert:
CI passed, and the rerun isn't going to find any additional dependencies on the `OWNERS` file.

/override ci/prow/periodic-images

PR #844, opened 2022-06-10
OWNERS: restore + add some contextual notes

miabbott:
Turns out that OWNERS is really important, so restore it and add some notes as to why.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,4 +1,13 @@
+# This file is required in order to control `/approve` and `/lgtm` and related
+# Prow commands on this repo.  It may contain non-Openshift org members and
+# aliases (defined in OWNERS_ALIASES). It will be synced to the
+# `openshift/release/ci-operator` repo where non-Openshift org members will be
+# filtered out and aliases will be expanded.
+
+# See: https://coreos.slack.com/archives/CBN38N3MW/p1654872921259859
+
 # See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
+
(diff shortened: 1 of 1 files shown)


travier (approved):

travier:
CC @cgwalters

cgwalters:
I can't merge here, I think I have "write" but not "admin" access.  The former does allow merging over when all required contexts from GH's perspective are met (which is distinct from what Prow gates on).

cgwalters:
I think we need a GH organization administrator here; it's not gating on the tests, it's gating on approve and lgtm and we can't do that anymore.

OWNERS
```diff
@@ -0,0 +1,33 @@
+# This file is required in order to control `/approve` and `/lgtm` and related
+# Prow commands on this repo.  It may contain non-Openshift org members and
+# aliases (defined in OWNERS_ALIASES). It will be synced to the
+# `openshift/release/ci-operator` repo where non-Openshift org members will be
+# filtered out and aliases will be expanded.
+
+# See: https://coreos.slack.com/archives/CBN38N3MW/p1654872921259859
```
wking, line 7: Slack links will rot out.  If you have something relevant from this discussion, probably just inline it here.

cgwalters:
> /lgtm
Also need `/approve`; thanks for the assist!

miabbott:
We reached out to @mrunalp and friends to help us get the necessary labels and overrides to make Prow/Tide happy

Context - https://coreos.slack.com/archives/CBUT43E94/p1654879867017369

PR #845, opened 2022-06-10
ci: Follow up from release changes and cleanups

travier:
ci: Do not fail on rpm-ostree missing

The validate job does not need to run inside COSA and thus will not habe
rpm-ostree installed.

---

ci: Regroup tests in qemu & metal buckets

- Only build the ostree aci container archive in the build step as it's
  the only artifact used for the layering test.
- Group the remaining tests in two buckets: qemu & metal.
- Disable the now uneeded tests.

---

ci: Prepare and clean-up test names

- Use explicit test names for 8.6
- Pre-create test names to be able to enable them in Prow before we
  merge the corresponding test code.
- Clean-up test names

---

ci: Remove now unused scripts

Those scripts have been merged into prow-entrypoint.sh.

---

ci: Remove Dockerfiles (moved to release)

Keep Dockerfiles in openshift release to be able to test changes.

See: https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image


travier:
As we've merged all changes on the openshift/release side, we now need to fix CI here first.

travier:
Two points:
- Looks like the qemu test timed out at 2h
- The `test-in-cluster` job has still been started even though we've removed it for now

PR #851, opened 2022-06-17
ci: Add an explicitly named test entry for COSA CI

travier:
ci: Add an explicitly named test entry for COSA CI

Adding an explicit test entry point will enable us to focus on tests
that matter for COSA CI (such as buildextends for all platforms) and not
on general RHCOS testing.

---

ci: Remove now unused test names

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -170,7 +170,7 @@ main () {
             cosa_build
             kola_test_metal
             ;;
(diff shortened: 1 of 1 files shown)


travier:
See https://github.com/coreos/coreos-assembler/pull/2922

travier:
COSA change in https://github.com/coreos/coreos-assembler/pull/2930

PR #852, opened 2022-06-17
Split manifest and update CI

travier:
ci: Add an explicitly named test entry for COSA CI

Adding an explicit test entry point will enable us to focus on tests
that matter for COSA CI (such as buildextends for all platforms) and not
on general RHCOS testing.

---

ci: Remove now unused test names

---

manifests: Split into common & RHEL 8.6 specific manifests

---

Add script to select which version to build

---

ci: Expllcitly select the version to build


travier:
Based on https://github.com/openshift/os/pull/851 & https://github.com/openshift/os/pull/850

travier:
Initial version of this PR had no default manifest symlink but this does not work well with https://github.com/coreos/coreos-assembler/blob/main/src/cmd-init#L135

travier:
Hum, I should have added a hold here until the discussion had reached a consensus on this one vs the other alternative. Well, we can always redo the layout later if we decide to.

jlebon:
Hmm, yeah I'm really not a fan of requiring running a script to set up the config dir. It makes things less declarative and more opaque.

If we need to fix cosa to handle symlinks correctly for some files, then can we do that instead?

travier:
Agree that this is not ideal but I don't have a better idea right now. The last workaround was basically to copy the entire repo into the subfolder which is even less ideal: https://github.com/openshift/os/pull/849/files#diff-cf4ac67e6195e36ef10dd53e72af30a0346517d3faf3857f88e74204c7373e3bR53-R66

travier:
See also https://github.com/coreos/coreos-assembler/pull/2934

travier:
We could make `cosa init` run this script by default when found, thus removing this step. The version could be stored in a distinct file if needed.

travier:
And:
> I have another bug with cosa replacing the config -> config-git symlink to an absolute symlink when buiding locally.

jlebon:
I opened https://github.com/coreos/coreos-assembler/pull/2936 which *should* fix the symlinking issues that Micah was hitting in https://github.com/openshift/os/pull/803#issuecomment-1156896671.

I'm cool with moving away from subdirs in the future, but short-term I'd still prefer that over a setup script.

travier:
Let's do https://github.com/coreos/coreos-assembler/pull/2934#issuecomment-1160706578 then and decide on an interface to select the variant/version from a file?

travier:
Otherwise we would be changing the layout to use sub dirs to then change it again to remove the sub dir.

jlebon:
I'm OK with that if others prefer, but would be more comfortable just reverting this and fixing up the subdirs approach. Then we can take our time to design something better.

travier:
If we decide to move forward to using subdirs then I can make a PR to move the current state to that as the manifests are now ready and split. Reverting this one will have us re-review the split for nothing.

PR #853, opened 2022-06-17
ci: Add all platform image builds for COSA CI test

travier:
Make sure that we can buildextend for all supported platforms during
test jobs run for COSA CI.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -109,6 +109,34 @@ kola_test_metal() {
     fi
 }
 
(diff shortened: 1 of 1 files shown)


travier:
This currently takes ~2h20min in CI here wo we might want to skip running some tests (maybe the ext. ones?)

travier:
Note that this is not tested by the CI here

miabbott:
This looks like a good idea; @travier want to rebase it and we'll `/retest`?

cgwalters:
Followup in https://github.com/openshift/os/pull/889

PR #854, opened 2022-06-17
manifests & ci: Add RHCOS 9 & SCOS

travier:
Alternative to https://github.com/openshift/os/pull/773


HuijingHei:
Thanks for working on this!
I prefer this one, as it looks neater

cgwalters:
Or I guess that's what https://github.com/openshift/os/pull/856 is?  I'll look at that one

PR #855, opened 2022-06-21
Rework default variant selection

travier:
Goes with https://github.com/coreos/coreos-assembler/pull/2934


travier:
Replaced by https://github.com/openshift/os/pull/901

PR #856, opened 2022-06-21
Add RHEL 9.0 and SCOS variants

travier:
Rework default variant selection

---

manifest & ci: Add RHEL 9.0 variant

---

manifest & ci: Add SCOS variant


bgilbert:
It looks as though all variants are currently sharing an `image.yaml`?  There are a couple `image.yaml` values that will need to diverge between RHEL 8 and 9.

travier:
Thanks for the info, will make that happen as part of the COSA changes

travier:
Replaced by https://github.com/openshift/os/pull/901

PR #860, opened 2022-06-21
Bug 2099812: manifest: bump to openvswitch2.16

dcbw:
- We want to keep host OVS and container OVS the same versions; we're using
  2.16 in ovn-kubernetes containers
- 2.16 has larger netlink buffers for upcalls which in high-traffic/upcall
  situations keeps things flowing well
- 2.16 was part of the 4.11 dev cycle from mid-Feb -> late-April and
  did not show data plane regressions in CI or perf/scale testing

@trozet @fleitner

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -223,7 +223,7 @@ packages:
  - cri-o cri-tools
  # Networking
  - nfs-utils
- - openvswitch2.15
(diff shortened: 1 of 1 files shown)


trozet:
I don' think I have permission.. @miabbott can you help out here for backport risk label?

PR #871, opened 2022-06-30
(WIP)-Running toolbox with no arguments

RyanJenkins99:
This is the first of my test cases this will check that toolbox runs with no arguments


ravanelli (commented):
I would suggest you to create a new `tests/kola/toolbox` dir for the kola tests, since we will adding other tests to it. @miabbott added a toolbox test in version, because it was checking rpm version for it.

PR #880, opened 2022-07-06
manifest: Require `skopeo >= 2:1.7.0`

cgwalters:
This is needed for the latest rpm-ostree;
https://github.com/coreos/rpm-ostree/issues/3819


cgwalters:
Note this PR *should* fail until we've got the newer version from 8.7 shipped in the 4.12 OCP channel.

travier (approved):

PR #882, opened 2022-07-07
ci: Remove version selection script

travier:
Remove the current multi-version setup that was merged in https://github.com/openshift/os/pull/852 as we are moving to a COSA based setup.


travier:
I'm removing that now before the alternative is ready to avoid creating dependencies on it.

travier:
CC @jlebon

miabbott:
> Remove the current multi-version setup that was merged in #852 as we are moving to a COSA based setup.

This looks fine, but is there context/breadcrumbs on how we are going to select the version via `coreos-assembler`?

travier:
Yes, the plan for COSA is in https://github.com/coreos/coreos-assembler/pull/2934

PR #886, opened 2022-07-11
fixes in-cluster derivation test

cheesesashimi:
This relocates the Dockerfile, hello-world source, etc. used for the in-cluster derivation test into this repository so we do not have to rely upon https://github.com/coreos/coreos-layering-examples. The reasoning is because coreos-layering-examples does not currently have CI set up to validate that the Dockerfiles it contains can be built.


travier:
This LGTM but I don't have the knowledge here.
/hold
/lgtm
/assign @cgwalters @jmarrero

jmarrero (approved):
lgtm

PR #887, opened 2022-07-11
extensions/Dockerfile: remove build arg

cheesesashimi:
When ci-operator builds this Dockerfile, it attempts to replace
registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest with a pullspec
to the RHCOS image it just built. However, since this takes a build arg,
ci-operator cannot substitute the build arg for the rhel-coreos:latest
image. This means that the extensions container could potentially be
built against the wrong RHCOS version.


jmarrero (approved):
lgtm, since it's being replaced by CI.

PR #888, opened 2022-07-11
image.yaml: disable secure boot

mike-nguyen:
Secure boot enabled by default causes issues with unsigned kernel
modules.

ref: https://bugzilla.redhat.com/show_bug.cgi?id=2106055

Requires https://github.com/coreos/coreos-assembler/pull/2971

### DIFF
--- a/image.yaml
+++ b/image.yaml
@@ -10,6 +10,10 @@ ignition-network-kcmdline: []
 
 ostree-format: "oci"
 
+# vmware-secure-boot changes the EFI secure boot option.
+# set false here due to https://bugzilla.redhat.com/show_bug.cgi?id=2106055
+vmware-secure-boot: "false"
+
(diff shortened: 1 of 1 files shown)


cgwalters:
/lgtm
as is but (in case you do any followup) I think it'd be good to also put a doc comment above the code in `image.yaml` linking to the bug.

(`git annotate/blame` will get one there, but better to save the step for stuff like this)

image.yaml
```diff
@@ -10,6 +10,7 @@ ignition-network-kcmdline: []
 
 ostree-format: "oci"
 
+vmware-secure-boot: "false"
```
miabbott, line 13: Let's add a comment here with a little context/warning about changing this value

mike-nguyen:
Added comment with link to BZ

travier (approved):

PR #889, opened 2022-07-11
ci/prow-entrypoint: Ensure we have a qemu image for followup builds

cgwalters:
This should fix https://github.com/coreos/coreos-assembler/pull/2919#issuecomment-1180824612

Fixes: c6e7dd9f99cfc9fe90ec87e041943743ab150cd6


travier:
Note that nothing here tests this change thus why I did the hack in #890

travier:
/close
Closing in favor of https://github.com/openshift/os/pull/891 until we test things more.

cgwalters:
okay, but wouldn't you say this change is still obviously correct?

PR #890, opened 2022-07-12
CI: Re-enable buildextend testing for COSA CI

travier:
This should fix coreos/coreos-assembler#2919 (comment)

Fixes: c6e7dd9
Fixes: coreos/coreos-assembler#2973

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -204,6 +204,7 @@ main () {
             setup_user
             cosa_init
             cosa_build
+            cosa_buildextend_all
             kola_test_qemu
             ;;
         "rhcos-86-build-test-metal")


cgwalters:
> fatal: x86_64 is not supported for building dasd images

OK.  I think we really need to have something like a `pipeline.yaml` stored in the config git, which configures the images to build.  Something like this:

```
$ cat pipeline.yaml
images:
  all: true
  except:
    - digitalocean
    - exoscale
```

Then `cosa buildextend-all` would read that.  And specifically for the dasd issue above, cosa itself already knows what images build on what platforms; we shouldn't replicate that logic in all pipelines.

travier:
https://github.com/openshift/os/pull/891 will fix our immediate issue and then we can figure out the failures in this one without having other PRs blocked on it.

travier:
This one is still taking too long. We will probably need to setup another test job that focuses on buildextends tests.
```
Process did not finish before 3h0m0s timeout
```

PR #891, opened 2022-07-12
ci: Make COSA test job like QEMU

travier:
Temporary workaround until we figure out CI failures during buildextend
for all platform images.


travier:
Nothing here is testing this change.
\override ci/prow/rhcos-86-build-test-metal 
\override ci/prow/rhcos-86-build-test-qemu

cgwalters:
OK though I still think https://github.com/openshift/os/pull/890 is an obviously-correct change that would likely have also fixed this.
/approve
/lgtm

PR #903, opened 2022-07-20
`image.yaml` updates for RHEL 9

bgilbert:
Switch to separate copies of `image.yaml` for each RHEL major.  Enable zstd compression in the RHEL 9 squashfs, which saves 53 MiB in the ISO image and should improve performance.  Document that we can't switch the VMware OS type yet.


bgilbert:
Is `ci/prow/periodic-images` known to be broken?

travier:
> Is `ci/prow/periodic-images` known to be broken?

Yes. This CI also does not test RHCOS 9 yet.

PR #904, opened 2022-07-21
rhel9: Switch to pulling toolbox from RHEL and not OCP

HuijingHei:
Fix issue:
When running `cosa fetch` for rhcos based on rhel9.0, get `error:
No matches for 'toolbox' in repo 'rhel-9-server-ose'`

### DIFF
--- a/manifest-rhel-9.0.yaml
+++ b/manifest-rhel-9.0.yaml
@@ -154,8 +154,3 @@ repo-packages:
   - repo: rhel-9-appstream
     packages:
       - nss-altfiles
-  - repo: rhel-9-server-ose
-    packages:
-      # eventually, we want the one from the container-tools module, but we're
-      # not there yet
-      - toolbox


travier:
We control this package so we should push the package instead.

cgwalters:
It's worth noting here there are *two* distinct things called `toolbox`:

- https://github.com/coreos/toolbox
- https://github.com/containers/toolbox

For historical reasons, RHEL CoreOS ships the first.  Fedora CoreOS has switched to shipping containers/toolbox (because Fedora did in general in https://src.fedoraproject.org/rpms/toolbox/c/33bd39b0f97b72a788c9ee564ce2b0e2a8a3c23e?branch=rawhide )

RHEL 9 has always shipped containers/toolbox as far as I know.

I think because of some differences between the two, OCP support wants to keep the first one for now.  Personally, I feel we should be dropping this legacy in the context of RHEL9 and switch to containers/toolbox.

So...as best I can tell there is definitely a `toolbox` in the RHEL9 AppStream repos; can you dig a bit more into why it's not being found?

travier:
It's trying to fetch it from `rhel-9-server-ose` which is empty right now.

travier:
As much as I would also like to move to the container toolbx, I'm not sure our ecosystem is ready yet. But maybe it's a good time to find out too.

manifest-rhel-9.0.yaml
```diff
@@ -154,7 +154,8 @@ repo-packages:
   - repo: rhel-9-appstream
     packages:
       - nss-altfiles
-  - repo: rhel-9-server-ose
+  # Temporarily using rhel-8-server-ose until rhel-9-server-ose is available
+  - repo: rhel-8-server-ose
```
travier, line 158:
```suggestion
  - repo: rhel-9-appstream
```

cgwalters:
> As much as I would also like to move to the container toolbx, I'm not sure our ecosystem is ready yet.

The RHEL9 switch seems like an excellent time to do this by default.

HuijingHei:
> there is definitely a toolbox in the RHEL9 AppStream repos; can you dig a bit more into why it's not being found?

Here is link http://rhsm-pulp.corp.redhat.com/content/dist/rhel9/9/x86_64/appstream/os/Packages/t/toolbox-0.0.99.3-3.el9_0.x86_64.rpm
It seems that we specify fetch `toolbox` from `rhel-9-server-ose` refer to https://github.com/openshift/os/blob/master/manifest-rhel-9.0.yaml#L157. We can remove that config if not needed.

WDYT?

cgwalters:
Let's do it - also how about we retitle the PR and commit message to something that shows it isn't temporary; e.g.:

> rhel9: Switch to pulling toolbox from RHEL and not OCP

The body text could pull in most of
https://github.com/openshift/os/pull/904#issuecomment-1191266178
too

travier (commented):
LGTM. CI here is not testing that.

travier (approved):

HuijingHei:
Seems CI is not working now, any suggestions?
```
[2/4] STEP 8/8: RUN rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}.yaml
error: parsing extensions.yaml: couldn't find base package kernel
[3/4] STEP 1/8: FROM quay.io/centos/centos:stream8 AS builder
Trying to pull quay.io/centos/centos:stream8...
error: build error: error building at STEP "RUN rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}.yaml": error while running runtime: exit status 1
```

PR #905, opened 2022-07-21
ci: Enable more kola tests for SCOS

travier:
Includes temporary workarounds for variant selection for non default
variants. Will be removed once COSA support for variants is completed.

---

el9: Re-enable /tmp on tmpfs

See: https://gitlab.com/redhat/centos-stream/rpms/systemd/-/blob/c9s/0004-Avoid-tmp-being-mounted-as-tmpfs-without-the-user-s-.patch
See: https://gitlab.com/redhat/centos-stream/rpms/systemd/-/commit/9efbb443aa9f8de751926cd903f9c56806d85972#b9c1ef0bb10dd8f850e36759e92628f539183e73

---

el9: Remove dbus-broker workaround

---

tests: Use new toolbox test

---

c9s: Disable iscsid.socket

---

denylist: Use versioned kola denylists

---

scos: Skip (currently) broken kola tests

---

ci: Remove disabled test and add instructions

Fixes: https://github.com/openshift/os/pull/937

---

ci: Enable more kola tests for SCOS


cgwalters:
Looks like CI is failing on extensions...we could perhaps disable that?  Or dive into figuring out the virt module and rhel9

travier:
SCOS is now hitting kola test failures cc @HuijingHei 
RHCOS is blocked on rhel-9-fast-datapath repo which has the wrong URL apparently.

HuijingHei:
> SCOS is now hitting kola test failures

Thanks @travier for the report, will look more about the failure

HuijingHei:
For the scos pxe-install failed, create https://github.com/openshift/os/issues/909 to track

cgwalters:
Might make sense to:

- change the CI config to avoid testing the things that are broken
- merge this PR to avoid any further regressions, and enable shipping e.g. `registry.ci.openshift.org/rhcos-devel/rhel-coreos-9:latest`
- track things like kola PXE testing as separate issues

travier:
I'm looking for the correct URL for this repo but have not found it yet:
```
Updating metadata for 'rhel-9-fast-datapath'...done
[0m[31merror: [0mUpdating rpm-md repo 'rhel-9-fast-datapath': cannot update repo 'rhel-9-fast-datapath': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Status code: 404 for http://base-4-12-rhel90.ocp.svc.cluster.local/rhel-9-fast-datapath/repodata/repomd.xml (IP: 172.30.115.61)
```

To be changed in https://github.com/openshift/release/blob/master/core-services/release-controller/_repos/ocp-4.12-rhel90.repo

travier:
https://github.com/openshift/release/pull/30894 should unblock RHEL 9.0

cgwalters:
> rhel-9-fast-datapath

fast-datapath is as I understand things in a "netherworld" between OpenShift and OpenStack - both of which use openvswitch.  It is technically not RHEL.  As I understand it also, perhaps because OpenShift is leading OpenStack here in pushing for RHEL9, we're the first to hit its absence.

I tried a while ago to escalate this with their managers but didn't get any traction.  Now that it's closer to the blocking path, seems that we should try again.

OTOH...RHCOS only ships openvswitch on the host because the networking team pushed hard for it - but it certainly *can* be run in a container.  I would definitely like to revisit that discussion.  In that case, we can just ship the RHEL8 openvswitch in a container!

dcbw:
@cgwalters I'm not sure what the ask is here... we're very unlikely to revisit the decision to ship OVS on the host.

Is the problem that something isn't connecting the RHEL9 fast-datapath repos into the RHCOS structure? If that's a problem we can certainly fix that...

travier:
@cgwalters The openvswitch packages exists for RHEL 9. I had made a mistake in the repo URL which is fixed in https://github.com/openshift/release/pull/30894.

cgwalters:
> we're very unlikely to revisit the decision to ship OVS on the host.

:cry:

HuijingHei:
`rhcos-90-build-test-metal` failed `error: Packages not found: skopeo >= 2:1.7.0`
will be fixed by https://github.com/openshift/os/pull/921

travier:
SCOS QEMU test is now hitting failures in kola tests(!)
RHCOS is failing directly with:
```
=== RUN   basic
--- FAIL: basic (29.20s)
        harness.go:1454: mach.Start() failed: machine c937034e-03bd-41fe-8d6c-d1e1428c51b0 entered emergency.target in initramfs
```

HuijingHei:
Thanks @travier for moving this forward, a big step!

HuijingHei:
https://github.com/coreos/fedora-coreos-config/pull/1881 will fix `ext.config.shared.var-mount.scsi-id`

travier:
https://github.com/coreos/coreos-assembler/pull/3011 > This one should fix `rhcos-90-build-test-metal`

travier:
Rebased and merged https://github.com/openshift/os/pull/929 & https://github.com/openshift/os/pull/932 into this one for testing.

HuijingHei:
@travier could you help to pick https://github.com/openshift/os/pull/936 (when it is merged?) to trigger the testing? I just test locally `pxe-install` passed, still want to confirm, thanks!

travier:
Added changes from https://github.com/openshift/os/pull/936. Thanks!

ci/prow-entrypoint.sh
```diff
@@ -109,6 +147,17 @@ kola_test_metal() {
     fi
 }
 
+# Temporary to get EL9 in CI
+kola_test_qemu_light() {
```
cgwalters: cc https://github.com/coreos/coreos-assembler/issues/3008

cgwalters (approved):
/lgtm

travier:
/hold
This one is still for testing CI failures.
You'll want to LGTM this one: https://github.com/openshift/os/pull/937 🙂

travier:
There is something weird going on here as the basic emu scenarios are passing for me locally for RHCOS 9.

travier:
/unhold

Making this one just for kola testing of SCOS.

travier:
🎉

PR #907, opened 2022-07-21
docs: Add instructions to build & test SCOS

travier:
Initial set of docs for SCOS with current workarounds.

I'll also update the RHCOS docs to match.

### DIFF
--- a/docs/development-scos.md
+++ b/docs/development-scos.md
@@ -1,0 +1,83 @@
+# Building and developing CentOS Stream CoreOS
+
+## Background
+
+CentOS Stream CoreOS (SCOS) is a derivative of both CentOS Stream and Fedora
+CoreOS (FCOS). The tool to build both SCOS and FCOS is [coreos-assembler]. The
+process detailled here is thus very similar to the one described in [Building
+Fedora CoreOS].
+
+## Build process
+
+- Make sure you're meeting the [prerequisites].
+- Setup a `cosa` alias, following the [upstream documentation][cosa-alias].
+- Always make sure that you have the latest version of the COSA container:
+  ```
+  $ podman pull quay.io/coreos-assembler/coreos-assembler
+  ```
+- Create and use a dedicated directory:
+  ```
+  $ mkdir scos
+  $ cd scos
+  ```
+- Clone the config repo (`openshift/os`):
+  ```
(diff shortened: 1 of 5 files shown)


cgwalters (approved):
Looks sane to me!

I'm hopeful we can land the coreos-assembler bits to streamline this, which would help deduplicate the docs here.

docs/development-scos.md
```diff
@@ -0,0 +1,79 @@
...
+- Clone the config repo (`openshift/os`):
+  ```
+  $ cosa init https://github.com/openshift/os.git
+  ```
+- **Temporary workaround until we have variant support in COSA:** Manually
+  select the SCOS variant:
+  ```
+  $ ln -snf "manifest-c9s.yaml" "src/config/manifest.yaml"
+  $ ln -snf "extensions-c9s.yaml" "src/config/extensions.yaml"
+  ```
```
bgilbert, line 32: Also need to repoint the `image.yaml` symlink.

docs/development-scos.md
```diff
@@ -0,0 +1,79 @@
...
+- Fetch packages and build SCOS ostree container and QEMU image:
+  ```
+  $ cosa fetch
+  $ cosa build
+  ```
+
+## Building CentOS Stream CoreOS images for other platforms than QEMU
+
+- You can build images for platforms that are supported in COSA using the
+  [`builextend` commands][builextend]:
```
bgilbert, line 54:
```suggestion
  [`buildextend` commands][buildextend]:
```

docs/development-scos.md
```diff
@@ -0,0 +1,79 @@
...
+- You may then run an ephemeral virtual machine using QEMU with:
+  ```
+  $ cosa run
+  ```
+
+## Testing CentOS Stream CoreOS with kola
+
+- You may then run tests on the image built with [`kola`][kola]:
+  ```
+  $ cosa kola run
```
bgilbert, line 71: Without `--upgrades` or `--basic-qemu-scenarios`, `cosa kola` is just an alias for `kola` these days, right?  Maybe we should drop the `cosa` part here.

travier: We still need the `cosa` part for the users of the cosa container. I'll extend this part.

PR #912, opened 2022-07-25
manifest: Pin systemd version

ravanelli:
- The RHCOS CI is failing due an update in systemd,
let's pin it until we can solve it

More info: https://bugzilla.redhat.com/show_bug.cgi?id=2109546


saqibali-2k (approved):
/approve

ravanelli:
Closing it, https://bugzilla.redhat.com/show_bug.cgi?id=2109546 doesn't seem to be affecting 4.10 with it

Issue #908, opened 2022-07-21
Base image not being updated


cgwalters:
I'm trying to debug

```
$ oc image info registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest | grep Created
Created:     6d ago
$
```

:arrow_right: today's postsubmit https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-openshift-os-master-images/1549794749240578048 says Nothing to promote, skipping...  yet https://github.com/openshift/release/commit/f793a65f5ace82ad65a24bbd95b1d29ee76b63c5 was seemingly trying to re-enable this?

---

I tried to fix this in https://github.com/openshift/release/pull/30701

cgwalters:
cc @travier @cheesesashimi

travier:
I've triggered a retest on a PR to see if things have improved but we might have to wait until a PR is merged to see if the fix above works?

cgwalters:
Another attempt https://github.com/openshift/release/pull/30794

cgwalters:
It worked!

```
$ oc image info registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
Name:        registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest
Digest:      sha256:2ea468f4858bb834eb1ed2331315aa8e93bd1e2b6054718ebc019a15529cd8ae
Media Type:  application/vnd.docker.distribution.manifest.v2+json
Created:     1h ago
...
```

PR #914, opened 2022-07-26
Bump fedora-coreos-config

coreosbot-releng:

```
Benjamin Gilbert (2):
      05core: fix permissions of /etc/sudoers.d/coreos-sudo-group
      05core: fix coreos-boot-edit.service race with switch-root

Dusty Mabe (3):
      tests/default-network-behavior-change: update for NM 1.39.8+
      overrides: fast-track kexec-tools-2.0.24-4.fc36 for s390x kdump fix
      tests: bump memory for tests because of DNF issue

LakshmiRavichandran1 (1):
      modify ext.config.files.setgid test as per upstream change to liblockfile ; fcos.filesystem test is deleted by #2986

Luca BRUNO (1):
      manifests/group: fix ID for "input" group
```


travier:
/retest
We don't need it right now but it's a good test for the fix from https://github.com/openshift/os/issues/908

PR #929, opened 2022-08-03
common.yaml: add `crypto-policies-scripts` which provides `fips-mode-setup`

HuijingHei:
`crypto-policies-scripts` includes `fips-mode-setup` tool, which
is needed by rhcos-fips.sh. The package is missing when building
SCOS image, we need to add it


cgwalters:
(I do think at some point we should consider trying to "mechanically" support FIPS for RHCOS too; see discussion about upstreaming the current rhcos bits in https://github.com/coreos/ignition/issues/1323 )

PR #933, opened 2022-08-03
Bug 2117353: [release-4.11] bump fedora-coreos-config submodule

mike-nguyen:
git shortlog --no-merges f99e2b78224c3c4fbc64da9a03e27665d4487f09..539f342b78478360c1939c2547a343107bb79849
Benjamin Gilbert (1):
      05core: fix coreos-boot-edit.service race with switch-root

Dusty Mabe (1):
      tests: bump memory for tests because of DNF issue

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -5,8 +5,6 @@
   tracker: https://bugzilla.redhat.com/show_bug.cgi?id=1782026
   arches:
    - s390x
-- pattern: ext.config.chrony.dhcp-propagation
-  tracker: https://github.com/coreos/fedora-coreos-config/commit/3dee27d4487ee0a484858fe60e3c5b1a792d48b9
 - pattern: coreos.ignition.journald-log
   tracker: https://github.com/coreos/coreos-assembler/issues/1173
 - pattern: coreos.ignition.mount.*


cgwalters:
Hmmm.  Hmm.  The thing is, if we were *only* adjusting the test, this would be a 100% clearly safe PR to merge.

Now, we rolled in a change that actually does touch on the booted OS content.  It's a clearly important bugfix.  It's *probably* safe.  But we're deep in the 4.11 release cycle.  I don't think we should try to "sneak in" or combine changes like this.

IOW I'd say please split out that dracut change into a distinct PR that can be evaluated and tested independently.

mike-nguyen:
> Hmmm. Hmm. The thing is, if we were _only_ adjusting the test, this would be a 100% clearly safe PR to merge.
> 
> Now, we rolled in a change that actually does touch on the booted OS content. It's a clearly important bugfix. It's _probably_ safe. But we're deep in the 4.11 release cycle. I don't think we should try to "sneak in" or combine changes like this.
> 
> IOW I'd say please split out that dracut change into a distinct PR that can be evaluated and tested independently.

Makes sense.

> Hmmm. Hmm. The thing is, if we were _only_ adjusting the test, this would be a 100% clearly safe PR to merge.
> 
> Now, we rolled in a change that actually does touch on the booted OS content. It's a clearly important bugfix. It's _probably_ safe. But we're deep in the 4.11 release cycle. I don't think we should try to "sneak in" or combine changes like this.
> 
> IOW I'd say please split out that dracut change into a distinct PR that can be evaluated and tested independently.

@cgwalters Makes sense to me.  What is the best way to address this?  Is it better just to bump the submodule to the commit only the test fix or revert and bump the submodule?  I've done the revert here https://github.com/coreos/fedora-coreos-config/pull/1891 is the PR for the revert.

cgwalters:
Unfortunately right now the way we use git submodules makes it really hard to also have the PR workflow work.

I think the way we'd have to do it is:

- Revert again the dracut change on rhcos-4.11 branch of fedora-coreos-config
- Redo this PR
- Merge this PR
- Revert *back* the dracut change (i.e. now it's back)
- Do a new PR

Alternatively and perhaps simpler:

- Push a rhcos-4.11-2 git branch of fedora-coreos-config
- Switch the submodule to point at that

IMO, the real long term solution is probably to stop using git submodules and switch to [git subtree](https://www.atlassian.com/git/tutorials/git-subtree).

cgwalters:
Alternatively, get a staff engineer to approve the dracut change too.
(Either way this needs staff-eng-approved, but clearly we can just shortcut that for test-only changes)

cgwalters:
Or of course, just denylist the test...

mike-nguyen:
Let's try reverting the dracut change and redo this PR.  If the tests pass then we are good.  If the tests fail, we can do the revert revert and bump the submodule again.

cgwalters:
https://github.com/openshift/os/pull/934

cgwalters:
OK though after https://github.com/openshift/os/pull/934 merges, I think we should keep this PR open because we definitely do want that `05core: fix coreos-boot-edit.service race with switch-root` bugfix for 4.11.1 or so.

mike-nguyen:
for 4.11.1 lets pick up the dhcp.propagation test memory bump and the race condition fix so we can get clean test runs.  We'll need revert https://github.com/openshift/os/pull/934 after this lands

travier:
Any reason not to add https://github.com/openshift/os/pull/934 directly in this PR to make sure that the test pass?

mike-nguyen:
> Any reason not to add #934 directly in this PR to make sure that the test pass?

Great point!  Added a revert to #934 to this PR .  

Side note: This test was still running with #934.  I just checked and it is because #934 denylisted `ext.config.chrony.dhcp-propagation` but when run in the RHCOS pipeline it is `ext.config.shared.dhcp-propagation`. We somehow got a build that passed for 4.11. :man_shrugging:

mike-nguyen:
@travier can we get the backport-risk-assessed and cherry-pick-approved labels on this ?

travier:
/label cherry-pick-approved
/label backport-risk-assessed
You should also have the power to set them :)

mike-nguyen:
> /label cherry-pick-approved /label backport-risk-assessed You should also have the power to set them :)

I wasn't sure if that is okay to do to my own PR

PR #934, opened 2022-08-03
Bug 2114964: denylist: Skip ext.config.chrony.dhcp-propagation

cgwalters:
The test needs adjustment, the OS content is fine.  Fix this
after 4.11 is out.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -5,6 +5,8 @@
   tracker: https://bugzilla.redhat.com/show_bug.cgi?id=1782026
   arches:
    - s390x
+- pattern: ext.config.chrony.dhcp-propagation
+  tracker: https://github.com/coreos/fedora-coreos-config/commit/3dee27d4487ee0a484858fe60e3c5b1a792d48b9
 - pattern: coreos.ignition.journald-log
   tracker: https://github.com/coreos/coreos-assembler/issues/1173
 - pattern: coreos.ignition.mount.*


mike-nguyen:
This test also builds a container with fedora 36 [tests/kola/podman/rootless-systemd](https://github.com/coreos/fedora-coreos-config/pull/1889/files#diff-70da451cffdbda28ba5478430f89ab1505a71c2160b86fb992cdc82f979a52cd)

cgwalters:
OK seriously let's just override all the paperwork for merging this - this is a corner case unimportant test that we know passes - *no product code changes are involved* - there's no need to do the "have a 4.12 BZ" dance etc etc.

cgwalters:
Ah, but that test appears to be fcos only.  So I think we can ignore that.

unrelated regular soapbox:

```
#   - The test is scoped to only FCOS because the `--retry-all-errors`
#     flag passed to `curl` is not avaiable on RHEL8.
```

But see, RHEL9 exists now...so again this "distribution matching" is wrong.

cgwalters:
For completeness we are already shipping the "real" fix for this in 4.12 since https://github.com/openshift/os/pull/914 merged.

cgwalters:
(CI is failing since build02 is apparently having problems right now, and our CI requires nested virt which is only setup on build02)

sdodson:
/retitle Bug 2114964: denylist: Skip ext.config.chrony.dhcp-propagation

cgwalters:
Huh actually...it looks like the build succeeded in https://url.corp.redhat.com/3d15ac9 (x86_64).  But still though it's likely to potentially flake on other architectures unless we merge this.

PR #936, opened 2022-08-04
scos: fix version check and enable /tmp on tmpfs

HuijingHei:
- `/etc/os-release` in scos does not include `RHEL_VERSION`, some version check cases failed based on that variable, this is to add workaround for scos
- enable /tmp on tmpfs (disabled by https://gitlab.com/redhat/centos-stream/rpms/systemd/-/blob/c9s/0004-Avoid-tmp-being-mounted-as-tmpfs-without-the-user-s-.patch)


travier:
Let's adapt the test to check C9S versions too. We can probably a switch case on `ID` to figure out if we're running scos or rhcos.

HuijingHei:
> Let's adapt the tests to check C9S versions too. We can probably a switch case on `ID` to figure out if we're running scos or rhcos.

 Good suggestion, updated!

travier:
Thanks! I'm adding them in https://github.com/openshift/os/pull/905 to get them tested.
/close

PR #937, opened 2022-08-04
CI: Partially enable RHCOS 9 & SCOS testing

travier:
Let's enable some CI that works right now so that at least we don't regress.

---

manifests: Add shared manifest for EL9 variants

---

repos: Update C9S SIG NFV repo name

---

ci: Partially enable RHCOS 9 & SCOS testing

- Only build RHCOS9, only run basic tests for SCOS
- Includes temporary workaround for variant selection for non default
  variants. Will be removed once COSA support for variants is completed.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -48,6 +48,19 @@ cosa_init() {
     cosa init --transient "${tmp_src}/os"
 }
 
+# Temporary hack while we setup variant support in COSA
+setup_variant() {
+    if [[ ${#} -ne 1 ]]; then
+        echo "This should have been called with a single 'variant' argument"
+        exit 1
+    fi
+    local -r variant="${1}"
+    echo "Selecting variant: ${variant}"
+    ln -snf "manifest-${variant}.yaml" "src/config/manifest.yaml"
+    ln -snf "extensions-${variant}.yaml" "src/config/extensions.yaml"
+    ln -snf "image-${variant}.yaml" "src/config/image.yaml"
+}
+
 # Do a cosa build & cosa build-extensions only.
 # This is called both as part of the build phase and test phase in Prow thus we
 # can not do any kola testing in this function.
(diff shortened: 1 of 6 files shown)


extensions-rhel-9.0.yaml
```diff
@@ -4,6 +4,8 @@
 
 repos:
   - rhel-9-nfv
+  # Temporarily included for kata-containers
+  - rhel-8-server-ose
```
jlebon, line 8: On this topic, one thing we should probably support in rpm-ostree is to have repos mentioned in `repo-packages` be enabled but used only for the mentioned packages. This would make things like this cleaner, so that we don't accidentally also ship other things from that repo that we didn't intend because of EVR ordering funkiness.

ci/prow-entrypoint.sh
```diff
@@ -203,7 +252,45 @@ main () {
...
+            setup_user
+            cosa_init
+            # Temporary. Will be removed once variant support is in COSA
+            setup_variant "c9s"
+            cosa_build
+            # Temporary light tests only until all tests pass
+            kola_test_metal_light
+            # kola_test_metal
+            ;;
+        "explicitely-disabled-test")
```
jlebon, line 293: What's this about?

jlebon (approved):
LGTM! I didn't review the symlink handling bits in depth (since we're cleaning that up soon anyway).

jlebon:
Aside: re. the Prow tests, any reason we don't combine the -qemu and -metal ones into a single test?

extensions-rhel-9.0.yaml
travier: Agree. Right now I'm doing that manually in the repo configs: https://github.com/openshift/os/pull/937/files#diff-cf4ac67e6195e36ef10dd53e72af30a0346517d3faf3857f88e74204c7373e3bR94

ci/prow-entrypoint.sh
travier: It's a way to have an entry-point for Prow that works but does nothing so that we can add new tests in the Prow config without having to do changes in this repo first.

travier: 1. Add a new test in openshift/release that calls this name
2. Make a PR that changes what this does
3. When ready, update openshift/release to get a new name
4. update your PR for the new name and merge

Without that, you have to hijack an existing test to test a new one without breaking CI for other PRs.

travier:
> Aside: re. the Prow tests, any reason we don't combine the -qemu and -metal ones into a single test? E.g. have only `rhcos-86-build-test`.

That's what we had before but this meant that we had a single test that lasted 3-4h. The two tests right now take 1h30 (qemu) and 2h30 (metal).

I'll make a test PR to see how things look like now if we merge the tests.

ci/prow-entrypoint.sh
jlebon: Gotcha, makes sense. Might help to add a comment there explaining that. Then we can link to it too for anyone working on adding new tests.

travier: Thinking about this more, I've simplified it in [`a5d371c` (#905)](https://github.com/openshift/os/pull/905/commits/a5d371cf78d2923a51ef9d48597c6d91cc1099cd)

PR #939, opened 2022-08-04
ci: Merge qemu & metal tests

travier:
Test PR to see how long it takes to run both qemu & metal tests in a single job (would save us a build of RHCOS).
/hold

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -196,12 +196,11 @@ main () {
             cosa_init
             cosa_build
             kola_test_qemu
+            kola_test_metal
             ;;
         "rhcos-86-build-test-metal")
-            setup_user
-            cosa_init
-            cosa_build
-            kola_test_metal
+            echo "Disabled"
+            exit 0
             ;;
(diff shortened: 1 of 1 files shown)


travier:
@jlebon: This one failed as it lasted more than 3h

travier:
We could increase the timeout, but I don't think it's a great experience when tests last more than 3h.

cgwalters:
Yeah to really make this sustainable we need to figure out how to get Prow jobs to actually share the ostree-container build.

Worst case I think we could do something hacky like push the image to a (separate) registry, and then have the other jobs poll/wait for it.

Once we have that, we can "shard" tests as much as we want.  For example,

- add a --shard=N option to `kola`
- Split up the kola tests into 2 shards i.e. `rhcos-86-test-1` and ``rhcos-86-test-2` prow contexts

Also, I still think we should get away from running all tests on all builds.  We also shouldn't *only* run the tests on qemu, now that we're in Prow we've always been able to use the Prow cloud credentials to e.g. run tests on AWS/GCP/etc.  Splitting off that alone, particularly to use the new `kola run --oscontainer` flow https://github.com/coreos/coreos-assembler/pull/3004 would be an entirely different way to shard and speed up tests.

PR #956, opened 2022-08-16
extensions: remove rpm-ostree build from Dockerfile

jmarrero:
The latest RHCOS image has a newer rpm-ostree that allows us to compose the extensions, no need to build a new version of rpm-ostree going forward.


jmarrero:
tested locally, working as expected.

PR #958, opened 2022-08-25
Updates for variant & yumrepo support in COSA

travier:
manifests: Rework symlinks to match variant support in COSA

To make the logic simpler, we're carrying the variant <-> version logic
with symlinks in this repo

See COSA support in https://github.com/coreos/coreos-assembler/pull/2934

---

manifests: Fix 'distro' variable for C9S

---

kola-denylist & ci: Use COSA variant support

---

docs: Update for variant & yumrepo support


HuijingHei:
I am OK with this, just confused about the `manifest.yaml -> manifest-rhel-coreos-8.yaml -> manifest-rhel-8.6.yaml`, is there any concern to do this? Why not ln `manifest.yaml -> manifest-rhel-8.6.yaml` directly?

travier:
Agree that this is confusing. I'll update the docs and add a note like below to explain.

The idea is that we want to have a default variant (no suffix) that is the one we want to build by default when you clone the repo. This is also the one shipped in production.

We also want explicit options to be able to work on a specific variant by specifying exact RHEL or CentOS Stream versions as variants (`rhel-8.6`, `rhel-9.0`, `c9s`).

But we also want to be able to track major streams/versions as we will make two pipelines, one for RHEL 8 based RHCOS and one for RHEL 9 based RHCOS (plus one for SCOS elsewhere). So we need `rhel-coreos-8`, `rhel-coreos-9` & `scos`. We could even have `scos-9` but we don't really need it right now.

Thus the idea behind the chain of symlinks is that we make explicit those decisions:
- default points to `rhel-coreos-8` as that's what we ship, no matter the version
- `rhel-coreos-8` & `rhel-coreos-9` point to the current minor version that we want to ship
- when we will switch the default to`rhel-coreos-9` and when we will update the version pointed to by `rhel-coreos-9`, we will get the default also updated "for free".

ci/prow-entrypoint.sh
```diff
@@ -233,53 +227,45 @@ main () {
             ;;
         "rhcos-cosa-prow-pr-ci")
             setup_user
-            cosa_init
+            cosa_init "rhel-coreos-8"
```
jlebon, line 230: Bikeshed: WDYT about `rhcos8` and `rhcos9` instead? Seems closer to what we're used to type and matches the other `scos` better too.

jlebon (approved):
LGTM! Thanks a lot for updating the docs.

ci/prow-entrypoint.sh
cgwalters: See https://github.com/coreos/coreos-assembler/pull/2934#discussion_r952978285

cgwalters: I agree typing `rhcos` is more convenient in quick emails and interactive chat.  But for anything a bit more formal I'd like to move to being more explicit because the world can only have so many acronyms.  "rhel-coreos" is not *that* much longer and is much, much more self-describing, even though it still contains one acronym.

travier: We've started using "rhel-coreos-8" & "rhel-coreos-9" for the machine-os-content & container names somewhere I don't remember exactly thus the change here. CC @cgwalters

travier: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/os/openshift-os-master__periodic.yaml#L101

cgwalters: I think my core argument here is: With the introduction of RHEL9, the abbreviation "rhcos" has become ambiguous outside of the most generic contexts.  Specifically, mirror.openshift.com for example can't just use `rhcos` - it has to become either `rhcos8` and `rhcos9` as you suggest, but I think while we are renaming things, let's use `rhel-coreos-8` and `rhel-coreos-9` there and in our disk images in general.

jlebon: No strong argument against `rhel-coreos-N`, but note that RHCOS as an acronym is pretty established at this point, including in official OpenShift docs. We do need to start adding the version string, but I'm not sure it follows that we need to also change the prefix.

jlebon:
I think we need to force a rebuild of the layered cosa image.

/retest all

jlebon:
CI failure looks legit now.

cgwalters:
The failing tests look like flakes to me offhand.
/retest

PR #959, opened 2022-08-25
extensions-el9: kata-containers is now available

travier:


jlebon:
OK, `build-image` was rebuilt. Now let's try it.
/retest

jlebon:
OK right. The cosa imagestream is updated in this cluster: https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/coreos/imagestreamtags/coreos-assembler%3Alatest, but looking at https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_os/959/pull-ci-openshift-os-master-images/1563211516721762304/build-log.txt, it appears to run on build03, and there indeed it's not updated: https://console.build02.ci.openshift.org/k8s/ns/coreos/imagestreamtags/coreos-assembler%3Alatest. And I'm not quite sure why.

Something must be updating it because the image there isn't that old. It should be mirrored from Quay.io via https://github.com/openshift/release/pull/31015. I'm not sure how often this mirroring takes place.

jlebon:
Wait, those images have the `io.openshift.build.*` keys, which tell me they were built by OpenShift. But we deleted all the BuildConfigs in https://github.com/openshift/release/pull/31015.

Ahhh, looks like we also need to delete the in-cluster objects: https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/coreos/buildconfigs. But now I have a doubt whether the reverse mirroring we put in place has actually been working or whether it just appeared that way because we were still building cosa. /cc @dustymabe 

But still, that doesn't answer the question here. The last cosa build from there is https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/coreos/imagestreamtags/coreos-assembler%3Alatest, and it does appear up to date. Maybe there's a delay between propagating to the other clusters? I'm not sure how these objects are shared between the different clusters TBH.

cgwalters:
> Ahhh, looks like we also need to delete the in-cluster objects: https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/coreos/buildconfigs.

Hmm I didn't think that manual intervention was necessary with the prow gitops flow...one problem here is we probably need to start doing regular reprovisioning for the CI clusters to flush out unmanaged state.

> Maybe there's a delay between propagating to the other clusters? 

There almost certainly is.  The job might be https://github.com/openshift/release/blob/master/ci-operator/jobs/infra-image-mirroring.yaml which looks like it's `@hourly`?

travier:
With `flatpak run --user org.kde.konsole --noclose`:
```
/bin/bash: /bin/bash: cannot execute binary file
```

PR #962, opened 2022-08-29
OCPBUGS-595: overlay: Add `rhcos-selinux-policy-upgrade.service`

cgwalters:
This will auto rebuild the policy if we detect local modifications;
it is only needed before RHEL 8.7 when we'll pick up the
combined underlying work from
https://bugzilla.redhat.com/show_bug.cgi?id=2057497

### DIFF
--- a/overlay.d/05rhcos/usr/libexec/rhcos-rebuild-selinux-policy
+++ b/overlay.d/05rhcos/usr/libexec/rhcos-rebuild-selinux-policy
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Executed by rhcos-selinux-policy-upgrade.service
+set -euo pipefail
+
+RHEL_VERSION=$(. /usr/lib/os-release && echo ${RHEL_VERSION:-})
+echo -n "RHEL_VERSION=${RHEL_VERSION:-}"
+case "${RHEL_VERSION:-}" in
+  8.[0-6]) echo "Checking for policy recompilation";;
+  *) echo "Assuming we have new enough ostree"; exit 0;;
+esac
+
+ls -al /{usr/,}etc/selinux/targeted/policy/policy.31
+if ! cmp --quiet /{usr/,}etc/selinux/targeted/policy/policy.31; then
+    echo "Recompiling policy due to local modifications as workaround for https://bugzilla.redhat.com/2057497"
+    semodule -B
+fi
--- a/overlay.d/05rhcos/usr/lib/systemd/system-preset/43-manifest-rhcos.preset
+++ b/overlay.d/05rhcos/usr/lib/systemd/system-preset/43-manifest-rhcos.preset
@@ -3,6 +3,7 @@
 
 # Upgrade fixes
 enable rhcos-usrlocal-selinux-fixup.service
+enable rhcos-selinux-policy-upgrade.service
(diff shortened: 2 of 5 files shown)


travier (approved):

cgwalters:
OK lifting draft, now tested manually, and I added an automated kola test.

overlay.d/05rhcos/usr/libexec/rhcos-rebuild-selinux-policy
```diff
@@ -0,0 +1,15 @@
...
+set -euo pipefail
+
+RHEL_VERSION=$(. /usr/lib/os-release && echo $RHEL_VERSION)
+echo -n "RHEL_VERSION=${RHEL_VERSION}"
+case "${RHEL_VERSION}" in
+  8.[0-6]) echo "Checking for policy recompilation";;
+  *) echo "Assuming we have new enough ostree"; exit 0;;
+esac
+
+ls -al /{usr/,}etc/selinux/targeted/policy/policy.31
```
travier, line 11: I'm not 100% sure this is stable across all of our releases.

travier:
OK, I've verified that it is `.31` on 4.8 so that's good enough.
/lgtm

overlay.d/05rhcos/usr/libexec/rhcos-rebuild-selinux-policy
cgwalters: Just above this, we exit if the OS isn't RHEL 8.[0-6].  This code very intentionally will not run on RHEL9, because we don't need it there.

cgwalters:
Oh actually
```
Aug 29 20:05:35.151475 rhcos-rebuild-selinux-policy[1256]: /usr/libexec/rhcos-rebuild-selinux-policy: line 4: RHEL_VERSION: unbound variable
```

jmarrero (approved):
lgtm

jmarrero (approved):
lgtm

cgwalters:
@jmarrero  Editing a comment doesn't seem to cause prow to re-evaluate the state, I think you need to do a new comment.

jmarrero:
> @jmarrero  Editing a comment doesn't seem to cause prow to re-evaluate the state, I think you need to do a new comment.

I do see a message that says:

'''This pull-request has been approved by: cgwalters, jmarrero, travier'''

But I will add it again.

cgwalters:
Hmm I swear I pushed a fix for the case of c9s yesterday, but not seeing it here.  Anyways I just changed it to use `${RHEL_VERSION:-}` consistently so we don't error out if it's unset.

cgwalters:
Ahh I *did* push that attempted change but to the test code, not the core logic :facepalm: - they're both similar looking shell scripts.  Fixed.

travier:
```
Wrote commit: 193b3ec7eef58756c1ba2577e236bd7deb07ba034a1613fb74f41f642a49d031
mv: cannot move 'cache/output.tar' to '/tmp/cosa/tmp/repo/commit.tar': Cannot allocate memory
```

We might need to request more memory for this pod?

travier (approved):
/lgtm

travier:
Failed at the 3h mark.

PR #967, opened 2022-08-31
ci/prow-entrypoint: Support explicit `init` verb

cgwalters:
I want to add a build of `rhel-coreos-9` to our periodics,
and while in theory we could also just do that via the full test suite
I'd like to avoid that for now.  This helps support a direct
build without tests.

IOW the prow periodic for rhel-coreos-9 can just do
```
src/ci/prow-entrypoint.sh init rhel-coreos-9
cosa build ostree
```

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -56,12 +56,8 @@ cosa_init() {
     cosa init --transient --variant "${variant}" "${tmp_src}/os"
 }
 
-# Do a cosa build & cosa build-extensions only.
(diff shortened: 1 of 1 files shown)


travier (commented):
/lgtm

travier:
I'll soon revisit building once in parallel all variants and then doing the tests from those builds.

cgwalters:
>  FAIL: ext.config.shared.root-reprovision.filesystem-only (635.20s)

This looks like an unrelated but potentially real problem - we somehow got hung up in the initramfs copying the rootfs.

PR #969, opened 2022-08-31
docs: Update build & test guides

travier:
The objective here is to make this guide canonical and "remove" the internal doc.


ashcrow:
>  Cannot allocate memory

Given this is a docs change it should be OK to skip.

/override periodic-images

PR #970, opened 2022-08-31
c9s: Move repo to the Git repo root

travier:
rpm-ostree will only use the repos it has defined in the manifest so it
is safe to always have to CentOS Stream repos available and makes things
easier for SCOS.


cgwalters:
/approve
/lgtm

But that said this PR inspired me to write https://github.com/coreos/fedora-coreos-config/pull/1947

Issue #954, opened 2022-08-16
scos: testiso with uefi get grub terminal


HuijingHei:
Build scos locally and run `kola testiso -S --qemu-firmware uefi --scenarios iso-live-login`, get grub terminal.

```
BdsDxe: loading Boot0001 "UEFI QEMU DVD-ROM QM00011 " from PciRoot(0x0)/Pci(0x1F,0x2)/Sata(0x5,0xFFFF,0x0)
BdsDxe: starting Boot0001 "UEFI QEMU DVD-ROM QM00011 " from PciRoot(0x0)/Pci(0x1F,0x2)/Sata(0x5,0xFFFF,0x0)
      Minimal BASH-like line editing is supported. For the first word,    
      TAB lists possible command completions. Anywhere else TAB lists     
      possible device or file completions.                                


grub>
```
A little like https://github.com/openshift/os/issues/953, but they have different error logs

HuijingHei:
Try to build live iso with latest grub2 (download and put under overrides/rpm/), but failed with 

```
INFO - Running command: ['/usr/bin/ostree', 'checkout', '--repo', '/srv/data/c9s-metal/tmp/repo', '--user-mode', '--subpath', '/usr/lib/bootupd/updates/EFI', '0e7fe829d8a8fa139891de0051c28f652a9a16eab87ebb3aec826c6533381e46', '/srv/data/c9s-metal/tmp/buildpost-live/efi']
Traceback (most recent call last):
  File "/usr/lib/coreos-assembler/cmd-buildextend-live", line 704, in <module>
    generate_iso()
  File "/usr/lib/coreos-assembler/cmd-buildextend-live", line 525, in generate_iso
    raise Exception(f"did not find exactly one EFI vendor ID: {vendor_ids}")
Exception: did not find exactly one EFI vendor ID: ['centos', 'redhat']
error: failed to execute cmd-buildextend-live: exit status 1
```

HuijingHei:
Try more about the uefi, run the following commands, get result is `PASS: iso-live-login (uefi + metal) (1m46.346s)`.
```
grub> ls
(proc) (cd0) (cd0,msdos2)
grub> configfile (cd0)/EFI/redhat/grub.cfg
```
Is there any suggestion about how to fix this? Thanks!

cgwalters:
[...]

PR #975, opened 2022-09-02
ci: Enable all metal kola tests for SCOS

travier:
This should work now that we have https://github.com/openshift/os/issues/954


HuijingHei (approved):
LGTM

PR #976, opened 2022-09-03
Bump fedora-coreos-config

coreosbot-releng:

```
Huijing Hei (2):
      dhcp-propagation: add workaround for scos and rhcos9
      rootless-systemd: make it work on rhcos
```


travier:
Added to https://github.com/openshift/os/pull/978

PR #972, opened 2022-09-02
Rename efi path and disable `podman.rootless-systemd` on c9s/rhel-9.0

HuijingHei:
- Rename EFI/redhat dir to EFI/vendor
    - The change in coreos-assembler is https://github.com/coreos/coreos-assembler/pull/3048
- denylist: disable `podman.rootless-systemd` on c9s/rhel-9.0
    - Enable `chrony.dhcp-propagation`, according to https://github.com/coreos/fedora-coreos-config/pull/1952

PR #978, opened 2022-09-05
Bump fedora-coreos-config, EFI path fix & denylist update

travier:
Bump fedora-coreos-config
```
Huijing Hei (2):
      dhcp-propagation: add workaround for scos and rhcos9
      rootless-systemd: make it work on rhcos
```
---

Rename EFI/redhat dir to EFI/vendor

Consider of building scos using centos stream, we want to make this
more flexiable to different ventor instead of fixed `redhat`. Change
with coreos-assember refer to https://github.com/coreos/coreos-assembler/pull/3048

---

denylist: Re-enable ext.config.shared.chrony.dhcp-propagation for EL9

---

denylist: Skip ext.config.shared.podman.rootless-systemd for EL9

---

denylist: Re-enable podman.dns for EL9


travier:
Merges https://github.com/openshift/os/pull/972 & https://github.com/openshift/os/pull/976

HuijingHei (approved):
/lgtm

PR #983, opened 2022-09-06
denylist: re-enable `bond-with-dhcp` and `bond-with-restart` for EL9

HuijingHei:
The change with scripts refer to https://github.com/coreos/coreos-assembler/pull/3079

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -57,11 +57,3 @@
   osversion:
     - c9s
     - rhel-9.0
-- pattern: rhcos.network.bond-with-dhcp
-  osversion:
-    - c9s
-    - rhel-9.0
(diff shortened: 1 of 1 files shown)


travier:
For the dnsmasq issue, should we update the tests instead of changing the default configuration or do we really want to update the default configuration?

HuijingHei:
> For the dnsmasq issue, should we update the tests instead of changing the default configuration or do we really want to update the default configuration?

I will dig more about this, thanks for your suggestion.

HuijingHei:
Change script https://github.com/coreos/coreos-assembler/pull/3079

travier:
OK, it's an RHCOS only test thus why is wasn't failing previously on FCOS.

Issue #950, opened 2022-08-11
Modify /etc/os-release to display Jira instead of Bugzilla


mike-nguyen:
Once the transition to Jira is completed, change the Bugzilla references to Jira.

Example on 4.11:
https://github.com/openshift/os/blob/fa6a502f35eeda25e8a95d0a4d7eabd39836c580/manifest.yaml#L149-L151

The new bug tracker will be:  https://issues.redhat.com/projects/OCPBUGS

travier:
We'll have to wait until the RHCOS component/board is created in this project or until we've made a short link that auto-fills the Jira issues with relevant informations.

PR #984, opened 2022-09-06
Switch to https://access.redhat.com/labs/rhir for bug reporting

cgwalters:
Closes: https://github.com/openshift/os/issues/950

OCP has migrated away from Bugzilla, so update the links
there.

I explicitly chose *not* to remove the
```
      REDHAT_BUGZILLA_PRODUCT="OpenShift Container Platform"
      REDHAT_BUGZILLA_PRODUCT_VERSION="${OCP_RELEASE}"
```

metadata because it's absolutely still valid to look there for
old bugs, and I am not aware of corresponding metadata keys for
JIRA - we could invent them of course, but it doesn't seem important
right now.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -34,7 +34,7 @@ Similarly, Fedora CoreOS is an "edition" of Fedora.
 
 ## Q: Where should I report issues with OpenShift Container Platform or Red Hat CoreOS?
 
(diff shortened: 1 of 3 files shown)


cgwalters:
(Actually a motivation I have for doing this right now is to force a rebuild across all architectures to ensure we pick up the new extensions container)

jmarrero (approved):
/lgtm

travier (approved):

PR #985, opened 2022-09-06
docs: Add rhaos-overrides.yaml

cgwalters:
I think the source of truth for what we ship should be defined
via git and changes that are tested via pull requests; not humans
manually invoking a Koji tag command.

We don't yet live in that world and may never for RHEL-8, though
we have the opportunity with C9S to do better.

For now, this file defines what we intend to ship in a public
human (and in theory machine) readable way.

### DIFF
--- a/docs/rhaos-overrides.yaml
+++ b/docs/rhaos-overrides.yaml
@@ -1,0 +1,16 @@
+# This file currently is just written (and read) by humans, but
+# it intends to define in a future machine-readable way the
+# desirhttps://github.com/ostreedev/ostree/pull/2569#issuecomment-1238416156ed RHEL packages we want to ship in OCP/RHCOS before
+# they're released by errata.
+# What we want in the future is related to https://issues.redhat.com/browse/GRPA-1358
+# Basically:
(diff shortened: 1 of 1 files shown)


docs/rhaos-overrides.yaml
```diff
@@ -0,0 +1,16 @@
+# This file currently is just written (and read) by humans, but
+# it intends to define in a future machine-readable way the
+# desirhttps://github.com/ostreedev/ostree/pull/2569#issuecomment-1238416156ed RHEL packages we want to ship in OCP/RHCOS before
```
travier, line 3:
```suggestion
# desired RHEL packages we want to ship in OCP/RHCOS before
```

docs/rhaos-overrides.yaml
```diff
@@ -0,0 +1,16 @@
...
+# they're released by errata.
+# What we want in the future is related to https://issues.redhat.com/browse/GRPA-1358
```
travier, line 5:
```suggestion
# What we want in the future is to implement https://issues.redhat.com/browse/COS-808
```

marmijo (approved):
/lgtm

Issue #986, opened 2022-09-07
RHCOS 4.12 s390x - Builds are failling because extension container uses non-s390x build image


jschintag:
On s390x, 4.12 builds are failling because `quay.io/centos/centos:stream8` is not available for s390x

```
[2/3] STEP 1/7: FROM quay.io/centos/centos:stream8 AS builder
Trying to pull quay.io/centos/centos:stream8...
Error: error creating build container: choosing an image from manifest list docker://quay.io/centos/centos:stream8: no image found in manifest list for architecture s390x, variant "", OS linux
error: exit status 125
```
[log.txt](https://github.com/openshift/os/files/9504172/log.txt)

/cc @jmarrero

jschintag:
Looking at the centos image, it seams stream9 also includes s390x.
https://quay.io/repository/centos/s390x?tab=tags
I don't know the needed package versions, but would it be possible to use the newer stream?

cgwalters:
I think we should actually be using RHEL (UBI) here.  We were just using stream to avoid entitlements for this section of the build; but we have the repos, we just need to plumb them through here too.

cgwalters:
Untested draft https://github.com/openshift/os/pull/987

cgwalters:
Hmm though actually because the MCO is having trouble relying on the network mode (xref current code in https://github.com/openshift/machine-config-operator/pull/3317 ) we could just drop the intermediate network server...

PR #987, opened 2022-09-07
extensions: Attempt to use ubi8 and repos too

cgwalters:
Closes: https://github.com/openshift/os/issues/986


jmarrero:
idk if createrepo_c is in the repos. Checking.

PR #988, opened 2022-09-07
extensions: use stream9 as builder because of s390x

jmarrero:
closes https://github.com/openshift/os/issues/986


jschintag:
Thanks @cgwalters and @jmarrero for taking such a quick look at it.

cgwalters:
CI doesn't cover this (although it could...that's another topic)

PR #989, opened 2022-09-07
tests: Add EL9 support for version checks

travier:
- Use `/etc/os-release.<osver>` to check versions
- Use single number versioning for SCOS as C9S does not have minor
  releases
- Skip shim from RHEL 8 in C9S
- Add a workaround to only read the first answer for package RELEASE as
  weirdly, querying the libgcc package on RHEL 8 returns the result
  twice.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -57,14 +57,6 @@
   osversion:
     - c9s
     - rhel-9.0
-- pattern: ext.config.version.rhel-major-version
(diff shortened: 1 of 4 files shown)


travier:
This will fail on 8.6 for an unknown reason that I'm investigating.

travier:
Found the weird reason. Somehow querying the libgcc package returns 2 times the result

PR #990, opened 2022-09-07
overlay.d/07c9s: Add OKD manifests and pre-pivot.sh script 

LorbusChris:
This commit adds manifests required for OKD, as well as the
`/bootstrap/pre-pivot.sh`script which is run by the installer [1].
The script rebases an FCOS-provisioned bootstrap node to SCOS before
continuing with the installation.
    
[1] https://github.com/openshift/installer/blob/d10d64c9843e37f500bcb8d66caf339e940b918f/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh#L27

/cc @vrutkovs
/cc @sherine-k 
/cc @lmzuccarelli
/cc @travier

### DIFF
--- a/overlay.d/07c9s/bootstrap/pre-pivot.sh
+++ b/overlay.d/07c9s/bootstrap/pre-pivot.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+set -euxo pipefail
+
+# This script is called by the OKD installer and runs in the
+# local context of the bootstrap node.
+# It rebases the booted operating system (e.g. Fedora CoreOS)
+# to CentOS Stream CoreOS 9
+
+# Load common functions
+. /usr/local/bin/release-image.sh
+
+# Copy manifests
+# Before rebasing the bootstrap node from FCOS to SCOS, ensure the OKD manifests are in place
+if [ ! -d /opt/openshift/openshift/ ]; then
+    mkdir -p /opt/openshift/openshift/
+fi
+cp -irvf manifests/* /opt/openshift/openshift/
+
+# Pivot to new os content
+MACHINE_OS_IMAGE=$(image_for centos-stream-coreos-9)
+rpm-ostree rebase --experimental "ostree-unverified-registry:${MACHINE_OS_IMAGE}"
+
+touch /opt/openshift/.pivot-done
+
+systemctl reboot
--- a/overlay.d/07c9s/bootstrap/manifests/90-okd-sca-pull-disable.yaml
+++ b/overlay.d/07c9s/bootstrap/manifests/90-okd-sca-pull-disable.yaml
@@ -1,0 +1,8 @@
+kind: Secret
+apiVersion: v1
+metadata:
+  name: support
+  namespace: openshift-config
+stringData:
+  scaPullDisabled: "true"
+type: Opaque
--- a/overlay.d/07c9s/statoverride
+++ b/overlay.d/07c9s/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>


overlay.d/07c9s/bootstrap/pre-pivot.sh
```diff
@@ -0,0 +1,37 @@
...
+if [ ! -d /opt/openshift/openshift/ ]; then
+    mkdir -p /opt/openshift/openshift/
+fi
+cp -irvf manifests/* /opt/openshift/openshift/
+
+# Disable all yum repos
+sed -i 's/^enabled=1/enabled=0/g' /etc/yum.repos.d/*.repo
+
+# Pivot to new ostree container image
+OS_IMAGE=$(image_for centos-stream-coreos-9)
```
LorbusChris, line 26: Note: This image doesn't exist yet in CI or any payloads. It'll be created with https://github.com/okd-project/okd-coreos-pipeline on operate-first.cloud, pushed to quay.io and imported into CI from there. A new release controller config will then be set up to create OKD/SCOS (CI) release payloads that include the image.

manifest-c9s.yaml
```diff
@@ -25,7 +25,8 @@ repos:
   - baseos
   - appstream
   - sig-nfv
-  # Temporarily include RHCOS 9 repo for cri-o, cri-tools, oc & hyperkube
+  - okd-copr
+  # Temporarily include RHCOS 9 repo for oc & hyperkube
```
travier, line 29: Can you update the `includepkg` list in `ci/prow-entrypoint.sh` too?

overlay.d/07c9s/bootstrap/pre-pivot.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -euxo pipefail
+
+# This script is called by the OKD installer and runs in the
+# local context of the bootstrap node.
+# It rebases the booted operating system (e.g. Fedora CoreOS)
+# to CentOS Stream CoreOS 9
+
+# Load common functions
+. /usr/local/bin/bootstrap-service-record.sh
```
travier, line 10: ?

travier:
Can you split the repo PR into it's own?

I don't understand the logic for the rest of the changes. Maybe if you wan to ship additional temporary files for OK we should use something like `/usr/lib/okd` instead of `/opt` or `/usr/local` and use `/var/lib/okd/something` for stamps / state?

overlay.d/07c9s/bootstrap/pre-pivot.sh
LorbusChris: This is run by the installer in the context of the bootstrap node: see https://github.com/openshift/installer/blob/d10d64c9843e37f500bcb8d66caf339e940b918f/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh#L27

LorbusChris:
Split the OKD-COPR repo addition out to https://github.com/openshift/os/pull/993

`/opt/openshift/openshift` is where the CVO expects manifests to be placed. When rebasing from a plain FCOS node, we have to make sure to copy them there before the rebase.

overlay.d/07c9s/bootstrap/pre-pivot.sh
```diff
@@ -0,0 +1,37 @@
...
+sed -i 's/^enabled=1/enabled=0/g' /etc/yum.repos.d/*.repo
+
+# Pivot to new ostree container image
+OS_IMAGE=$(image_for centos-stream-coreos-9)
+rpm-ostree rebase --experimental "ostree-unverified-registry:${OS_IMAGE}"
+
+# Delete cached rpm repo metadata
+rpm-ostree cleanup -m
+
+touch /opt/openshift/.pivot-done
```
LorbusChris, line 32: @travier this stamp file is an (undocumented) OKD installer API: https://github.com/openshift/installer/blob/d10d64c9843e37f500bcb8d66caf339e940b918f/data/data/bootstrap/files/usr/local/bin/bootstrap-pivot.sh#L10

This script will only be run if the file doesn't already exist.

manifest-c9s.yaml
LorbusChris: updated over in https://github.com/openshift/os/pull/993

cgwalters:
We shouldn't ship content in `/opt` (really `/var/opt`) in an ostree commit.  This stuff needs to be in `/usr`.

cgwalters:
/hold
per above

LorbusChris:
@cgwalters could we put this in `/bootstrap` instead? That's where the installer will expect `pre-pivot.sh`: https://github.com/openshift/installer/pull/6443. I've updated this PR, PTAL.

overlay.d/07c9s/bootstrap/pre-pivot.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -euxo pipefail
+
+# This script is called by the OKD installer and runs in the
+# local context of the bootstrap node.
+# It rebases the booted operating system (e.g. Fedora CoreOS)
+# to CentOS Stream CoreOS 9
+
+# Load common functions
+. /usr/local/bin/release-image.sh
```
cgwalters, line 10: I think it's problematic to have content shipped in the operating system read code injected from outside (in this case via Ignition).  Instead I think we should have a cleaner, well-defined API between the two.

Hmm, in this case though, this code is so small that what's the value in having it shipped by the OS image instead of fully injecting it via Ignition?

Since it's only for the bootstrap node, that seems to make even more sense since we don't care about it on production nodes.

overlay.d/07c9s/bootstrap/pre-pivot.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
cgwalters, line 1: Binaries shipped as part of the ostree image need to be in `/usr` (not `/usr/local`).  So *if* we were to ship this, it could be e.g. `/usr/libexec/ocp-node-bootstrap.sh` or so.

But I'm questioning whether we really need this tiny shell script in the OS versus injected via Ignition to the bootstrap node, which is where half of this logic is anyways.

LorbusChris: FCOS' layered image ships this script in `/bootstrap/pre-pivot.sh` today, and I wanted to re-use as much of that as possible.
You are right though, that we could just add this in the installer directly as ignition. I'll fold the bootstrap pivot into https://github.com/openshift/installer/pull/6443

PR #991, opened 2022-09-07
extensions: Enable SCOS extension builds with COSA

LorbusChris:
Not sure this'll work/

/cc @travier 
/cc @lmzuccarelli

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -7,8 +7,14 @@ RUN mkdir /os
 WORKDIR /os
 ADD . .
 ARG COSA
-RUN if [[ -z "$COSA" ]] ; then ci/get-ocp-repo.sh ; fi
-RUN rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}.yaml
+RUN variant="" \
+    && if [[ -f 'src/config.json' ]]; then \
+        variant="-"+$(jq --raw-output '."coreos-assembler.config-variant"' 'src/config.json'); \
+    fi \
+    && if [[ -z "$COSA" ]] && [ "$variant" != "scos" ] && [ "$variant" != "c9s" ]; then \
+        ci/get-ocp-repo.sh; \
+    fi \
+    && rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}${variant}.yaml
 
 ## Creates the repo metadata for the extensions & builds the go binary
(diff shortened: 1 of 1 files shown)


LorbusChris:
I wonder is the `FROM registry.ci.openshift.org/rhcos-devel/rhel-coreos:latest as os` getting replaced by COSA, or do we need to update that too for SCOS, too?

travier:
No need to CC me on all PRs here. I actively watch this repo 🙂

travier:
You can override some `FROM` in a container image with `--from` for podman builds but not sure that you can replace them all. You might need to make an entire copy of this Containerfile here for SCOS.

LorbusChris:
My question is, when this is run via COSA, does COSA replace that FROM directive in any way? (i.e. is this run in the oscontainer that was just composed, or does COSA really pull that image, which I imagine would have to come from from a previous build?)

/cc @cgwalters

cgwalters:
BTW @jmarrero wrote most of the code here so adding him

jmarrero:
This passed here, I do a quick test in the next hr locally and approve it if it looks good.

jmarrero:
@LorbusChris locally I get:

```
[1/3] STEP 6/6: RUN if [[ -z "$COSA" ]] ; then ci/get-ocp-repo.sh && variant="" ; else variant="-"+$(jq --raw-output '."coreos-assembler.config-variant"' 'src/config.json') ; fi     && rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}${variant}.yaml
jq: error: Could not open file src/config.json: No such file or directory
Error: error building at STEP "RUN if [[ -z "$COSA" ]] ; then ci/get-ocp-repo.sh && variant="" ; else variant="-"+$(jq --raw-output '."coreos-assembler.config-variant"' 'src/config.json') ; fi     && rpm-ostree compose extensions --rootfs=/ --output-dir=/usr/share/rpm-ostree/extensions/ {manifest,extensions}${variant}.yaml": error while running runtime: exit status 2
error: exit status 2
```

jmarrero:
> It's not immediately clear to me why the `src` folder wouldn't be created when running COSA locally. I'll have to dig in a little more it seems.

It is there, you prep the directories it's just not finding the config.json file and I don't see it in either the fedora config or os repositories. Is this specific to scos? Is there a config.json in a SCOS repo somewhere?

jmarrero:
Looks good locally now with cosa.

jmarrero:
giving it another test now.

jmarrero (approved):
/lgtm

PR #1001, opened 2022-09-09
github: Add ShellCheck workflow

travier:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1180

Added manually here (and not via the repo templates) as I don't think the bot has the right to make a pull request here (?)

Edit: Not sure the GitHub workflow run here. We'll likely have to adapt this one for Prow

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -197,6 +197,35 @@ validate() {
     git submodule update --init --recursive
     # Basic syntax check
     ./fedora-coreos-config/ci/validate
+
+    # Validate shell scripts with ShellCheck
+    if [[ -z "$(command -v shellcheck)" ]]; then
+        sudo dnf install -y ShellCheck
+    fi
+
+    local found_errors="false"
+    # Let's start with error, then we can do warning, info, style
+    local -r severity="error"
+
+    set +x
+    while IFS= read -r -d '' f; do
+        shebang="$(head -1 "${f}")"
+        if [[ "${f}" == *.sh ]] || \
+            [[ ${shebang} =~ ^#!/.*/bash.* ]] || \
+            [[ ${shebang} =~ ^#!/.*/env\ bash ]]; then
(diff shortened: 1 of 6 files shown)


travier:
Needs https://github.com/coreos/fedora-coreos-config/pull/1966

cgwalters:
> Edit: Not sure the GitHub workflow run here. We'll likely have to adapt this one for Prow

Currently Github Actions are disabled for the entire github.com/openshift organization.  But back then, actions by default had write access to the repository.  Nowadays, administrators can globally enforce read-only access, which basically obviates security concerns, so we could ask to have them turned on I think.

travier:
Waiting on https://github.com/openshift/os/pull/1002

travier:
I'd say let's start with this one as we already have a validation test and we can re-use the GitHub Action workflow later.

cgwalters:
> sudo dnf install -y ShellCheck
> sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?

Our default CI jobs in Prow run as non-root.  I think we should make another image for this that's pre-built; there may even be one.  Or ask to enable GH actions read-only.

travier:
Always interesting what we can find using Google:
- https://github.com/openshift/installer/blob/master/hack/shellcheck.sh
- https://quay.io/repository/coreos/shellcheck-alpine
Looking for the source/sync job now :)

cgwalters:
xref https://github.com/coreos/rpm-ostree/pull/4008 where a different team at RH has a custom Github Action for this

travier:
Blocked on https://github.com/coreos/fedora-coreos-config/issues/2117 + f-c-c bump

mike-nguyen (approved):

travier:
We ended up adding ShellCheck directly to the build root in https://github.com/coreos/fedora-coreos-config/pull/2119

Issue #1000, opened 2022-09-09
RHCOS 4.12 ppc64le - Fatal glibc error: CPU lacks ISA 3.00 support (POWER9 or later required)


mike-nguyen:
During buildextensions-container, the following error is observed.

```
[2/3] STEP 1/7: FROM quay.io/centos/centos:stream9 AS builder
Trying to pull quay.io/centos/centos:stream9...
Getting image source signatures
Copying blob sha256:9b09ed90406389312cd520cb33c32bab856c85bed2dc6f36202447ddcd7b87c5
Copying config sha256:dd933935f5be7e75951c3debb8370ad169a65b2d49176175bbda04b06c73b9e7
Writing manifest to image destination
Storing signatures
[2/3] STEP 2/7: COPY --from=os /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
--> 56ab1d5ff40
[2/3] STEP 3/7: RUN dnf install -y createrepo_c golang
Fatal glibc error: CPU lacks ISA 3.00 support (POWER9 or later required)
Error: error building at STEP "RUN dnf install -y createrepo_c golang": error while running runtime: exit status 127
error: exit status 127
```

cgwalters:
> Fatal glibc error: CPU lacks ISA 3.00 support (POWER9 or later required)

This is almost certainly the ppc64le equivalent of https://github.com/coreos/coreos-assembler/pull/3070/commits/28fc80b0f57b10e79e43253a12596db3f9f8de46 - I bet it would work to add a conditional there and use `-cpu power9`.

cgwalters:
cc @ravanelli

ravanelli:
> `cpu power9`.

ummmm, I can give it a try. But I think RHEL9 or CentOs9 wasn't suppose to run in a P8 host ...

cgwalters:
>  But I think RHEL9 or CentOs9 wasn't suppose to run in a P8 host ...

Ah right sorry, I think you're right in this case we are using KVM.   OK whee, so we're using stream9 to make s390x work, but we can't use stream9 for ppc64le in our current setup.  

Hmmmm.  I guess we could use stream8 only for ppc64le here for now...

cgwalters:
Or alternatively, use stream9 only for s390x.

[...]

PR #1004, opened 2022-09-12
extensions: use Fedora as a builder

cgwalters:
Closes: https://github.com/openshift/os/issues/1000

RHEL doesn't ship `createrepo_c`, so we were using CentOS Stream 8 for that.

Then we hit on the fact that C8S doesn't exist for s390x, so switched to C9S.

But RHEL9 made a push to require newer hardware baselines.  This is problematic for parts of the RHCOS pipeline which are deployed on Power8 hardware.

So...Fedora hasn't made the hardware baseline switches, so for now let's just use that as a lowest common denominator.  The base image and packages exist on all architectures we care about.


cgwalters:
OK so we're just waiting on s390x validation on this I think

ravanelli:
> OK so we're just waiting on s390x validation on this I think

The s390x test passed too

jmarrero:
Thanks for fixing this!

PR #1006, opened 2022-09-13
c9s.repo: Download gpg keys

LorbusChris:
This enables extension container builds, which are run in a non-CentOS root, where the key files are not present locally.

/cc @travier

### DIFF
--- a/c9s.repo
+++ b/c9s.repo
@@ -4,7 +4,7 @@ baseurl=http://mirror.stream.centos.org/9-stream/BaseOS/$basearch/os
 gpgcheck=1
 repo_gpgcheck=0
 enabled=1
-gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
+gpgkey=https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official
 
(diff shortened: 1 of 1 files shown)


cgwalters:
> This enables extension container builds, which are run in a non-CentOS root, where the key files are not present locally.

Why can't you install `distribution-gpg-keys` in whatever container you are using this repository from?

LorbusChris:
@cgwalters the buildroot is the previously composed scos container image (with the first FROM directive from https://github.com/openshift/os/blob/master/extensions/Dockerfile having been replaced by COSA), and c9s does not seem to ship the `distribution-gpg-keys` rpm anywhere. (there is https://cbs.centos.org/koji/packageinfo?packageID=7374, but it was last built last year, and only ships in openstack-*-build yum repos)

cgwalters:
It is in EPEL: https://dl.fedoraproject.org/pub/epel/9/Everything/x86_64/Packages/d/distribution-gpg-keys-1.77-1.el9.noarch.rpm

But...actually, for the extensions container builds for C9S, we can just use the .repo files from `/etc/yum.repos.d` - we don't need to use the ones from the config git.

PR #1007, opened 2022-09-13
Revert "extensions: Enable SCOS extension builds with COSA"

LorbusChris:
This reverts commit 7bcca86eb807f27e4b0899f7c4443ae9c0d153d2.

When running `cosa build-extensions-container`, the `src/config.json` file is not present, so the condition for setting a non-empty variant will never be true.

This can be mitigated by running `cosa update-variant default scos` beforehand, pointing the symlinks for `{manifest,extensions}.yaml` to the desired variant.

/cc @jmarrero 
/cc @travier 

I think it might be best to have COSA pass the variant to the Dockerfile as an arg, but for now this can be worked around per the above.

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -7,14 +7,8 @@ RUN mkdir /os
 WORKDIR /os
 ADD . .
 ARG COSA
-RUN variant="" \
-    && if [[ -f 'src/config.json' ]]; then \
-        variant="-"+$(jq --raw-output '."coreos-assembler.config-variant"' 'src/config.json'); \
-    fi \
-    && if [[ -z "$COSA" ]] && [ "$variant" != "scos" ] && [ "$variant" != "c9s" ]; then \
-        ci/get-ocp-repo.sh; \
-    fi \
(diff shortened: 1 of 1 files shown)


jmarrero:
I see, so we don't need to change anything for this release? SCOS pipeline will use the correct repos by calling `cosa update-variant default scos` ?

jmarrero (approved):
lgtm

cgwalters:
> When running cosa build-extensions-container, 

That's one possible path for building the extensions container.  But, note that part of our design for the extensions container is that it has a `Dockerfile` that can also be built using any container build system.  (`cosa build-extensions-container` is just a wrapper for that)

Unless OKD has an explicit reason to use cosa for this, I'd suggest using a container-native build process for it.  We already have one in Prow:
https://github.com/openshift/release/blob/ceac95613d29f9c51b397757f6272f03a1932f86/ci-operator/config/openshift/os/openshift-os-master__periodic.yaml#L102

LorbusChris:
Should we get this in?

jmarrero:
I guess I need to standardize my lgtm to /lgtm if not I will keep missing these.

travier:
Note that `cosa update-variant` is not meant to be used in CI as it changes the content of the repo. It's only there for version updates.

PR #1011, opened 2022-09-14
Convert `platforms.yaml` to a symlink

bgilbert:
Now that `platforms.yaml` has landed in FCOS testing-devel, we can use that copy rather than replicating it here.

Followup to #1010.


jlebon (approved):
Woohoo!

jlebon:
For the rhcos-90 one, I see the same failure on e.g. https://github.com/openshift/os/pull/1103. For the scos-9 one, the failure here is different (it failed on a kernel warning), but looking at [the console](https://github.com/openshift/os/files/10384675/console.txt), I don't think it's related to this change. So my vote is to go ahead and override these tests so we can get this in.

bgilbert:
SGTM!

/override ci/prow/rhcos-90-build-test-qemu
/unhold

PR #1012, opened 2022-09-16
Add `--fail` to all invocations of `curl`

cgwalters:
See https://github.com/openshift/machine-config-operator/pull/3317#issuecomment-1248765511 where it looks like release-controller failed and we got a corrupted repo file.  It would have been a bit easier to debug if we printed the error.


travier (commented):
/lgtm

PR #1018, opened 2022-09-21
ci/prow-entrypoint: Add an `update-variant` command

cgwalters:
So...basically we're in a state where we have this nice `cosa update-variant` command except there's a lot of heavy lifting around the repositories and even some changes to the repo configs that happen in *this* entrypoint.  The latter of which is a bigger problem because changes like that won't
happen for production builds.

Anyways back to the original problem I was looking at: All the entrypoints here really only support a toplevel thing that does it all like `/src/ci/prow-entrypoint.sh rhcos-90-build-test-qemu` but what I want to do is *just* build the oscontainer.

Add an `update-variant` wrapper that does the same thing as the cosa command but also runs the heavily load-bearing `prepare_repos` function.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -229,6 +229,10 @@ main () {
             cosa_init "rhel-coreos-8"
             cosa_build
             ;;
+        "update-variant")
+            cosa update-variant default "$2"
+            prepare_repos
+            ;;
         "rhcos-cosa-prow-pr-ci")
             setup_user
             cosa_init "rhel-coreos-8"


travier:
Where you would need that? This changes the repo content (changes the symlinks) and the main idea behind the direct variant support in cosa was to avoid doing that to not "dirty" the repo state. Can we use that instead?

cgwalters:
Hmm; why does `cosa update-variant` not just replace the contents of `src/config.json` ?

travier:
I can make a new one but it would require cleaning caches and builds.

LorbusChris:
On a side note, we're using `cosa update-variant` in the [SCOS pipeline](https://github.com/okd-project/okd-coreos-pipeline) because the extensions Dockerfile hardcodes `{manifest,extensions}.yaml` so the symlinks have to be updated first (and we're not just building the Dockerfile directly, as we'd loose the extensions artifact metadata that way).

PR #1021, opened 2022-09-28
rhel9: Add `/etc/localtime` pointing to `UTC` by default

cgwalters:
See https://issues.redhat.com/browse/LOG-3117

Something is creating it for RHCOS8, but seemingly not in the RHCOS9 build.  (I bet Anaconda does this by default, so it will be on most Anaconda-based systems though)

This ensures that containers can mount it from the host; though we should also chase down and fix those containers to do this in a better way (one approach is to mount all of `/etc`).

### DIFF
--- a/common-el9.yaml
+++ b/common-el9.yaml
@@ -26,6 +26,14 @@ postprocess:
      grep -q '# RHEL-only: Disable /tmp on tmpfs' /usr/lib/systemd/system/basic.target
      echo '# RHCOS-only: we follow the Fedora/upstream default' >> /usr/lib/systemd/system/basic.target
      echo 'Wants=tmp.mount' >> /usr/lib/systemd/system/basic.target
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+     # See https://issues.redhat.com/browse/LOG-3117
+     # Something changed between rhel8 and rhel9 to not generate this by default
+     # but we have containers that expect it to be mounted so for now let's continue
+     # generating it.
+     ln -sr /usr/share/zoneinfo/UTC /etc/localtime
 
 # Packages that are only for SCOS & RHCOS 9
 packages:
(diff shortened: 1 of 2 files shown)


HuijingHei:
Check on FCOS also does not have `/etc/localtime`, just curious if can reproduce on FCOS
```
[core@cosa-devsh ~]$ ll /etc/localtime /usr/share/zoneinfo/UTC
ls: cannot access '/etc/localtime': No such file or directory
-rw-r--r--. 17 root root 114 Jan  1  1970 /usr/share/zoneinfo/UTC
[core@cosa-devsh ~]$ rpm -q systemd
systemd-250.8-1.fc36.x86_64
```

HuijingHei (approved):
LGTM

lucab (commented):
If we really want to, I'd suggest we do this through a systemd-tmpfiles snippet instead.

But I'd rather not do this, as it is [well defined](https://www.freedesktop.org/software/systemd/man/localtime.html) that `/etc/localtime` may not be present:
```
If /etc/localtime is missing, the default "UTC" timezone is used.
```

cgwalters:
I completely agree that code should be changed to work without this.  But it's hard to avoid things picking up implementation details.

I think I'd say let's give some time for the logging folks to fix their container, then try to remove this again later.

travier:
This is for C9S only (`manifest-c9s.yaml`). You want to add that to https://github.com/openshift/os/blob/master/common-el9.yaml or do it via a `tmpfiles.d` entry like Luca said.

cgwalters:
I don't want to do it via `tmpfiles.d` because that becomes node-local state, whereas with this approach we can actually remove it when we believe the workaround isn't needed anymore.

travier:
> I don't want to do it via `tmpfiles.d` because that becomes node-local state, whereas with this approach we can actually remove it when we believe the workaround isn't needed anymore.

Good point!
/lgtm
/unhold

PR #1025, opened 2022-09-30
[release-4.10] OCPBUGS-1938: Bump fedora-coreos-config submodule for ISO mount fix

jlebon:
```
Colin Walters (1):
      live-generator: Use a loop to work around potential races with udev on el8
```

I.e. https://github.com/coreos/fedora-coreos-config/pull/1986


jlebon:
FYI, I've tested this commit on top of 4.10 manually already (see https://github.com/coreos/fedora-coreos-config/pull/1986#issuecomment-1260056393).

cgwalters:
From my PoV we should just bypass the official bug backport process for this; it's basically just overhead.

Crucially, only *new nodes* (or whole new clusters) will hit the new code, so risk of any wide fallout is really low.

sdodson:
We'll get things merged in order, the backport process is meant not only to ensure testing but to provide assurances that someone who was delivered this fix in 4.10.z doesn't then go repeat the process in 4.11 and lack the fixes. I'll grease the wheels so that 4.11 PR merges, it only needs lgtm now. Once that merges I'll make sure this gets in too.

ashcrow:
This backport still needs a lgtm

sdodson:
Failure seems legit, 8.4 lacking newest os-tree?

cgwalters:
It's a bug in our CI configuration here https://github.com/openshift/release/pull/32820
(Which note, is distinct from the ART build configuration, which is its own problem)

The actual failure is because https://github.com/coreos/coreos-assembler/pull/3114 merged, otherwise we might not have noticed CI was using cosa main.

cgwalters:
OK fun, this is now hitting ye old 9p issues https://github.com/coreos/coreos-assembler/issues/1812

/retest

mike-nguyen:
I build this PR locally and RHCOS built with no issues.  The kola and testiso test also passed.

PR #1031, opened 2022-10-20
c9s: Add `c9s-` to repo names, use public OCP ART repo

cgwalters:
In the future, we may want to mix in *some* C9S packages into an otherwise "RHCOS9" build.  We hence need the ability to enable both repositories at the same time.

Also, it's just generally more obvious (i.e. debuggable) to have the stream included in the repository name.

Next, add the public OCP RPM dependency repo, and use it instead of the private one.  The downside is we aren't getting continually refreshed content.  But, I don't understand why we don't just fix that and publish these RPMs continuously.

Finally, it seems that `conmon-rs` is not yet in that repo, so we keep the OKD COPR for now to get that.  But I think we should fix this and publish that in the OCP RPM repo too.

### DIFF
--- a/ocp-public.repo
+++ b/ocp-public.repo
@@ -1,0 +1,10 @@
+[ocp-412-el8-beta-public]
+name=OpenShift 4.12 RHEL8 RPMs
+baseurl=https://mirror.openshift.com/pub/openshift-v4/$basearch/dependencies/rpms/4.12-el8-beta/
+# For some sad reason, the OCP RPMs aren't GPG signed?  Anyways we should stop caring about RPM GPG
+# and move to pulling RPMs from OCI artifacts
+gpgcheck=0
+repo_gpgcheck=0
+enabled=1
+gpgkey=file:///usr/share/distribution-gpg-keys/redhat/RPM-GPG-KEY-redhat8-release
+
--- a/c9s.repo
+++ b/c9s.repo
@@ -1,4 +1,4 @@
-[baseos]
+[c9s-baseos]
 name=CentOS Stream 9 - BaseOS
 baseurl=http://mirror.stream.centos.org/9-stream/BaseOS/$basearch/os
 gpgcheck=1
@@ -6,7 +6,7 @@ repo_gpgcheck=0
 enabled=1
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
-[appstream]
+[c9s-appstream]
 name=CentOS Stream 9 - AppStream
 baseurl=http://mirror.stream.centos.org/9-stream/AppStream/$basearch/os
 gpgcheck=1
@@ -14,7 +14,7 @@ repo_gpgcheck=0
 enabled=1
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
-[nfv]
+[c9s-nfv]
 name=CentOS Stream 9 - NFV
 baseurl=http://mirror.stream.centos.org/9-stream/NFV/$basearch/os
 gpgcheck=1
@@ -22,7 +22,7 @@ repo_gpgcheck=0
 enabled=1
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-Official
 
-[rt]
+[c9s-rt]
 name=CentOS Stream 9 - RT
 baseurl=http://mirror.stream.centos.org/9-stream/RT/$basearch/os
 gpgcheck=1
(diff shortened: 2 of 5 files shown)


cgwalters:
Bigger picture actually...there I think is some confusion over whether c9s/scos is "owned" by this repo or by https://github.com/openshift/okd-machine-os/

(I guess really related to this is whether c9s/scos should also be built as e.g. a new variant in the new forthcoming RHCOS pipeline, or whether it's entirely up to okd-machine-os.  But note today a tricky thing here is that I think we actually do want c9s/scos for all architectures, and I don't think the Tekton stuff is doing that today, right?)

aleskandro:
Hi @cgwalters the tekton pipeline is almost ready to consider building for other arches, as far as we have infra to build scos on other arches. For arm64 should be possible

LorbusChris:
/hold
in case someone wants to look at the Prow SCOS build failures.

travier:
I don't think we should use the public mirrors. Having SCOS in CI here has value if it uses the same package versions as the ones we use in latest development RHCOS. Otherwise we're just testing something else and not even what will OKD/SCOS use.

Right now we don't test booting clusters for PRs so cri-o & kubelet don't get that much testing. But I'd say that's planned?

travier:
And I think we really need to find a way to push those builds publicly and not rely on openshift builds for cri-o & kubelet for SCOS

ocp-public.repo
```diff
@@ -0,0 +1,10 @@
+[ocp-412-el8-beta-public]
```
aleskandro, line 1: Hello @cgwalters,

looking better at the PR, I see you're including the public mirror for the rpms (cri-o, clients, hyperkube...). That repo is only available for the x86_64 arch, though. We relied on the builds from the artifacts and hyperkube images (both from Prow and the payload tekton pipeline) to get those packages till now. Any plan to have the same repo for arm64 exposed?

LorbusChris:
> And I think we really need to find a way to push those builds publicly and not rely on openshift builds for cri-o & kubelet for SCOS

+1. The CentOS Cloud SIG has already requested builds for those currently missing components be done on CentOS Stream koji, so we'll eventually have to do that.

travier:
I think this needs more discussion so clearing LGTM
/lgtm cancel

cgwalters:
> +1. The CentOS Cloud SIG has already requested builds for those currently missing components be done on CentOS Stream koji, so we'll eventually have to do that.

So...man this is a really confusing topic.  Today, for OCP ART has automation that auto-syncs when e.g. one commits to github.com/openshift/kubernetes and auto-builds an RPM.  Are we going to set up parallel infrastructure that auto-builds in CS koji?

ocp-public.repo
cgwalters: The problem seems to be here
https://github.com/openshift/aos-cd-jobs/blob/70b48b971e244c931a93185f10910d5acd11ac20/jobs/build/publish-rpms/Jenkinsfile#L47

But it's not obvious to me what's in that bucket etc.

aleskandro: Are those jobs running on x86-only agents? If yes, we'd need other changes [here](https://github.com/openshift/aos-cd-jobs/blob/70b48b971e244c931a93185f10910d5acd11ac20/jobs/build/publish-rpms/collect_deps.py), right?

+1 for @travier and @LorbusChris from my side

LorbusChris:
> So...man this is a really confusing topic. Today, for OCP ART has automation that auto-syncs when e.g. one commits to github.com/openshift/kubernetes and auto-builds an RPM. Are we going to set up parallel infrastructure that auto-builds in CS koji?

It's a hard requirement for C9S not to sneak in anything build in a RHEL env.
So we'll definitely have to do that for rpm-ostree, cri-o and other things that do not live in the openshift org. For those I'm hoping we can leverage packit for building these things directly in C9S koji for each commit.
For `openshift/oc` and `openshift/kubernetes` we might be able to do c9s image based builds in Prow and sync them to a public yum repo somewhere (similar to ART does for OCP today)

cgwalters:
> It's a hard requirement for C9S not to sneak in anything build in a RHEL env, and preferably on their infra.

Yeah, another way to look at this is ensuring the sources are published, and actually another benefit is we can do builds and tests even if internal RHT infrastructure is down.

> So we'll definitely have to do that for rpm-ostree, cri-o and other things that do not live in the openshift org.

Careful now; rpm-ostree is part of RHEL.  While we have shipped different versions as part of OCP for RHEL8, I hope to end that practice in the RHEL9 era by having OCP builds rebuild/re-ship a specific build that was committed to C9S.  (Kind of like the kernel pin)

travier:
We should only ever need it for cri-o, oc & hyperkube. All other projects either are in RHEL now an thus are updated directly in Stream, or an managed in CentOS SIGs.

travier:
We also have kata-containers & kernel-rt from SIGs.

travier:
/close
Closing as we choose another option.

Issue #799, opened 2022-05-11
Rework build process to generate `rhel-coreos-base` distinct from `ocp-rhel-coreos`


cgwalters:
# Reworking RHEL CoreOS to be more like OKD and towards quay.io/openshift/node-base:rhel10

This pre-enhancement originated in [this github issue](https://github.com/openshift/os/issues/799).

A foundational decision in early on OpenShift 4 was to create RHEL CoreOS.  Key
aspects of this were:

* kubelet would not be containerized (negative experience with "system containers")
* More crucially, we wanted to ship a *tested* combination of operating system and cluster
* Also, the operating system updates should come in a container image

We're several years in now, and have learned a lot.  This proposal calls for
reworking how we build things, but will avoid changing these key aspects.
 
## Rework RHCOS disk images to not have OCP content

When we speak of RHEL CoreOS, there are two independent things at play:

* disk images (AMI, qcow2, ISO, etc.)
* OS update container

In this base proposal, the disk images shift to *only RHEL content*.

- `kubelet` will not be in the AMI.
- The version will change to something of the form `$rhel.$datestamp`, e.g. `9.2.20220510.1`

Additionally, there will be a new container image called `rhel-coreos-base` that
will exactly match this.

These disk images will generally only be updated at the GA release of each RHEL, and will not contain security updates.

In phase 0, openshift-installer will continue to have [rhcos.json](https://github.com/openshift/installer/blob/release-4.14/data/data/coreos/rhcos.json).  Disk images will continue to be provided at e.g. [mirror.openshift.com](https://mirror.openshift.com/pub/openshift-v4/dependencies/rhcos/4.14/).  

However, the disk images will be much more likely to be shared across OCP releases in a *bit for bit* fashion.

## machine-os-content/rhel-coreos-9

The key change here is that OCP content, including `kubelet` move into a container
image that *derives* from this base image.  One can imagine it as the following `Containerfile`:

```dockerfile
FROM rhel-coreos-base
RUN rpm-ostree install openshift-hyperkube
```

This is in fact [currently done for OKD](https://github.com/openshift/okd-machine-os/blob/master/Dockerfile).

```mermaid
flowchart TD
    rpms[RHEL rpms] --> base[quay.io/openshift/rhel-coreos-base:9]-- Add kubelet, crio, openvswitch --> ocpnode[quay.io/openshift/rhel-coreos:9]
```

In phase 0, this new image will likely be built by the current CoreOS pipeline.

[...]

PR #1032, opened 2022-10-21
c9s: Add distribution-gpg-keys rpm from EPEL

LorbusChris:
This is required for the extensions build running inside the base OS container via COSA as otherwise the GPG key files are not available there.

### DIFF
--- a/c9s.repo
+++ b/c9s.repo
@@ -53,3 +53,11 @@ gpgcheck=1
 repo_gpgcheck=0
 enabled=1
 gpgkey=https://download.copr.fedorainfracloud.org/results/@OKD/okd/pubkey.gpg
+
+[epel-9]
+name=Extra Packages for Enterprise Linux 9
+baseurl=https://dl.fedoraproject.org/pub/epel/9/Everything/$basearch
+gpgcheck=1
+repo_gpgcheck=0
+enabled=1
+gpgkey=file:///usr/share/distribution-gpg-keys/epel/RPM-GPG-KEY-EPEL-9
--- a/manifest-c9s.yaml
+++ b/manifest-c9s.yaml
@@ -26,6 +26,7 @@ repos:
   - appstream
   - sig-nfv
   - okd-copr
+  - epel-9
   # Temporarily include RHCOS 9 repo for oc & hyperkube
   - rhel-9-server-ose
 
@@ -119,6 +120,9 @@ packages:
  # We include the generic release package and tweak the os-release info in a
  # post-proces script
  - centos-release
+ # Include distribution-gpg-keys for having the GPG keys available
+ # when building the extensions image within the base container via COSA
+ - distribution-gpg-keys
 
 # Packages pinned to specific repos in SCOS 9
 repo-packages:
@@ -136,3 +140,6 @@ repo-packages:
     packages:
       - cri-o
       - cri-tools
+  - repo: epel-9
+    packages:
+      - distribution-gpg-keys


travier (commented):
/lgtm

travier:
But I'm wondering, which package from EPEL are we using? Aren't those key available in the C9S repos?

LorbusChris:
The `distribution-gpg-keys` rpm is not available in C9S itself unfortunately (https://kojihub.stream.centos.org/koji/search?match=glob&type=package&terms=distribution-gpg-keys). We only use the EPEL repo to get that one package.

We currently have this workaround in place: https://github.com/okd-project/okd-coreos-pipeline/blob/581420489cb09616c5b2225869d6e96995712c75/manifests/tekton/tasks/base/cosa-init.yaml#L64-L65

See also https://github.com/openshift/os/pull/1006

cgwalters:
If we were to do this only in RHCOS/C9S, it'd create another divergence from FCOS - i.e. we should consider this also for FCOS if we were to do this.  Oh wait, we're actually doing this only for C9S, which means *just that stream* diverges a bit.

On one hand, the idea of adding `distribution-gpg-keys` actually aligns with the conceptual shift of *CoreOS to be a *base image*.  It's certainly convenient to have these keys there.

It's a bit unfortunate that this package is in EPEL, because enabling the EPEL repos by default is a whole other world of stuff.

All this said, which actual keys are you needing?  And for what packages? I think we talked about this before, but can you remind me?  Do you have a link to a PR/code that needs this?

cgwalters:
Oh I see, sorry you did link it here.  OK, right and I'm paging back in the context from https://github.com/openshift/os/pull/1006

(Ohhh wow and I see https://github.com/okd-project/okd-coreos-pipeline/blob/581420489cb09616c5b2225869d6e96995712c75/manifests/tekton/tasks/base/cosa-init.yaml#L80 - so it begins... 

I think honestly we are all under a bit of confusion as to who builds scos and how/where and with what config.  https://github.com/okd-project/okd-coreos-pipeline/commit/e7fc6d895e8064696e1f93651b49ab5e58aca63b seems exactly like the kind of thing that conceptually should have been in *this* repository right? )

Anyways um...but wait is the problem that you're getting the `c9s.repo` file in this git repo into the extensions container build?  Why isn't it using the `.repo` files from `centos-release`?

LorbusChris:
> (Ohhh wow and I see https://github.com/okd-project/okd-coreos-pipeline/blob/581420489cb09616c5b2225869d6e96995712c75/manifests/tekton/tasks/base/cosa-init.yaml#L80 - so it begins...

indeed :joy: :sob: 

> I think honestly we are all under a bit of confusion as to who builds scos and how/where and with what config. [okd-project/okd-coreos-pipeline@e7fc6d8](https://github.com/okd-project/okd-coreos-pipeline/commit/e7fc6d895e8064696e1f93651b49ab5e58aca63b) seems exactly like the kind of thing that conceptually should have been in _this_ repository right? )

The plan has always been to involve the CoreOS team as much as possible, but it was made quite clear that the there are limited resources to work on the OKD pipeline and that the team had to focus on building RHCOS 9. Also we needed (and still need) something that does not require an `/ok-to-test` for external folks. Having the pipeline independent from the manifests is good I think, even though it'd probably be better to pin packages in a manifest properly (could be in a fork, which is then PR'd here). The repo to build from is actually a parameter of the pipeline, so building from a forked repo is trivial.

> Anyways um...but wait is the problem that you're getting the `c9s.repo` file in this git repo into the extensions container build? Why isn't it using the `.repo` files from `centos-release`?

Yes that's exactly the problem. Everything is copied into the container to build the extensions, which is required to get the manifests etc in there (https://github.com/openshift/os/blob/de8da2165b6af189df27a493ceeaa806091605eb/extensions/Dockerfile#L8). Is there even a way to tell `rpm-ostree compose` to _not_ use the .repo file(s) in the dir it's invoked from?

LorbusChris:
On a side note: The extensions build regularly breaks for us for so far unknown reasons. I think it would be clearer to do the extensions build in the same env as the base build instead of in a nested container.

travier:
Is this because the builds run in RHCOS? https://github.com/openshift/os/blob/master/extensions/Dockerfile#L2
You should override the FROM line with the SCOS OCI image instead.

\lgtm cancel

cgwalters:
> Is this because the builds run in RHCOS? https://github.com/openshift/os/blob/master/extensions/Dockerfile#L2
> You should override the FROM line with the SCOS OCI image instead.

Yeah, `cosa build-extensions-container` does this automatically.  But the messy thing here is that cosa is all about generating base images, whereas in the new world we want OKD to derive from FCOS.  (And we *also* want OCP-RHCOS to [drive from pure RHEL](https://github.com/openshift/os/issues/799) I think)

Though at at even higher level, to me again the extensions container is just a crutch - anyone who wants to inject code into the host can do it on their own in a fully "container native" way, although it does take over OS updates.

LorbusChris:
The problem is that the extensions build runs inside an SCOS base container (with the FROM directive being replaced automatically by COSA as Colin mentioned), and that very SCOS base does not have the GPG keys at the locations specified in `c9s.repo`. The `distribution-gpg-keys` RPM does ship them at those locations, so adding it to the base SCOS compose solves this problem.

I agree that the extensions container is a crutch (and the nested way in which its currently built, too). We're not currently shipping it with OKD/SCOS. However if we wanted to do that, we'd need the metadata that `cosa build-extensions-container` generates, so we couldn't just do the extensions container build with podman/without COSA.

Happy to talk about this synchronously, however next week I'll only be available on Wednesday or Friday (due to a public holiday on Monday, travels on Tuesday, and the Mini DevConf.cz in Brno on Thursday :slightly_smiling_face: ).

travier:
Ah OK, now I think I understand. I don't really like adding a dependency on EPEL. Could the SCOS pipeline do a sed on the c9s repo config files to fix the path before starting the build-extensions-container step? We could also ship symlinks in the image which is also fine for me.

LorbusChris:
@travier we currently hack around this with `sed` in the pipeline, which we'd like to avoid.
I think including this RPM is the most straightforward way to solve this.
EPEL isn't an ideal source, but we can work on getting `distribution-gpg-keys` packaged for c9s properly as a follow-up.

LorbusChris:
@travier please see the comment above. Can we merge this?

cgwalters:
One thing I could imagine doing here is teaching the dnf stack to automatically look in `/etc/pki/rpm-gpg/$x` if the repo file references `/usr/share/distribution-gpg-keys/$x` and it doesn't exist.

This issue is annoying I agree because otherwise I think pretty much everything else is set up to work in both src != target mode (how we build the "base image" with cosa) and src == target (extensions container).

Perhaps for now, we could carry the sed invocation inside the extensions container build, or do it automatically in coreos-assembler?

cgwalters:
It's worth noting here that this isn't biting FCOS as it doesn't ship an extensions container today.

travier:
Alternative in https://github.com/openshift/os/pull/1116

travier:
Better option in https://github.com/openshift/os/pull/1118

PR #1038, opened 2022-10-27
OCPBUGS-2980: [release-4.11] Pin to libsemanage-2.9-8.el8

HuijingHei:
```
[jlebon]
  With -9.el8, ext.config.rebuild-selinux-policy fails:
  https://github.com/openshift/os/issues/1036

  We need to debug this, but for now let's unblock CI and dev pipelines.
```
cherry-pick https://github.com/openshift/os/commit/247e64abe9d8225ea852222da689bfb5c99adf65


travier:
We'll need a Jira card.

PR #1044, opened 2022-10-28
OCPBUGS-2926: platforms: add secondary serial console on OpenStack/VirtualBox/VMware

bgilbert:
Previously, `openstack console log <server>` was useful for collecting console logs from OpenStack VMs, and the serial console was similarly useful in VMware, but we disabled serial console output on both platforms.  Re-enable it as a secondary console so it gets as much information as possible without interfering with the graphical console. Do the same on VirtualBox to stay in sync with FCOS.

Untested on OpenStack.

Fixes: https://issues.redhat.com/browse/OCPBUGS-2926
xref: https://github.com/coreos/fedora-coreos-config/pull/2038

### DIFF
--- a/platforms.yaml
+++ b/platforms.yaml
@@ -35,6 +35,17 @@ aarch64:
     kernel_arguments:
       - console=tty0
       - console=ttyAMA0,115200n8
+  openstack:
+    # Graphical console primary, serial console available for logging
+    # https://docs.openstack.org/diskimage-builder/latest/elements/bootloader/README.html
+    # https://issues.redhat.com/browse/OCPBUGS-2926
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyAMA0,115200n8
+      - console=tty0
   packet:
     # https://metal.equinix.com/developers/docs/resilience-recovery/serial-over-ssh/#limitations
     grub_commands:
@@ -51,6 +62,14 @@ aarch64:
       - terminal_input serial console
       - terminal_output serial console
 ppc64le:
+  openstack:
+    # Graphical console primary, serial console available for logging
+    # petitboot doesn't understand GRUB console commands
+    # https://docs.openstack.org/diskimage-builder/latest/elements/bootloader/README.html
+    # https://issues.redhat.com/browse/OCPBUGS-2926
+    kernel_arguments:
+      - console=hvc0
+      - console=tty0
   qemu:
     # petitboot doesn't understand GRUB console commands, but we need to
     # pass console kargs
@@ -99,6 +118,17 @@ x86_64:
     kernel_arguments:
       - console=tty0
       - console=ttyS0,115200n8
+  openstack:
+    # Graphical console primary, serial console available for logging
+    # https://docs.openstack.org/diskimage-builder/latest/elements/bootloader/README.html
+    # https://issues.redhat.com/browse/OCPBUGS-2926
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyS0,115200n8
+      - console=tty0
   packet:
     # https://metal.equinix.com/developers/docs/resilience-recovery/serial-over-ssh/#limitations
     grub_commands:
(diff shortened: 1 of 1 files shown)


travier:
/hold
We'll merge that right after the FCOS changes land.

EmilienM:
Thanks @bgilbert ! This looks good to me but I'll ping a few other folks to confirm from the OpenStack side:
@stephenfin @mdbooth

bgilbert:
/retest

The Fedora CoreOS version merged.  It'd be nice to get some OpenStack testing here though.

mdbooth:
> /retest
> 
> The Fedora CoreOS version merged. It'd be nice to get some OpenStack testing here though.

I'll do my best. How can I test this?

bgilbert:
@mdbooth Thanks!  I'll send you an OpenStack image with the change.

mdbooth:
@bgilbert New image fixes console log on OpenStack. Thanks!

/lgtm

bgilbert:
Great, thanks for testing!

Lifting hold from https://github.com/openshift/os/pull/1044#issuecomment-1294693361 since the FCOS changes have landed.

/hold release

platforms.yaml
```diff
@@ -99,6 +118,17 @@ x86_64:
...
+    # Graphical console primary, serial console available for logging
+    # https://docs.openstack.org/diskimage-builder/latest/elements/bootloader/README.html
+    # https://issues.redhat.com/browse/OCPBUGS-2926
+    grub_commands:
+      - serial --speed=115200
+      - terminal_input serial console
+      - terminal_output serial console
+    kernel_arguments:
+      - console=ttyS0,115200n8
+      - console=tty0
```
SeanMooney, line 131: this is more or less inline with our documentation 
https://docs.openstack.org/image-guide/centos-image.html#configure-console

a standard centos 9 cloud image has

BOOT_IMAGE=(hd0,msdos1)/boot/vmlinuz-5.14.0-165.el9.x86_64 root=UUID=8db50efa-f8ad-4ee1-a332-07c7abb146c6 ro console=ttyS0,115200n8 no_timer_check net.ifnames=0 crashkernel=1G-4G:192M,4G-64G:256M,64G-:512M

ubuntu 22.04 is a little more minimal 

BOOT_IMAGE=/boot/vmlinuz-5.15.0-48-generic root=LABEL=cloudimg-rootfs ro console=tty1 console=ttyS0

but in general i believe this is more or less correct

in my experience i have not needed to set the grub_commands to enable serial output as in most distros adding the kernel args is enough.

both set 

ubuntu sets

# Set the grub console type
GRUB_TERMINAL=console

and 
centos sets 
GRUB_TERMINAL_OUTPUT="console"

in there cloud images by default
so presumably that is what is being configured by 

  - terminal_input serial console
  - terminal_output serial console

overall i think this should work but its probably more verbose than required

bgilbert: Yup, we wanted to preserve graphical console access (`console=tty0`) and allow the GRUB menu to be accessed via serial console.

SeanMooney: in general dont think you need to do anything special to preserve the graphical console.

in OpenStack we generate XML like this

```
<serial type='tcp'>
      <source mode='bind' host='192.168.1.167' service='10004' tls='no'/>
      <protocol type='raw'/>
      <log file='/var/lib/nova/instances/97b112d3-7702-42b6-97d3-1e582961cd09/console.log' append='off'/>
      <target type='isa-serial' port='0'>
        <model name='isa-serial'/>
      </target>
      <alias name='serial0'/>
    </serial>
    <console type='tcp'>
      <source mode='bind' host='192.168.1.167' service='10004' tls='no'/>
      <protocol type='raw'/>
      <log file='/var/lib/nova/instances/97b112d3-7702-42b6-97d3-1e582961cd09/console.log' append='off'/>
      <target type='serial' port='0'/>
      <alias name='serial0'/>
    </console>
```

we define a serial port then use that as the source for the console

```
  <graphics type='vnc' port='5904' autoport='yes' listen='192.168.1.167'>
      <listen type='address' address='192.168.1.167'/>
    </graphics>
    <audio id='1' type='none'/>
    <video>
      <model type='virtio' heads='1' primary='yes'/>
      <alias name='video0'/>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x01' function='0x0'/>
    </video>

```

then later we define a graphics device of type vnc or spice that exports that console

i generally have had good experience just setting "console=tty0 console=ttyS0,115200n8"

the kernel args are more or less identical to what i would generally set although i always listed the serial console second.

so as i said the kernel argmumets look correct to me and should work to allow the console to be accessed either via horizon or via the console log command.


the grub_commands I'm less sure about but i don't think they would break anything if included so without testing the final image by botting a VM and installing a graphical ide i don't see any redflags here.

bgilbert: Right, it's the `console=tty0` argument that directs the kernel to write to the graphical console.  If you only specify `console=ttyS0,115200n8` it'll only send output to the serial port.  Regarding the order of the `console` arguments, the last argument becomes the primary console and that gets a few extra pieces of output.  Here, I've opted to preserve the full UX of the graphical console, and collect logs on the serial port as best as we can.

For context, this change is part of a larger project to rework the default console settings for each of the supported FCOS/RHCOS platforms.  I think we have a good handle on the general approach; the main thing here is to make sure we've fixed the use case reported in [OCPBUGS-2926](https://issues.redhat.com/browse/OCPBUGS-2926), and it sounds as though we're good on that front too.  Thanks for your feedback! :+1:

bgilbert:
The `scos-9-build-test-qemu` failures are https://github.com/coreos/fedora-coreos-tracker/issues/1325.

PR #1062, opened 2022-11-15
OCPBUGS-3705: [release-4.9] bump fedora-coreos-config

dustymabe:
```
$ git shortlog --no-merges 96217fd..6d912d4
Dusty Mabe (1):
          tests/kola/root-reprovision: add `reprovision` tag
```


mike-nguyen:
/retitle OCPBUGS-3705: [release-4.9] bump fedora-coreos-config

ashcrow:
The `build-test-qemu` is failing on:

> Cannot download Packages/t/tpm2-tss-2.3.2-3.el8.x86_64.rpm ([log reference](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_os/1062/pull-ci-openshift-os-release-4.9-build-test-qemu/1592591016676823040#1:build-log.txt%3A1188))

ashcrow:
Looks like the network issue is over but now we're seeing `semodule: unrecognized option '--rebuild-if-modules-changed'`

dustymabe:
> Looks like the network issue is over but now we're seeing `semodule: unrecognized option '--rebuild-if-modules-changed'`

This feels a bit like one of `rpm-ostree` or `selinux` has been updated in the package set is using an option that was outdated or is too new.

jlebon:
So yes, https://github.com/coreos/rpm-ostree/pull/4122 breaks composing target systems that have old semanage without the switch. We could fix this, but the larger issue here is that it looks like Prow is using the latest cosa to build 4.9.

dustymabe:
> the larger issue here is that it looks like Prow is using the latest cosa to build 4.9.

Can we override that test here then (especially considering we are only changing tags on tests)?

jlebon:
Yeah, the change itself can't really break anything in the old pipeline.

jlebon:
> We could fix this, but the larger issue here is that it looks like Prow is using the latest cosa to build 4.9.

I think https://github.com/openshift/release/pull/34089 fixes this.

PR #1071, opened 2022-11-22
OCPBUGS-3960: Use kernel-rt from ose repo

vfreex:
There is no kernel-rt in 8.6 EUS/TUS.
We temporarily put the latest 8.6 kernel-rt into OSE repo.


travier:
The invalid bug thing here is really weird.

travier:
We need this for master, no for 4.13 which is frozen right now

travier:
/retitle OCPBUGS-3960: Use kernel-rt from ose repo

PR #1074, opened 2022-11-24
scripts: drop `upload-ami`

jlebon:
We don't need this anymore now that this branch has moved to the new pipeline. The pipeline already knows how to share with accounts, trigger follow-up kola tests and make images public.


dustymabe (approved):
/lgtm

cgwalters:
Hooray for deleting code!

PR #951, opened 2022-08-11
ci: Enable more tests for RHCOS 9

travier:
Test PR to have CI runs to figure out broken kola tests.

PR #1079, opened 2022-12-01
Switch to RHEL 9.0 by default

travier:
Done via:
```
$ cosa update-variant default rhel-coreos-9
```

Revert this commit to move RHCOS back to RHEL 8.6 by default.

### DIFF
--- a/extensions.yaml
+++ b/extensions.yaml
@@ -1,1 +1,1 @@
-extensions-rhel-coreos-8.yaml
+extensions-rhel-coreos-9.yaml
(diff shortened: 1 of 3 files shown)


cgwalters:
Let's fold https://github.com/openshift/os/pull/951 into this...I don't think we should make the switch without tests running in Prow.

travier:
We need to set the pipeline to explicitly build 8.6 by default for now before we can merge this one so I'd prefer to keep them separated and hold on this one.

mrunalp:
@knobunc fyi

travier:
@mrunalp Note that while we'll likely merge this at some point, this is not what will switch us to RHCOS 9 in OCP.

travier:
We merged this (for 9.2) in https://github.com/openshift/os/pull/1189

PR #1084, opened 2022-12-07
Revert "image*.yaml: Enable injection of OCP `machine-os` metadata by…

cgwalters:
… default"

This reverts commit f4646dc2d5a83f9101f3b40e3235cebd1ee5c58a.

We're seeing failures here because as of right now the `machine-os-content` and `rhel-coreos-8` are not aliases; and even if they were it appears that `oc` would need work to handle this.

### DIFF
--- a/image-rhel-8.6.yaml
+++ b/image-rhel-8.6.yaml
@@ -8,9 +8,6 @@ squashfs-compression: gzip
 # defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
 ignition-network-kcmdline: []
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Status quo is that we have both old and new container image in the release image.

jlebon:
> Status quo is that we have both old and new container image in the release image.

OK. I asked because I thought the tooling required those labels to be in place but it seems like that's not the case.

/lgtm

cgwalters:
Not sure what failed there, but this can't break it.

travier:
SCOS is failing on something else.

PR #1085, opened 2022-12-07
kola-denylist: skip `rhcos.upgrade.from-ocp-rhcos` for now

jlebon:
There are a few issues with that test we need to fix first. Let's skip it for now to not block on this. We also have upgrade tests in OpenShift CI.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -35,6 +35,8 @@
   tracker: https://bugzilla.redhat.com/show_bug.cgi?id=2080063
   arches:
     - s390x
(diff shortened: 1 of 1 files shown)


jlebon:
We'll need to backport this all the way back to 4.9 I think.

jlebon:
Let's hold this for now. We have a [hacky short-term workaround](https://github.com/coreos/coreos-ci-lib/pull/136) for it and we're hoping to fix the test soon so that we don't need to do all the paperwork to denylist and then later un-denylist it.

/hold

jlebon:
I think we're not going to need this.

PR #1086, opened 2022-12-07
kola-denylist: skip rhcos.upgrade.from-ocp-rhcos for now

mike-nguyen:
There are a few issues with that test we need to fix first. Let's skip it for now to not block on this. We also have upgrade tests in OpenShift CI.


cgwalters (approved):

jlebon:
I'm thinking now we just keep the coreos-ci-lib hack until we fix the test so we never have to actually merge this in all the branches with all the right stamps. WDYT @mike-nguyen?

mike-nguyen:
Sounds good to me!

PR #1087, opened 2022-12-07
tests: Add a test that does a kernel and kernel-rt override

cgwalters:
Motived by seeing a recent bug come through in this area
https://issues.redhat.com/browse/OCPBUGS-4580

(Of course, layering is going to *greatly* improve this situation...
 when we at least go to installing kernel-rt as part of a build process)


travier:
\lgtm
```
--- PASS: ext.config.rpm-ostree.replace-rt-kernel (614.51s)
```

PR #1092, opened 2022-12-12
[rhcos-4.8/s390x] Disable luks.tang test

jschintag:
The new cluster is running nested virtualization, making it to slow to properly run this test.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -17,3 +17,8 @@
    - s390x
 - pattern: coreos.ignition.journald-log
   tracker: https://github.com/coreos/coreos-assembler/issues/1173
+# New s390x cluster is to slow to run this test
+- pattern: luks.tang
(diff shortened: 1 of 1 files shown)


travier:
@mrunalp @aravindhp Hey folks. We need a valid bug label override for this PR. It's to skip a test in our pipeline for a specific arch. This is needed as we moved the pipeline to a new cluster and the test is failing there.

aravindhp:
@travier please get this PR approved. Also, are we losing coverage by this test not running? PS: from your commit message it seems like the test is running slowly and not failing.

mike-nguyen:
This test passes but due to our limitations in the infrastructure for 4.8 it fails.

mike-nguyen:
/assign @cgwalters

for approval

cgwalters:
/label bugzilla/valid-bug
I don't think we need a bug for this.

PR #1095, opened 2022-12-12
kola-denylist: Add a few ext.config.shared.networking

cgwalters:
These are all root caused to injecting networking-related kargs triggers a selinux denial in `systemd-network-generator`:

```
[    4.171674] audit: type=1400 audit(1670858949.421:4): avc:  denied  { add_name } for  pid=1214 comm="systemd-network" name="90-eth1.network" scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:net_conf_t:s0 tclass=dir permissive=0
```

This is fixed in c9s, and I think the fix is
https://github.com/fedora-selinux/selinux-policy/commit/c86d943538f907c2e6b20ffda0a8d2b5b5bd2e34

I don't think the denial here is fatal though for our use case, it's only about writing networkd files which we don't use.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -62,3 +62,20 @@
   osversion:
     - c9s
   snooze: 2022-11-15
+
+- pattern: ext.config.shared.networking.force-persist-ip
+  tracker: https://github.com/fedora-selinux/selinux-policy/commit/c86d943538f907c2e6b20ffda0a8d2b5b5bd2e34
+  osversion:
+    - rhel-9.0
+- pattern: ext.config.shared.networking.mtu-on-bond-kargs
+  tracker: https://github.com/fedora-selinux/selinux-policy/commit/c86d943538f907c2e6b20ffda0a8d2b5b5bd2e34
+  osversion:
+    - rhel-9.0
+- pattern: ext.config.shared.networking.no-persist-ip
+  tracker: https://github.com/fedora-selinux/selinux-policy/commit/c86d943538f907c2e6b20ffda0a8d2b5b5bd2e34
+  osversion:
+    - rhel-9.0
+- pattern: ext.config.shared.networking.prefer-ignition-networking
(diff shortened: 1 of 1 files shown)


kola-denylist.yaml
```diff
@@ -62,3 +62,20 @@
   osversion:
     - c9s
   snooze: 2022-11-15
+
+- pattern: ext.config.shared.networking.force-persist-ip
+  tracker: https://github.com/fedora-selinux/selinux-policy/commit/c86d943538f907c2e6b20ffda0a8d2b5b5bd2e34
+  osversion:
+    - rhel-9.0
```
jlebon, line 69: Let's snooze them until e.g. mid-January? Snoozing isn't the best UX, but it's a crude way to ensure we don't keep disabling those tests well past when they're no longer needed.

jlebon (commented):
OK right, and the reason why we can't disable the service entirely is because we may still need parts of it (https://github.com/coreos/fedora-coreos-tracker/issues/1059#issuecomment-1005066095).

kola-denylist.yaml
cgwalters: I don't think the tests will ever pass in rhel 9.0.  Note the dot-zero.

jlebon (approved):

kola-denylist.yaml
jlebon: In that case, I think ideally these tests should be skipped on RHEL 9.0 entirely. We have `distros`, but I don't think we can limit per version yet, but it could make sense to add.

cgwalters: Agree, though in practice what I think will happen is that we branch and main/master retargets for 9.2 at some point where this is already fixed.

Another way to say this is - the tests should automatically be un-denied when we retarget, which is the main goal.

cgwalters:
(This needs a `/lgtm`)

Issue #1096, opened 2022-12-13
[4.13-9.2] ext.config.shared.networking failures


marmijo:
`ext.config.shared.networking.nameserver` and `ext.config.shared.networking.bridge-static-via-kargs` are failing in rhel-9.0 builds. We need  to investigate the cause of these failures.

xref: https://github.com/openshift/os/pull/1095 & https://github.com/openshift/os/pull/1097

Adam0Brien:
both tests passed, cant reproduce failure.

PR #1097, opened 2022-12-13
kola-denylist: Add more ext.config.shared.networking tests

marmijo:
Add two ext.config.shared.networking tests that are failing in rhel-9.0 until we can request a backport to fix the failures.

See: https://github.com/openshift/os/issues/1096 and https://github.com/openshift/os/pull/1095


travier (commented):
/lgtm

cgwalters:
Ah, sorry I missed these two originally.

cgwalters:
No need to wait for a full CI cycle here

Issue #942, opened 2022-08-05
RHCOS9: ext.config.shared.files.dracut-executable


travier:
```
cat ext.config.shared.files.dracut-executable.txt
Aug 05 10:21:01 qemu0 kola-runext-dracut-executable[4618]: + . /var/opt/kola/extdata-20/commonlib.sh
Aug 05 10:21:01 qemu0 kola-runext-dracut-executable[4619]: ++ mktemp -d
Aug 05 10:21:01 qemu0 kola-runext-dracut-executable[4618]: + tmpd=/tmp/tmp.e0cru8ZTKt
Aug 05 10:21:01 qemu0 kola-runext-dracut-executable[4620]: + cd /tmp/tmp.e0cru8ZTKt
Aug 05 10:21:01 qemu0 kola-runext-dracut-executable[4620]: + lsinitrd --unpack /boot/ostree/rhcos-1652aedc1e413294e3c57cc603b6e904d5eacd98c3f92ea4872513a6f2084279/initramfs-5.14.0-70.17.1.el9_0.x86_64.img
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4636]: + grep -v clevis-luks-common-functions
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4635]: + find /tmp/tmp.e0cru8ZTKt/usr/bin /tmp/tmp.e0cru8ZTKt/usr/sbin /tmp/tmp.e0cru8ZTKt/usr/libexec '!' -perm -0111
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4636]: /tmp/tmp.e0cru8ZTKt/usr/sbin/nvmf-autoconnect.sh
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4618]: + fatal 'Found non-executable scripts in initrd'
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4618]: + echo 'Found non-executable scripts in initrd'
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4618]: Found non-executable scripts in initrd
Aug 05 10:21:04 qemu0 kola-runext-dracut-executable[4618]: + exit 1
```

travier:
We likely need to file a bug for dracut in RHEL 9.0

HuijingHei:
[...]

PR #1101, opened 2022-12-14
kola-denylist: Drop fixed tests from list

travier:
See: https://github.com/openshift/os/issues/942


travier:
To merge after the other ones thus marked as draft to not trigger CI.

PR #1102, opened 2022-12-14
[4.12] manifest-rhel-8.6: Update repo for kernel package

gursewak1997:


cgwalters (approved):

manifest-rhel-8.6.yaml
```diff
@@ -157,7 +157,7 @@ packages:
 # Packages pinned to specific repos in RHCOS
 repo-packages:
   # we always want the kernel from BaseOS
```
cgwalters, line 159: This comment is out of date now, and should probably be changed to reference the reason why we're doing this.

travier:
Don't we want this just for the RT kernel?

ashcrow:
We want it for both vanilla and rt.

travier:
Will need a bug, etc.

ashcrow:
I've overridden all but the `looks good to me` label.

ashcrow:
cc @sosiouxme

manifest-rhel-8.6.yaml
gursewak1997: Should've updated this before re-testing but we will be switching back to BaseOS in the next few weeks.
Not ideal but let's ignore this due to time constraints.

PR #1104, opened 2023-01-03
development-rhcos.md: drop 4.11 and earlier version

HuijingHei:
Remove this as yumrepo work was backported to older branches and
internal redhat-coreos repo is using a single branch for the RHCOS
versions.

### DIFF
--- a/docs/development-rhcos.md
+++ b/docs/development-rhcos.md
@@ -55,11 +55,10 @@ and later. For older versions, see the internal documentation.
   export RHCOS_REPO="..."
   ```
 
-- **For 4.12 and later only:** Clone the config repo (`openshift/os`), passing
-  as argument the internal Git repo which includes the RPM repo configs and
-  optionaly the specific branch. As the Red Hat CA are not included in the cosa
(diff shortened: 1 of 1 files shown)


jlebon (commented):
This is correct, though since the yumrepo work was backported to older branches, the commands marked for `4.12` and later should work equally well for 4.11 and earlier. So while we're here, maybe let's instead drop the "For 4.11 and earlier only" section entirely (and also the "For 4.12 and later only" bolded text higher up). WDYT?

jlebon:
Also, thanks for catching this! Ideally, it would've been updated when we did the backporting work.

HuijingHei:
> This is correct, though since the yumrepo work was backported to older branches, the commands marked for `4.12` and later should work equally well for 4.11 and earlier. So while we're here, maybe let's instead drop the "For 4.11 and earlier only" section entirely (and also the "For 4.12 and later only" bolded text higher up). WDYT?

Sounds good to me, thanks for your suggestion!

jlebon (approved):
Thanks!

PR #1109, opened 2023-01-17
Bug 2133997: manifest: Lock multipath

ravanelli:
- Kola luks.tang is failing due multipath issues, using the current multipath version.
 - We have two versions of multipath in the rhel9 repo, the first version 8.7-7.el9 is causing kola to fail. Let's lock it to 8.7-7.el9_0.1 until a new version is available

### DIFF
--- a/manifest-rhel-9.0.yaml
+++ b/manifest-rhel-9.0.yaml
@@ -114,7 +114,8 @@ packages:
  # We include the generic release package and tweak the os-release info in a
  # post-proces script
  - redhat-release
-
+ # See https://github.com/openshift/os/issues/1110
+ - device-mapper-multipath-0.8.7-7.el9_0.1
 # Packages pinned to specific repos in RHCOS 9
 repo-packages:
   # we always want the kernel from BaseOS


manifest-rhel-9.0.yaml
```diff
@@ -114,7 +114,11 @@ packages:
  # We include the generic release package and tweak the os-release info in a
  # post-proces script
  - redhat-release
-
+ # We have two versions of multipath in the rhel9 repo,
+ # The first version 8.7-7.el9 is causing kola to fail
+ # Let's lock it to 8.7-7.el9_0.1 until a new version is 
+ # available
```
jlebon, line 120: Is there an issue/bug we can paste here to track reverting this pin?

Aside: although RHCOS doesn't have base lockfiles, it should still be possible to have a `manifest-lock.overrides.yaml` file we put this in and have cosa pick it up. I think before we can do that though we'd need to make lockfiles variant-aware somehow, e.g. including it in the lockfile names.

ravanelli: There is an internal issue.  We probably should open one upstream, and other to enable the `manifest-lock.overrides.yaml`

ravanelli: https://github.com/openshift/os/issues/1110

travier:
/lgtm
We probably need this for 4.13 too. I'll skip the bug chain (as we're not shipping 9.0 in 4.12) once we have that sorted out.

travier:
@mrunalp @aravindhp Would appreciate an override here. This does not touch RHCOS as we ship it in 4.12. This is for the 9.0 preview.

mike-nguyen:
Usually the `bugzilla/valid-bug` removes the `bugzilla/invalid-bug` label. Not sure what's going on here.

PR #1103, opened 2022-12-15
Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Adam Piasecki (1):
      denylist: aws.nvme: extend snooze and re-apply to all streams

Benjamin Gilbert (3):
      workflows: update actions to current major versions
      Revert "workflows: add stamp file in openshift/os to ensure we push rebases"
      Consistently Use Title Case for Service Descriptions

Colin Walters (1):
      overlay/35-coreos-live: Lazily unmount /proc/cmdline

Dusty Mabe (17):
      denylist: extend snooze for ext.config.kdump.crash on ppc64le
      denylist: extend snooze for default-network-behavior-change test
      Revert "denylist: snooze timesyncd.dhcp-propagation until January"
      denylist: deny ext.config.ntp.timesyncd.dhcp-propagation on rawhide
      manifests: install the x86_64 emulator on ppc64le instances
      manifests: explicitly name criu criu-libs
      overrides: fast-track container-selinux; add test for container tmpfs mount
      denylist: extend snooze again for platforms.aws.nvme test
      workflows: normalize the denylist in the promotion-diff workflow
      denylist: drop denial from default-network-behavior-change test
      overlay.d: only run coreos-enable-network on first boot
      overlay.d: create new 25-azure-udev-rules overlay
      overlay.d: add change/move to action for azure sriov rules
      coreos-teardown-initramfs: fix spelling error
      coreos-teardown-initramfs: ignore lo.nmconnection
      tests: update comment in bootif net propagation test
      tests: set rd.neednet=1 in bootif net propagation test

Jonathan Lebon (3):
      overrides: drop grub2 pin, bump to 2.06-72.fc37
      ignition-ostree-firstboot-uuid: nuke libblkid cache after UUID restamp
      overlay: always use `-p` with `blkid`

Joseph Marrero (2):
      tests/kola/rpm-ostree: add test for kernel replacement
      kola/rpm-ostree/kernel-replace: update to work on all arches

Michael Armijo (1):
      tests: remove trust-cpu

Nikita Dubrovskii (1):
      overlay.d: remove boot-disk workaround for secex ignition config

gursewak1997 (1):
      kola: Update kola tests to use new YAML format for kola options
```

PR #1111, opened 2023-01-20
common.yaml: include the 25azure-udev-rules overlay from FCOS

dustymabe:
This was originally introduced in https://github.com/coreos/fedora-coreos-config/pull/2176. Let's pick it up here in RHCOS too.


dustymabe:
I had originally added this commit to https://github.com/openshift/os/pull/1103 but the bot that managed that PR kicked it out so here it is in a new PR.

dustymabe:
ok this one should pass CI now.

cgwalters (approved):

PR #1116, opened 2023-01-25
Workaround for missing SIG GPG keys

travier:
Import CentOS Stream 9 + SIGs GPG keys

---

scos: Only pull packages from RHCOS repos by default

Public builds and the OKD SCOS pipeline should use other repos.

---

extensions: Use vendored GPG keys for SCOS builds


travier:
Another option would be to get those keys from the COSA container before the extension build.

travier:
Verified passing locally. CI here does not test that.

travier:
Alternative in https://github.com/coreos/coreos-assembler/pull/3325 & (openshift/os PR TODO)

travier:
Note that technically the main key is already in the image in `/etc/pki/rpm-gpg` and we're missing only the SIG keys but it's simpler to handle all keys the same way.

travier:
/hold
Found another better option

travier:
Closing in favor of https://github.com/openshift/os/pull/1118

PR #1117, opened 2023-01-25
extensions: Use COSA vendored GPG keys for SCOS builds

travier:
scos: Only pull packages from RHCOS repos by default

Public builds and the OKD SCOS pipeline should use other repos.

---

extensions: Use COSA vendored GPG keys for SCOS builds


travier:
Goes with https://github.com/coreos/coreos-assembler/pull/3325

travier:
Closing in favor of https://github.com/openshift/os/pull/1118

PR #1121, opened 2023-01-26
c9s.repo: use external GPG key for extras-common

vrutkovs:
Fixes 
```
Updating metadata for 'extras-common'...done
[0m[31merror: [0mUpdating rpm-md repo 'extras-common': Failed to download gpg key for repo 'extras-common': Status code: 404 for https://www.centos.org/keys/RPM-GPG-KEY-CentOS-SIG-Extras-SHA512 (IP: 81.171.33.201)
error: failed to execute cmd-fetch: exit status 1
```
when building C9S build

### DIFF
--- a/c9s.repo
+++ b/c9s.repo
@@ -20,7 +20,7 @@ baseurl=https://mirror.stream.centos.org/SIGs/9-stream/extras/$basearch/extras-c
 gpgcheck=1
 repo_gpgcheck=0
 enabled=1
-gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-Extras-SHA512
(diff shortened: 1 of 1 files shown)


travier:
We explicitly did that in https://github.com/openshift/os/commit/a3f86316a0bacd5b4f02bb28ca95ecf647a4600a

travier:
In what step is the OKD pipeline failing? This should be part of the COSA image.

vrutkovs:
cosa-build-baseos is failing, which runs `cosa fetch`. It happens before extensions build

travier:
This file is part of the COSA image and has been for a while already. Why is this failing in `cosa fetch`? Is this step running inside a COSA container?

vrutkovs:
Hmm, right, its indeed there. Probably some misconfiguration in my pipeline

PR #1122, opened 2023-01-26
[release-4.12] kola-denylist: 9.0: Skip package downgrade test

travier:
RHCOS 4.12-9.0 will only be a preview so we don't need to be strict on package downgrades.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -80,3 +80,6 @@
   tracker: https://github.com/openshift/os/issues/1096
   osversion:
     - rhel-9.0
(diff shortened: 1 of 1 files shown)


travier:
CI here is not going to test that and this does not change RHCOS 8 as shipped in 4.12.
/label backport-risk-assesed

ravanelli (approved):
/lgtm

travier:
@aravindhp @mrunalp
Hey folks. This is another selected fix for the 4.12-9.0 dev preview. This does not touches RHCOS as we ship it in 4.12.
Would appreciate an override here.
Thanks.

PR #1123, opened 2023-01-26
tests/rhel-major-version: Also ignore shim-aa64

travier:
The shim packages are still from RHEL 8 in C9S. e81936c ignored the x86 package but missed the aarch64 one.

Fixes: e81936c tests: Add EL9 support for version checks


cgwalters:
Just going to merge this, no need to wait for all the CIs

travier:
@cgwalters Be careful when direct merging things here as last time that created issue with branches.

travier:
Looks like this was not propagated to the 4.13 & 4.14 branches.

cgwalters:
Oh ugh...ok.  I'll get someone to remove my permissions to do that :smile:

PR #1124, opened 2023-01-26
Add image-c9s.yaml

LorbusChris:
Adds a dedicated image config for CentOS Stream 9.

`oc adm release new` and the CVO require the `machine-os` label to be present on one of the payload images.
For c9s, we let COSA add the required labels to the ostree container image.

In OKD/SCOS, this image is aliased as both `machine-os-content` and `centos-stream-coreos-9` in the payload.

### DIFF
--- a/image-c9s.yaml
+++ b/image-c9s.yaml
@@ -1,1 +1,20 @@
-image-rhel-9.0.yaml
+# See https://github.com/coreos/coreos-assembler/pull/298
+size: 16
+
+# Disable networking by default on firstboot. We can drop this once cosa stops
+# defaulting to `ip=dhcp,dhcp6 rd.neednet=1` when it doesn't see this key.
+ignition-network-kcmdline: []
+
+# In OKD, keep OpenShift metadata on our container image, but xref https://github.com/openshift/os/issues/1047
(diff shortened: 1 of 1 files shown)


travier:
Have we fixed the issue from https://github.com/openshift/os/pull/1084 ?

cgwalters:
Hmm right no we haven't...this will actually break the current plan to use C9S for OCP temporarily because we haven't aliased `machine-os-content` there.

Ugh so this all starts to pile up onto digging out of the DTK hole i.e. https://github.com/openshift/driver-toolkit/pull/102 and https://github.com/openshift/driver-toolkit/pull/103 and https://issues.redhat.com/browse/MCO-392

cgwalters:
OK well I guess for now we could change the OKD pipeline to do this.

LorbusChris:
Carrying this downstream would be OK for now, too.

travier:
OK, once we have https://github.com/openshift/os/pull/1127, we should be able to merge this one and have SCOS be SCOS again.

LorbusChris:
@travier @cgwalters #1127 has merged, do you think we can unhold and merge this one now?

travier:
LGTM
/unhold

PR #1126, opened 2023-01-27
c9s: Add temporary COPR repo for openvswitch on s390x

travier:


travier:
CI here won't test that change.

travier:
Arg, it's failing as we only have that for s390x 🙁

jlebon:
I think you should be able to work around the arch issue by putting the repo addition in a separate manifest, and including it using an `arch-include` on s390x only.

Can you include a link to the email thread where this COPR repo is referenced by a CentOS NFV SIG team member in the commit message and also as a comment in the yum repo file?

travier:
Closing in favor of https://github.com/openshift/os/pull/1127.

PR #1127, opened 2023-01-27
Add RHEL 9.2 variant

travier:
c9s: Restore okd-copr repo by default for c9s

---

EL9: Move C9S overlay to the EL9 shared manifest

Will be used by the RHEL 9.2 variant.

---

Add RHEL 9.2 variant and set it as default for rhel-coreos-9

---

9.2: Add temporary COPR repo for openvswitch on s390x

See: https://lists.centos.org/pipermail/centos-devel/2023-January/142752.html


travier:
Will need `openshift/release` changes for CI.

travier:
See: https://github.com/openshift/release/pull/35789

manifest-rhel-9.2.yaml
```diff
@@ -0,0 +1,157 @@
...
+  - rhel-9.0-server-ose-4.13
+  # Uncomment and switch to the following once ready
+  # - rhel-9.2-baseos
+  # - rhel-9.2-appstream
+  # - rhel-9.2-fast-datapath
+  # - rhel-9.2-server-ose-4.13
+
+arch-include:
+  s390x:
+    # Temporary repo for openvswitch on s390x
```
cgwalters, line 36: I think only shipping a particular package on a particular architecture is going to trip up ART's consistentency checks; we should use the same content across all arches.

manifest-rhel-9.2.yaml
```diff
@@ -0,0 +1,157 @@
...
+automatic-version-suffix: "-"
+# Keep this is sync with the version in postprocess
+mutate-os-release: "4.13"
+
+postprocess:
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+
+     # Tweak /usr/lib/os-release
```
cgwalters, line 52: (Not a new problem but we should factor this out into a common include)

cgwalters (approved):

travier:
Pushed an update that should build this 9.2 variant via the 9.0 Prow jobs so that we can test this in CI.

cgwalters:
```
 16:48:35 qemu0 kola-runext-rootless-systemd[1956]: Errors during downloading metadata for repository 'fedora':
Jan 30 16:48:35 qemu0 kola-runext-rootless-systemd[1956]:   - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-37&arch=x86_64&countme=1 [getaddrinfo() thread failed to start]
Jan 30 16:48:35 qemu0 kola-runext-rootless-systemd[1956]:   - Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-37&arch=x86_64 [getaddrinfo() thread failed to start]
Jan 30 16:48:35 qemu0 kola-runext-rootless-systemd[1956]: Error: Failed to download metadata for repo 'fedora': Cannot prepare internal mirrorlist: Curl error (6): Couldn't resolve host name for https://mirrors.fedoraproject.org/metalink?repo=fedora-37&arch=x86_64 [getaddrinfo() thread failed to start]
Jan 30 16:48:36 qemu0 kola-runext-rootless-systemd[1904]: Error: error building at STEP "RUN dnf -y update && dnf -y install systemd httpd && dnf clean all && systemctl enable httpd": error while running runtime: exit status 1
```

Hmmm...that's usually from having an old container runtime that's still giving EPERM for `clone()`...  It looks like things are a bit of a mess here, we have
`  containers-common-2:1-47.el9.x86_64 (appstream)`
yet `  podman-3:4.2.0-4.rhaos4.12.el9.x86_64 (rhel-9.0-server-ose-4.13)`?

cgwalters:
Ohhh I see, we're just hitting https://bugzilla.redhat.com/show_bug.cgi?id=2123246 again now that it's missing from the denylist...

cgwalters:
I'm piking up the ball :soccer: on this PR for the rest of the day.

jlebon:
xref https://github.com/openshift/os/pull/1129#issuecomment-1409276434

cgwalters:
OK, pushed a few test ignores/fixes.

cgwalters:
Argh, same failures...but I'm not reproducing this locally?  :cry:

cgwalters:
OK, finally figured out that the kola logs for the nonexclusive tests are in https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_os/1127/pull-ci-openshift-os-master-rhcos-90-build-test-qemu/1620159375001784320/artifacts/test/artifacts/kola/non-exclusive-test-bucket-0/4a597719-8ee2-4665-ac53-0eb383eefa04/ which helped.

I think I'm getting some yum repository config drift versus CI though, will try to dig into that tomorrow.

cgwalters:
OK we have green CI with some ignored tests.  I think we should proceed in order to find the more important bugs, so adding `/lgtm` (insert obligatory [Obama awarding self medal meme](https://knowyourmeme.com/memes/obama-awards-obama-a-medal) with repect to my additional commits).  But also `/hold` - @travier if you agree with this please feel free to lift the hold and let's move onto the next things to do here!
/lgtm
/hold

travier:
LGTM
/unhold

PR #1129, opened 2023-01-30
Allow kernel to be source from the ART plashet

jupierce:
OpenShift and RHEL have come to an agreement whereby OCP will be able to release the RHEL kernel before the normal 6 week cadence of RHEL.
https://issues.redhat.com/browse/RHELPLAN-144928 tracks this agreement.
https://issues.redhat.com/browse/ART-5663 discussed ART/RHCOS agreement on how this would be handled at the pipeline level. The decision was to remove the constraint requiring the kernel to be source from baseos exclusively. By doing so, it ART has a more recent kernel build in its repository, it will be preferred.

### DIFF
--- a/manifest-rhel-8.6.yaml
+++ b/manifest-rhel-8.6.yaml
@@ -156,10 +156,6 @@ packages:
 
 # Packages pinned to specific repos in RHCOS
 repo-packages:
-  # we always want the kernel from BaseOS
-  - repo: rhel-8.6-baseos
-    packages:
-      - kernel
   # we want the one shipping in RHEL, not the equivalently versioned one in RHAOS
   - repo: rhel-8.6-appstream
     packages:


jlebon:
I think we need to also update `manifest-rhel-9.0.yaml` and `manifest-c9s.yaml` to match. If this merges first, we'll have to make sure to update https://github.com/openshift/os/pull/1127 too before merging.

cgwalters:
Offhand I'm thinking this should merge into 4.12 directly; for 4.13 we're aiming at RHEL9 and I don't think this kernel flow has been looked at in that context at all, right?  We're just going to ship the C9S kernel for now AFAIK, right?

travier:
Agree with Colin that this should go to 4.12 directly has this does not make much sense for 4.13 right now.

PR #1131, opened 2023-01-30
Allow kernel to be sourced from the ART plashet

jupierce:
OpenShift and RHEL have come to an agreement whereby OCP will be able to release the RHEL kernel before the normal 6 week cadence of RHEL. https://issues.redhat.com/browse/RHELPLAN-144928 tracks this agreement. https://issues.redhat.com/browse/ART-5663 discussed ART/RHCOS agreement on how this would be handled at the pipeline level. The decision was to remove the constraint requiring the kernel to be source from baseos exclusively. By doing so, it ART has a more recent kernel build in its repository, it will be preferred.

### DIFF
--- a/manifest-rhel-8.6.yaml
+++ b/manifest-rhel-8.6.yaml
@@ -156,10 +156,6 @@ packages:
 
 # Packages pinned to specific repos in RHCOS
 repo-packages:
-  # we always want the kernel from BaseOS
-  - repo: rhel-8.6-baseos
-    packages:
-      - kernel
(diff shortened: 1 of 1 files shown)


jmarrero:
Not sure of the history of why we added 
```
# we always want the kernel from BaseOS
```
Looks like @travier added that comment. Deferring to him.

ashcrow:
Do we have an associated bug or will this require an override?

travier:
/lgtm
/label backport-risk-assessed
/label cherry-pick-approved
I'm good with an override and no bug.

PR #1132, opened 2023-01-31
Bump fedora-coreos-config

coreosbot-releng:

```
Dusty Mabe (1):
      tests/kola/ntp: use different server for NTP for DHCP propagation tests

Jonathan Lebon (3):
      coreos-livepxe-rootfs: print initramfs version on error
      coreos-livepxe-rootfs: move TLS comment to above `--insecure`
      coreos-livepxe-rootfs: use `curl --fail`

Timothée Ravier (1):
      tests: Enable rpmdb-sqlite for EL9

gursewak1997 (1):
      kola/setuid: Update known files with SetUID bit set See https://github.com/coreos/fedora-coreos-tracker/issues/1391
```


mike-nguyen (approved):
/lgtm

PR #1136, opened 2023-01-31
rhel-9.2: Drop sig-nfv on s390x

cgwalters:
The previous PR to try to have this build on s390x wasn't tested manually and CI doesn't cover it, so predictably it didn't work. The reason is because the `sig-nfv` repo doesn't have content there.

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -22,7 +22,6 @@ repos:
   - appstream
   # For NFV & Virtualization SIG GPG keys
   - extras-common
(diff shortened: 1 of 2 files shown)


cgwalters:
CI here will only cover the "it still builds on x86_64" aspect
/skip

jlebon (approved):

jlebon:
Neat, I had forgotten that `arch-include` can take a string or a list of strings. Assuming you sanity-checked the x86_64 case locally still works, I'd say let's just merge this and not wait for CI.

PR #1138, opened 2023-02-01
[release-4.12] kola-denylist: Deny upgrade tests

ravanelli:
- The verify-no-pkg-downgrades test is a subtest, we can't deny a subtest only, need to deny all upgrade tests instead.
 - RHCOS 4.12-9.0 will only be a preview so we don't need to be strict on it.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -80,6 +80,7 @@
   tracker: https://github.com/openshift/os/issues/1096
   osversion:
     - rhel-9.0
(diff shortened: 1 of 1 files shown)


ravanelli:
Same as https://github.com/openshift/os/pull/1122

/label backport-risk-assessed

ravanelli:
@aravindhp @mrunalp
It is a fix for https://github.com/openshift/os/pull/1122. Not BZ needed. Can you add the BZ override here? Please.

PR #1140, opened 2023-02-01
9.2: Switch to using openvswitch from 9.0

cgwalters:
The COPR thing breaks our lockfile handling, which wants to find the same NEVRAs across architectures.

Further, trying to ship a COPR repo just gives me the willies; there's absolutely no required reviews for changes or CI etc.

If we need fixes post-9.0 for openvswitch, then we'll do something custom.


manifest-rhel-9.2.yaml
```diff
@@ -150,3 +142,6 @@ repo-packages:
       - cri-tools
       - openshift-clients
       - openshift-hyperkube
+  - repo: rhel-9.0-fast-datapath
```
jlebon, line 145: Aside: I've been thinking for a while we should support specifying repos here that aren't listed in the toplevel `repos` key. The semantic would be that we enable this repo only for the specified packages and nothing else.

jlebon (approved):
/approve
/lgtm

cgwalters:
> error: Unknown rpm-md repository: rhel-9.0-fast-datapath

Hmm digging

cgwalters:
was a flake but anyways doesn't matter since rhel8 will be gone after this PR

PR #1142, opened 2023-02-02
tests: Unify format and update kola config to YAML

travier:
tests: Use KOLA_EXT_DATA/commonlib.sh

Use common shared bash library and functions instead of copy/pasted
definitions.

---

tests: Unify layout with YAML'ed kola config

Unify test layout and format and use YAML based config for kola.


jlebon (approved):
Nice!
/approve
/lgtm

PR #1144, opened 2023-02-02
rhel-9.2: Add `centos-release-virt-common` on !s390x

cgwalters:
In the previous change, I re-broke the s390x virt-common issue. I had actually (tried to) test things locally, but right now we don't correctly cache-bust `cosa build-extensions-container` when the base image is modified because cosa doesn't really understand containers.

To recap, two problems intersect here:

- In RHEL, there is only one production release GPG key, but in CentOS for some reason there are multiple, one for base OS and one for SIGs etc.  And those SIG keys are in separate packages.
- Compounding the above, the Virt SIG isn't built for s390x, so we have to make this architecture conditional

### DIFF
--- a/centos-virt.yaml
+++ b/centos-virt.yaml
@@ -1,0 +1,7 @@
+# In RHEL, there is only one production key.
+# In CentOS, we need a separate package for the separate keys.
+# Now, we can't just include this *unconditionally* since it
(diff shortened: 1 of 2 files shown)


cgwalters:
These two tests don't matter

PR #1145, opened 2023-02-02
Update development-rhcos.md

aaradhak:
In the development-rhcos.md , the cosa init command has the flag --yumrepo,  looks like it should be _--yumrepos_ instead of --_yumrepo_

`cosa init --yumrepo "${RHCOS_REPO}" https://github.com/openshift/os.git getopt: option '--yumrepo' is ambiguous; possibilities: '--yumrepos' '--yumrepos-branch'`


jlebon:
Good catch, thanks!
/approve
/lgtm

PR #1146, opened 2023-02-02
Bump to OVS 3.1 on RHEL9

dcbw:
Container side bumps to 3.1 in https://github.com/openshift/ovn-kubernetes/pull/1544 and we should to keep the host OS side in sync.

### DIFF
--- a/common-el9.yaml
+++ b/common-el9.yaml
@@ -49,4 +49,4 @@ postprocess:
 
 # Packages that are only for SCOS & RHCOS 9
 packages:
- - openvswitch2.17
+ - openvswitch3.1
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -154,6 +154,3 @@ repo-packages:
       - cri-tools
       - openshift-clients
       - openshift-hyperkube
-  - repo: rhel-9.0-fast-datapath
-    packages:
-      - openvswitch2.17


jlebon:
One thing to note is that for our RHCOS 9.2 builds, we're currently pulling openvswitch from 9.0 because C9S doesn't have s390x (see https://github.com/openshift/os/pull/1140). We'll need to make sure to have s390x builds of 3.0 somehow before we can do this.

dcbw:
@jlebon will this work for packages that are cross-tagged into OCP 4.13, that aren't yet in FDP repos? OVS 3.1 is going to get cross-tagged this week, and that's what we'd really like to use here.

dcbw:
Tested with all 3 workers updated to OVS 3.1 and rebooted. Ran openshift-tests conformance:

1/585/585  -- The single failure was because my local kubectl is old, not the cluster's fault.
16/2345/2345 -- Failures are storage tests caused by EBS volumes not found
86/455/455 -- Failures because `panic: mkdir /tmp/fixture-testdata-dir4125216929: no space left on device` from my local machine

In short, I don't see any blockers to OVS 3.1 in RHCOS9 so far. I'd like to wait until we've cross-tagged OVS 3.1 and then I'll update this PR and remove the WIP, then I think we're good to go.

dcbw:
(going to be a couple hours before AOS mirrors sync the tagged RPMS...)

jlebon:
> @jlebon will this work for packages that are cross-tagged into OCP 4.13, that aren't yet in FDP repos? OVS 3.1 is going to get cross-tagged this week, and that's what we'd really like to use here.

Yes, that should be fine.

Can you also drop the bits that were added to `manifest-rhel-9.2.yaml` as part of https://github.com/openshift/os/pull/1140?

dcbw:
@jlebon we *do* want to pull from Fast Datapath repos though; this cross-tag is just because OVS 3.1 isn't in a released Errata yet, but will be in the next couple weeks. That's too long to wait for Devel Freeze though, so cross-tag for now.

If we want to keep the current (at least in <= 4.12) behavior of always preferring Fast Data Path, do I need to revert those hunks?

jlebon:
Hmm, so currently 4.12 and earlier don't use repo pinning. Whatever the latest openvswitch2.17 found is what gets pulled in. I guess there we're just relying on the fact that it only lives in the FDP repo anyway IIUC.

If we want to strengthen that, SGTM. Temporarily for 4.13, you can update the repo pin in `manifest-rhel-9.2.yaml` to point to `rhel-9.0-server-ose-4.13`. Once 9.2 repos are available, we'll [uncomment these lines](https://github.com/openshift/os/blob/f277f3da87a73eed9f2bfd872dd3e0e72a5b586b/manifest-rhel-9.2.yaml#L29-L33) and can update the pin to pull from the new 9.2 FDP repo instead.

dcbw:
> Hmm, so currently 4.12 and earlier don't use repo pinning. Whatever the latest openvswitch2.17 found is what gets pulled in. I guess there we're just relying on the fact that it only lives in the FDP repo anyway IIUC.

@jlebon that's actually what we want on the data plane side; we want latest FDP to get pulled in so that we get automatic CVE fixes. We only want very short-term states of cross-tagging specific packages before they get errata-ed in FDP, at which point FDP should take over.

> If we want to strengthen that, SGTM. Temporarily for 4.13, you can update the repo pin in `manifest-rhel-9.2.yaml` to point to `rhel-9.0-server-ose-4.13`. Once 9.2 repos are available, we'll [uncomment these lines](https://github.com/openshift/os/blob/f277f3da87a73eed9f2bfd872dd3e0e72a5b586b/manifest-rhel-9.2.yaml#L29-L33) and can update the pin to pull from the new 9.2 FDP repo instead.

Our ideal case for RHCOS/data plane is the current 4.12 behavior. We just have to jump ahead a bit via the cross-tag because of the devel freeze deadlines; but when the 3.1 builds show up in FDP we'll happily use those.

Can we just select OVS from *either* of server-ose or FDP, whichever is newest? Does that mean just removing anything referencing openvswitch3.1 from `manifest-rhel-9.2.yaml` entirely?

dcbw:
@jlebon eg, could we do something like this:

```
  - repo: rhel-9.0-server-ose-4.13
    packages:
      - conmon-rs
      - cri-o
      - cri-tools
      - openshift-clients
      - openshift-hyperkube
      # Don't care which of [FDP, server-ose] we get it from; pick best available
      - openvswitch3.1
  - repo: rhel-9.0-fast-datapath
    packages:
      # Don't care which of [FDP, server-ose] we get it from; pick best available
      - openvswitch3.1
```

So that it gets picked up from either one, whichever is has the newer RPM?

travier:
Given:
> Can we just select OVS from either of server-ose or FDP, whichever is newest?

This LGTM. Holding to make sure we all agree.
/lgtm
/hold

dcbw:
@travier I'm OK with the change as-is, I think that'll do what we want for now. If we need to switch things up later I'll do another PR down the line.

jlebon:
> @jlebon eg, could we do something like this:
> 
> ```
>   - repo: rhel-9.0-server-ose-4.13
>     packages:
>       - conmon-rs
>       - cri-o
>       - cri-tools
>       - openshift-clients
>       - openshift-hyperkube
>       # Don't care which of [FDP, server-ose] we get it from; pick best available
>       - openvswitch3.1
>   - repo: rhel-9.0-fast-datapath
>     packages:
>       # Don't care which of [FDP, server-ose] we get it from; pick best available
>       - openvswitch3.1
> ```
> 
> So that it gets picked up from either one, whichever is has the newer RPM?

I'd have to double check but this probably won't work as expected. I think eventually we should support something like:

```yaml
- repos:
    - repo1
    - repo2
  packages:
    - foobar
```

But anyway, if you want to just have the default behaviour of "pick whatever is the latest in all enabled repos", then what this PR is doing is correct.

/lgtm
/approve

PR #1147, opened 2023-02-02
OCPBUGS-4762: [release-4.10] manifest: bump to openvswitch2.17

dcbw:
We need to consolidate OVS streams and move off OVS 2.16. Bump host OS OVS to 2.17 which has a good track record the past 10 months in 4.11+.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -223,7 +223,7 @@ packages:
  - cri-o cri-tools
  # Networking
  - nfs-utils
- - openvswitch2.16
+ - openvswitch2.17
  - dnsmasq
  - NetworkManager-ovs
  # Extra runtime


cgwalters:
This is already happening as part of us switching to 9.2 for 4.13, so I think we can close this?

cgwalters:
See https://issues.redhat.com/browse/COS-1926 for the latest, but specifically here https://github.com/openshift/os/blob/059bf420452b2b246255a869ce70f0bb11196a5f/manifest-rhel-9.2.yaml#L25

(Incidentally...having openvswitch on the host has recently caused us a not-small amount of pain...moving it into a container would have significantly helped recent rhel9 work)

cgwalters:
Oh I see, you want to get this all the way back to 4.10?  Then I guess it's indeed easiest to land in master now.
/approve

Note though this should really be tested somewhat manually pre-merge directly in a modified 4.10 build I think.

cgwalters:
And now that I look sorry this is already directly against release-4.10...

From your PoV is this tested?

travier:
/label cherry-pick-approved
/label backport-risk-assessed
Waiting on confirmation before we LGTM this one.

sdodson:
@dcbw I think they're just waiting on your confirmation that this has been tested in 4.10.

flavio-fernandes (commented):
/lgtm

PR #1162, opened 2023-02-13
Add `hotfixes-rhel-coreos-9.yaml`

cgwalters:
This builds on https://github.com/coreos/coreos-assembler/pull/3352 to add the latest build of RHEL8 rpm-ostree inside our extensions container so the MCD can live-apply it.

Part of https://issues.redhat.com/browse/OCPBUGS-7275

### DIFF
--- a/hotfixes-rhel-coreos-9.yaml
+++ b/hotfixes-rhel-coreos-9.yaml
@@ -1,0 +1,5 @@
+hotfixes:
+  - osmajor: "8"
+    repo: rhel-8.6-appstream
+    packages:
+      - rpm-ostree # For https://issues.redhat.com/browse/OCPBUGS-7275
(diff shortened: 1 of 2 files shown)


hotfixes-rhel-coreos-9.yaml
```diff
@@ -0,0 +1,5 @@
+hotfixes:
+  - osmajor: "8"
+    repo: rhel-8.6-appstream
```
travier, line 3: 8.6 here but you mention 8.8? Isn't this RPM more likely to be a custom rhaos build instead?

cgwalters: Yeah, this is just a demonstration change.  That said, I do think it'd likely make sense to get the change directly into 8.6.z too.

cgwalters:
OK as of right now, we changed plan and are shipping the fixed rpm-ostree back to 4.12.  But, at least we have this technology now in case we need it in the future.

PR #1167, opened 2023-02-17
rhel-9.2: Disable centos repos by default

cgwalters:
We need to work in disconnected environments by default, and default-enabled repos will be attempted to be fetched by rpm-ostree when doing node-local kernel overrides today for e.g. kernel-rt.

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -116,6 +116,13 @@ postprocess:
 
      ---
      EOF
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+     # We need to work in disconnected environments by default, and default-enabled
+     # repos will be attempted to be fetched by rpm-ostree when doing node-local
+     # kernel overrides today for e.g. kernel-rt.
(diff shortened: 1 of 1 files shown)


manifest-rhel-9.2.yaml
```diff
@@ -116,6 +116,13 @@ postprocess:
 
      ---
      EOF
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+     # We need to work in disconnected environments by default, and default-enabled
+     # repos will be attempted to be fetched by rpm-ostree when doing node-local
+     # kernel overrides today for e.g. kernel-rt.
+     for x in /etc/yum.repos.d/*.repo; do sed -i -e s,enabled=1,enabled=0, $x; done
```
travier, line 125:
```suggestion
     for x in /etc/yum.repos.d/*.repo; do sed -i -e 's,enabled=1,enabled=0,' $x; done
```

travier: likely not worth a respin

travier (commented):
/lgtm

travier (approved):

PR #1172, opened 2023-02-21
ci/prow-entrypoint: simplify `kola testiso` invocation

jlebon:
We've reworked `kola testiso` so that things like BIOS, UEFI, Secure Boot, multipath, metal4k variants are tested by default when you just call `kola testiso`:

https://github.com/coreos/coreos-assembler/pull/3298

This will increase the number of tests this code currently runs. We can always trim it back down with `--denylist-test` in the future if wanted.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -134,19 +134,7 @@ kola_test_metal() {
     cosa compress --artifact=metal --artifact=metal4k
 
     # Run all testiso scenarios on metal artifact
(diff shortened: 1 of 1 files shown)


jlebon:
To be merged and backported as part of https://github.com/coreos/coreos-assembler/pull/3298.

HuijingHei:
This will fix current CI issue.
```
+ kola testiso -S --scenarios pxe-install,pxe-offline-install,iso-install,iso-offline-install,iso-live-login,iso-as-disk,miniso-install --output-dir tmp/kola-metal
Error: unknown flag: --scenarios
```
/lgtm

HuijingHei (approved):
LGTM

PR #1174, opened 2023-02-23
Bump fedora-coreos-config

coreosbot-releng:

```
Benjamin Gilbert (1):
      denylist: snooze another test that uses Butane 1.5.0-experimental

Huijing Hei (1):
      ignition-ostree-transposefs: add workaround when removing the zram device

Jonathan Lebon (1):
      workflow: don't remove graduated overrides on next-devel when disabled
```


travier:
This includes https://github.com/coreos/fedora-coreos-config/pull/2226 for https://issues.redhat.com/browse/COS-2038

PR #1176, opened 2023-02-24
Bump fedora-coreos-config

coreosbot-releng:

```
Benjamin Gilbert (1):
      denylist: extend snooze on Butane 1.5.0-experimental tests

Colin Walters (1):
      overlay/live-generator: Allow 50% of RAM for /etc and /var

Dusty Mabe (2):
      denylist: deny kdump.crash test on aarch64 for f38 streams
      overlay.d/35coreos-live: pull in coreos-enable-network.service on live boots
```


dustymabe (approved):
/lgtm

PR #1165, opened 2023-02-15
Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Benjamin Gilbert (1):
      denylist: snooze tests that use Butane 1.5.0-experimental

Dusty Mabe (1):
      denylist: snooze files.file-directory-permissions test

Huijing Hei (1):
      network: fix `default-network-behavior-change` script

Nikita Dubrovskii (4):
      s390x: secex: remove the Ignition config before entering the emergency-shell
      s390x: secex: disable login on serial console
      s390x: secex: disable Ignition's logs
      s390x: secex: decrypt ignition config on firstboot

Shilpi Das (1):
      denylist: Changing tracker issues for coreos.boot-mirror and coreos.boot-mirror.luks

Timothée Ravier (2):
      denylist: Snooze ext.config.toolbox until 2023-02-21 for rawhide
      Revert "denylist: Snooze ext.config.toolbox until 2023-02-21 for rawhide"

gursewak1997 (1):
      denylist: snooze ext.config.extensions.module for now Issue: coreos/fedora-coreos-tracker#1420
```

PR #1177, opened 2023-02-27
denylists: Re-enable tests

HuijingHei:
denylists: re-enable `ext.config.shared.networking.default-network-behavior-change`

- Sync in https://github.com/openshift/os/pull/1165

---

denylists: re-enable `luks.*` on `!x86_64`
- Sync in https://github.com/openshift/os/pull/1174

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -51,10 +51,6 @@
 - pattern: ext.config.version.rhel-matches-rhcos-build
   tracker: ''
 
-# This test is broken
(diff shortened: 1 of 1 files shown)


cgwalters:
So this didn't work...on ppc64le we still got
```
17:34:08  --- FAIL: luks.sss.t1 (179.72s)
17:34:08          luks.go:171: Unable to create test machine: machine fc88394d-34d8-403e-bfbd-0d186a92946f entered emergency.target in initramfs
```

PR #1178, opened 2023-02-27
c9s: Switch to pulling directly from composes.stream.centos.org

cgwalters:
There's some lag in the mirror network apparently, and we want the newer kernel.


c4rt0 (approved):
/lgtm

cgwalters:
We're still in a situation where we're double-CI testing so
/override ci/prow/rhcos-90-build-test-qemu
/override ci/prow/rhcos-90-build-test-metal

PR #1181, opened 2023-02-28
c9s: Get conmon-rs from rhcontainerbot COPR

LorbusChris:
Until there are official C9S appstream or SIG builds of conmon-rs available, get it from the integration builds maintained by upstream. This ensures a current version of conmon-rs is consumed during builds outside of Prow, i.e. OKD/SCOS release builds.


cgwalters:
Seems sane
/approve
/lgtm

Issue #1187, opened 2023-03-09
c9s: ext.config.shared.ignition.resource.remote started failing


cgwalters:
> 08:16:04    ignition-2.15.0-1.el9.x86_64 (appstream)
> ...
> 08:30:39  --- FAIL: ext.config.shared.ignition.resource.remote (6.55s)
> 08:30:39          harness.go:1642: mach.Start() failed: machine 92acf525-eb20-4023-8f3a-8a61c2880dbf entered emergency.target in initramfs

```
[    3.058090] ignition[693]: failed to fetch config: unsupported config version
[    3.058654] ignition[693]: failed to acquire config: unsupported config version
[    3.059189] ignition[693]: Ignition failed: unsupported config version
[[0;1;31mFAILED[0m] Failed to start [0;1;39mIgnition (fetch-offline)[0m.
See 'systemctl status ignition-fetch-offline.service' for details.
[[0;1;38;5;185mDEPEND[0m] Dependency failed for [0;1;39mIgnition Complete[0m.
[[0;1;38;5;185mDEPEND[0m] Dependency failed for [0;1;39mInitrd Default Target[0m.
[[0;32m  OK  [0m] Stopped target [0;1;39mInitrd /usr File System[0m.[    3.064433] systemd[1]: ignition-fetch-offline.service: Main process exited, code=exited, status=1/FAILURE
[    3.065106] systemd[1]: ignition-fetch-offline.service: Failed with result 'exit-code'.
```

cgwalters:
Seems this is a regression, our previous run used `ignition-2.14.0-5.rhaos4.12.el9.x86_64 `

travier:
Looks like https://github.com/coreos/fedora-coreos-config/pull/2268

travier:
But agree that we should be using Ignition from C9S. I think we still waiting for a new compose with the latest one from it :/

travier:
https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/AppStream/x86_64/os/Packages/ignition-2.15.0-1.el9.x86_64.rpm > It should be there now. Wondering why we're not picking it from there.

travier:
I had asked ART to untag those pacakges in https://issues.redhat.com/browse/ART-6055 but looks like one sneaked in again?

cgwalters:
Argh, the snooze thing sometimes feels like a self-inflicted footgun.  

[...]

PR #1188, opened 2023-03-09
manifest/9.2: Ensure package come from RHEL by default

travier:
Make sure that we use RHEL (and not RHAOS) packages by default.

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -147,6 +147,15 @@ repo-packages:
       - nss-altfiles
       # Use the new containers/toolbox
       - toolbox
+      # RHCOS packages that are now part of RHEL and that will have to be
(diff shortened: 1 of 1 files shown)


travier:
```
ignition-2.15.0-1.el9.x86_64 (appstream)
```

This one is using the right version of Ingnition

jlebon:
Seems sane, but is there a misconfiguration or bug here? Why aren't we already picking the highest NEVRA?

/approve

cgwalters:
I'm confused, in https://github.com/openshift/os/issues/1187 we *are* using this version of Ignition and we are hitting that linked bug.

travier:
Hum, indeed. That's weird

travier:
OK so now this fails in 8.6 which does not have the new Ignition.

travier:
Cleaning things in https://github.com/openshift/os/pull/1189

PR #1194, opened 2023-03-14
s390x/kola-denylist: Re-enable working tests

jschintag:
fips.enable* is working on newest s390x. Remove the test from denylist. For more details see #546

coreos.ignition.mount.* is fixed with gdisk-1.0.3-9.el8. Remove the test from denylist.

ext.config.shared.ignition.stable-boot is fixed in RHEL 9. Only disable for RHEL 8 builds.


jlebon:
Sweet!
/lgtm

aaradhak (approved):
LGTM

PR #1196, opened 2023-03-14
denylist: Drop coreos.ignition.journald-log from denylist

gursewak1997:
Enabling `ignition.journald-log` as the test now runs and passes for RHCOS.


dustymabe:
> Closes: [coreos/coreos-assembler#1173](https://github.com/coreos/coreos-assembler/issues/1173)

Let's remove this from the commit and description of this PR. Unfortunately that issue doesn't exactly describe the underlying problem here and should be dealt with separately.

dustymabe (approved):
/lgtm

PR #1202, opened 2023-03-16
OCPBUGS-8008: manifest/9.2: Keep legacy coreos/toolbox & Revert "Ensure package come from RHEL by default"

travier:
Revert "manifest/9.2: Ensure package come from RHEL by default"

We already have to use coreos-installer from the RHAOS repo and it's
likely that we will have to do it again in the future for 4.14.

Remove this to avoid having to make a PR for each package.

This reverts commit 49f9824db2b6f929567d2d41cea6c97df766cf37.

---

manifest/9.2: Keep legacy coreos/toolbox

Unfortuantely, we're not yet ready to move to the new containers/toolbox
workflow. We'll attempt that for a future release.

See: https://issues.redhat.com/browse/OCPBUGS-8009
Partially reverts: https://github.com/openshift/os/commit/65776fa6e94fc25d73fd32a26f38547d55f61d44

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -145,8 +145,6 @@ repo-packages:
     packages:
       # We want the one shipping in RHEL, not the equivalently versioned one in RHAOS
       - nss-altfiles
-      # Use the new containers/toolbox
-      - toolbox
   - repo: rhel-9.0-server-ose-4.13
     packages:
       - conmon-rs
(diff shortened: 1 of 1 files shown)


cgwalters:
> error: No matches for 'toolbox' in repo 'rhel-9.0-server-ose-4.13'

Yeah...

cgwalters:
OK I started on a toolbox build (:cry:) but it now blocks on https://issues.redhat.com/browse/ART-6326

PR #1207, opened 2023-03-16
Enable sandboxed-containers extension on s390x

### LINKED ISSUE #1206: Enable sandboxed-containers extension in s390x
The title says it all, we are interested in having the sandboxed-containers operator available for s390x architecture.

etrunko:
Fixes #1206


mike-nguyen:
The [sig-virtualization repo](https://github.com/openshift/os/blob/2b2d749cbff9474d9c06adb745ffc5e03f8b4809/c9s.repo#L49-L55) doesn't have builds for [s390x](http://mirror.stream.centos.org/SIGs/9-stream/virt/).

etrunko:
Thanks for the notice, I will change it from x86_64 to $basearch and push force my commit.

cgwalters:
> The [sig-virtualization repo](https://github.com/openshift/os/blob/2b2d749cbff9474d9c06adb745ffc5e03f8b4809/c9s.repo#L49-L55) doesn't have builds for [s390x](http://mirror.stream.centos.org/SIGs/9-stream/virt/).

Right, this is going to be a repeated problem for us to do CentOS Stream builds.  Honestly it's annoying because it's not too hard to do builds in full emulation; it's what COPR does.

> Thanks for the notice, I will change it from x86_64 to $basearch and push force my commit.

That won't work.  I think actually what we need to do for now is probably not build the extensions container at all on c9s-s390x or something.

travier:
Replaced by https://github.com/openshift/os/pull/1326
/close

PR #1208, opened 2023-03-20
kola-denylist: Remove RHEL 8.6 entries

jschintag:
Since we are now no longer building RHEL 8,6 content, remove tests that are only disabled for RHEL 8.6


cgwalters (approved):

gursewak1997 (approved):
lgmt

PR #1218, opened 2023-03-24
extensions-c9s: use `match-base-evr: kernel` for kernel-rt

jlebon:
The kernel-rt merge has already happened in CentOS Stream 9. So using `match-base-evr` now works there. This means that rpm-ostree will guarantee that the kernel-rt package matches the version of the base kernel.

We don't build SCOS in the CoreOS pipelines today but let's exercise this knob as preparation for when the change propagates to RHEL.


jlebon:
CI is in a bad shape, but I've tested this locally.

jmarrero (approved):

PR #1231, opened 2023-03-31
Get containers stack from highest NEVRA

cgwalters:
We've oscillated on whether to get podman from RHEL or OCP channels.  Current decision is to just float with highest version.

In particular this ensures we get a `containers-common` that we want from the OCP channels.


cgwalters:
(Nevermind I was modifying c9s, AFAICS we're already floating for RHEL.)

Issue #1233, opened 2023-04-03
`ext.config.shared.chrony.dhcp-propagation` failing on 4.10 due to archived f35 repos


jlebon:
```
kola-runext-dhcp-propagation[1699]: + podman build -t dnsmasq .
kola-runext-dhcp-propagation[1699]: STEP 1: FROM registry.fedoraproject.org/fedora:35
kola-runext-dhcp-propagation[1699]: Trying to pull registry.fedoraproject.org/fedora:35...
kola-runext-dhcp-propagation[1699]: Getting image source signatures
kola-runext-dhcp-propagation[1699]: Copying blob sha256:6d1077252e7b0fc316ef55821b6e0097659a7592d41ea0ffe82f59648b076174
kola-runext-dhcp-propagation[1699]: Copying config sha256:175e191aee2ce4358b33628edf777b66d9fb251cb4e6be7c2df131c6bf83a539
kola-runext-dhcp-propagation[1699]: Writing manifest to image destination
kola-runext-dhcp-propagation[1699]: Storing signatures
kola-runext-dhcp-propagation[1699]: STEP 2: RUN dnf -y install systemd dnsmasq iproute iputils && dnf clean all && systemctl enable dnsmasq
kola-runext-dhcp-propagation[1699]: Fedora 35 - aarch64                              25 MB/s |  75 MB     00:03
kola-runext-dhcp-propagation[1699]: Fedora 35 openh264 (From Cisco) - aarch64       5.3 kB/s | 2.5 kB     00:00
kola-runext-dhcp-propagation[1699]: Fedora Modular 35 - aarch64                      94 kB/s | 3.2 MB     00:35
kola-runext-dhcp-propagation[1699]: Fedora 35 - aarch64 - Updates                   6.4  B/s | 196  B     00:30
kola-runext-dhcp-propagation[1699]: Errors during downloading metadata for repository 'updates':
kola-runext-dhcp-propagation[1699]:   - Status code: 404 for https://d2lzkl7pfhq30w.cloudfront.net/pub/fedora/linux/updates/35/Everything/aarch64/repodata/repomd.xml (IP: 108.138.82.221)
kola-runext-dhcp-propagation[1699]:   - Status code: 404 for http://dl.fedoraproject.org/pub/fedora/linux/updates/35/Everything/aarch64/repodata/repomd.xml (IP: 38.145.60.23)
[...]

PR #1234, opened 2023-04-03
[release-4.10] denylist: add `ext.config.shared.chrony.dhcp-propagation`

jlebon:
This is failing now due to stale Fedora content. Denylist for now at least to unblock builds.

See: https://github.com/openshift/os/issues/1233


travier:
/lgtm
/label backport-risk-assessed
/label cherry-pick-approved

CC @ashcrow @mrunalp @aravindhp for a `valid-bug` label override. Thanks

Issue #1237, opened 2023-04-04
c9s: Secure Boot tests failing


travier:
See:

dustymabe:
Once we figure out the secureboot problem with c9s we should revert https://github.com/openshift/os/commit/f779213c9af7abc42873e4642f6775d7ce4efabe and https://github.com/openshift/os/commit/db54842e16efbdc28b6a3fd8cfbeb0653423fda9

aaradhak:
The tests `iso-live-login.uefi-secure` & `iso-as-disk.uefi-secure` fail with c9s
```
[coreos-assembler]$ kola testiso -S iso-live-login.uefi-secure iso-as-disk.uefi-secure --output-dir tmp/kola-metal
⚠️  Skipping kola test pattern "ext.config.rpm-ostree.replace-rt-kernel":
  👉 https://github.com/openshift/os/issues/1099
Ignoring verification of signature on metal image
Running test: iso-as-disk.uefi-secure
FAIL: iso-as-disk.uefi-secure (10m0.002s)
    timed out after 10m0s
Running test: iso-live-login.uefi-secure
FAIL: iso-live-login.uefi-secure (10m0.002s)
    timed out after 10m0s
Error: harness: test suite failed
2023-06-09T17:21:53Z cli: harness: test suite failed
```

Console log output:
```
^[[2J^[[01;01H^[[=3h^[[2J^[[01;01H^[[2J^[[01;01H^[[=3h^[[2J^[[01;01HBdsDxe: loading Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x3,0x0)
BdsDxe: failed to load Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x3,0x0): Access Denied
BdsDxe: No bootable option or device was found.
BdsDxe: Press any key to enter the Boot Manager Menu.
```

LorbusChris:
Not sure if related, but SCOS still pulls this ancient version of shim from the C9S repos:
```
Earliest changed package: shim-x64-15-15.el8_2.x86_64 at 2020-08-01 01:55:42 UTC
```

dustymabe:
[...]

PR #1238, opened 2023-04-04
ci: Skip Secure Boot tests for SCOS

travier:
ci: Skip Secure Boot tests for SCOS

---

denylist: Skip Secure Boot tests for SCOS

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -131,7 +131,7 @@ kola_test_qemu() {
         cosa kola --basic-qemu-scenarios --skip-secure-boot
     fi
     kola run-upgrade -b rhcos -v --find-parent-image --qemu-image-dir tmp/ --output-dir tmp/kola-upgrade
-    cosa kola run --parallel 2
+    cosa kola run --parallel 2 --output-dir tmp/kola-all
 }
 
 # Build metal, metal4k & live images and run kola tests
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -117,7 +117,19 @@ cosa_build() {
 # Build QEMU image and run all kola tests
 kola_test_qemu() {
     cosa buildextend-qemu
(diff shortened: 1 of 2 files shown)


ci/prow-entrypoint.sh
```diff
@@ -117,9 +117,20 @@ cosa_build() {
 # Build QEMU image and run all kola tests
 kola_test_qemu() {
     cosa buildextend-qemu
-    cosa kola --basic-qemu-scenarios
+
+    # Skip Secure Boot tests on SCOS for now: https://github.com/openshift/os/issues/1039
```
jlebon, line 121: Did you mean https://github.com/openshift/os/issues/1237 here?

jlebon (approved):

travier:
Cross linking https://github.com/coreos/coreos-assembler/issues/3418

kola-denylist.yaml
```diff
@@ -25,3 +25,13 @@
...
+
+- pattern: iso-live-login.uefi-secure
+  tracker: https://github.com/openshift/os/issues/1237
+  osversion:
+  - c9s
+
+- pattern: iso-as-disk.uefi-secure
+  tracker: https://github.com/openshift/os/issues/1237
+  osversion:
+  - c9s
```
jlebon, line 37: Note that could be a single `pattern: *.uefi-secure` entry. But obviously fine as is too!

jlebon (approved):

cgwalters:
LGTM though I probably would have said we should pin on a working kernel.

dustymabe:
are the CI failures expected?

travier:
The RHCOS CI failures are expected as we are in a mid state with RHEL 9.2.

The SCOS builds should pass. I agree that pinning would have been better but unfortunately C9S kernel get garbage collected quite quickly so that would mean another failure soon.

So skipping the Secure Boot still let's us test the rest of the kernel for now.

PR #1243, opened 2023-04-05
overlay: add 30rhcos-nvme-compat-udev

jlebon:
The systemd in RHEL 9.2 now ships this commit:
https://github.com/systemd/systemd/commit/5118e8e71d

This changes how symlinks created from device serials with leading spaces are interpreted, from e.g.

```
nvme-QEMU_NVMe_Ctrl__foobar -> ../../nvme0n1
```

to

```
nvme-QEMU_NVMe_Ctrl_______foobar -> ../../nvme0n1
```

This in turn breaks anything/anyone that relied on the previous symlinks. Things in the storage stack for example that auto-discover block devices, or simply just customers customizing their storage.

Add back a rule similar to the previous version so that we keep creating the previous symlink. Also ship this in the initramfs in case Ignition configs referenced them.

The motivating factor for the systemd change was to fix incorrect behaviour when e.g. the model name contained a `/`; udevd would create a literal directory as a result:

https://github.com/systemd/systemd/issues/19309

This rule we're adding back is still susceptible to this issue, though we haven't seen it in the wild for RHCOS. We could deviate from the old rule and have something custom to handle this but it's probably not worth the effort. The new rule is still there creating the fixed symlink, so users could still use that one if such a case arises.

Fixes: https://issues.redhat.com/browse/OCPBUGS-11375
See also: https://github.com/systemd/systemd/issues/27155

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -22,6 +22,7 @@ ostree-layers:
   - overlay/20platform-chrony
   - overlay/21dhcp-chrony
   - overlay/25azure-udev-rules
+  - overlay/30rhcos-nvme-compat-udev
 
 arch-include:
   x86_64:
--- a/overlay.d/30rhcos-nvme-compat-udev/statoverride
+++ b/overlay.d/30rhcos-nvme-compat-udev/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>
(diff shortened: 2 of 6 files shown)


jlebon:
I think with https://github.com/coreos/coreos-assembler/pull/3419, we can add an external test for this. Will look at that.

jlebon:
Now with a test!

cgwalters (approved):
Nice, looks good to me!

cgwalters:
Probably we failed to complete all the kola tests...we need to stop running all the tests on every PR

PR #1244, opened 2023-04-05
ci: Neuter `rhcos-92-build-test-*` for now

cgwalters:
FIXME skipped until we have updated mirrors in Prow


jlebon:
Backported to 4.13 in https://github.com/openshift/os/pull/1246.

PR #1245, opened 2023-04-06
s390x/tests: Do not run nvme-nvme-compat-symlink on s390x

jschintag:
qemu on s390x does not support nvme disk emulation.


c4rt0 (commented):
/lgtm

jlebon:
Thanks for the fix!

jlebon:
Backported to 4.13 in https://github.com/openshift/os/pull/1246.

Issue #1216, opened 2023-03-23
[4.13-9.2] Allow root login and password authentication that both are disabled for rhcos8


HuijingHei:
Check /etc/ssh/sshd_config in rhcos9, which should allow root login, also password authentication.
```
[root@cosa-devsh ~]# grep PermitRootLogin /etc/ssh/sshd_config
#PermitRootLogin prohibit-password

[root@cosa-devsh ~]# grep PasswordAuthentication /etc/ssh/sshd_config
#PasswordAuthentication yes
```
They are both disabled for rhcos8 according to https://github.com/openshift/os/blob/master/common.yaml#L65-L67, and mismatch the sed condition in rhcos9, is it necessary to disabled them for rhcos9? (The change should be minor, but need to confirm the requirement)

HuijingHei:
From @cgwalters :
OK I think what happened here is...a long time ago, the redhat-coreos config git was a fork of fedora-coreos-config.   This specific line of code dates to https://github.com/openshift/os/commit/3178c3d866f2f4b4cbdd4d09f45d7b02025ff985
Then in FCOS, https://github.com/coreos/fedora-coreos-config/commit/48be8b6672ebbd48ff6f032738f883744e28fdd5 happened, but we never dropped the corresponding logic from RHCOS.
My PoV is we should delete that line, and inherit the FCOS logic which is really the same as RHEL9's logic now too.
So by deleting this little bit of code, we make RHEL CoreOS more like RHEL.  (And also, not different from FCOS)

cgwalters:
OK yeah, I traced the change here back to pre-public commit e6a1207b439cf4b4e73547afd8c878fed72683e0 Fri May 31 2019 -0400 AKA https://url.corp.redhat.com/cef7fe3

This is still relatively early days.  It was later that year in commit 7c21ba398aca1a1787845038539c7906a2181e04 Thu Jul 18 2019 -0400 titled "Use fedora-coreos-config as a git submodule" when things were unified, but we didn't merge those two bits of code.

cgwalters:
Just to reiterate though: since we're on track to ship 4.13 with this change anyways (i.e. switching from `PermitRootLogin no` to `PermitRootLogin prohibit-password`) and it'd take extra effort to undo the switch now and return to RHEL8 code, I think what's best is to just delete the text and effectively inherit the change from RHEL8 to RHEL9.

HuijingHei:
[...]

PR #1251, opened 2023-04-11
OCPBUGS-11613: manifests: Use a file drop in for sshd config in EL9

travier:
manifests: Use a file drop in for sshd config in EL9

The first two sed lines were silently broken with the default
configuration change in EL 9. Restore a subset as a drop-in config file:

- PasswordAuthentication: Kept disabled for now until the MCO learns to
  disable it once a password is set for a user.
- PermitRootLogin: Kept disabled as well for now. The new default in
  RHEL 9 is to not accept a password to log in as root. We will likely change
  back to this default in an future release.
- ClientAliveInterval: Kept as is until we investigate more.

Fixes: https://github.com/openshift/os/issues/1216
See: https://issues.redhat.com/browse/OCPBUGS-11613

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -63,13 +63,6 @@ postprocess:
      #!/usr/bin/env bash
      set -xeo pipefail
 
-     # Disable PasswordAuthentication in SSH
-     sed -i "s|^PasswordAuthentication yes$|PasswordAuthentication no|g" /etc/ssh/sshd_config
-     # Disable root login because don't do that.
-     sed -i "s|^PermitRootLogin yes$|PermitRootLogin no|g" /etc/ssh/sshd_config
-     # Enable ClientAliveInterval and set to 180 per https://bugzilla.redhat.com/show_bug.cgi?id=1701050
-     sed -i "s|^#ClientAliveInterval 0$|ClientAliveInterval 180|g" /etc/ssh/sshd_config
-
      # TEMPORARY: Create /etc/vmware-tools/tools.conf to ensure RHCOS shows up properly in VMWare
      # See https://jira.coreos.com/browse/RHCOS-258
      if [ "$(uname -m)" == "x86_64" ]; then
--- a/overlay.d/07el9/etc/ssh/sshd_config.d/40-rhcos-defaults.conf
+++ b/overlay.d/07el9/etc/ssh/sshd_config.d/40-rhcos-defaults.conf
@@ -1,0 +1,10 @@
+# Disable PasswordAuthentication and PermitRootLogin to preserve RHCOS 8
+# defaults for now
+# See: https://issues.redhat.com/browse/OCPBUGS-11613
+# See: https://github.com/openshift/os/issues/1216
+PasswordAuthentication no
+PermitRootLogin no
+
+# Enable ClientAliveInterval and set to 180
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1701050
+ClientAliveInterval 180


overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
```diff
@@ -0,0 +1,6 @@
+# Disable PasswordAuthentication in SSH
+PasswordAuthentication no
+# Disable root login because don't do that
+PermitRootLogin no
```
cgwalters, line 4: I think this would then again re-break the change made for tmt in

https://github.com/teemtee/tmt/commit/cd9d3176da1355fb99cd9a085498a7b23de0ca08

Given the MCO doesn't support setting a password for `root`, I think we can safely drop this.

(That said, the tmt change should have been using a drop-in too...)

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
```diff
@@ -0,0 +1,6 @@
+# Disable PasswordAuthentication in SSH
```
cgwalters, line 1: Let's change this comment to now explicitly link to https://issues.redhat.com/browse/OCPBUGS-11613

And also call out that this is a change we need to carry over because we did this in the RHEL8 timeframe, but we could potentially drop it by default in the future (for example, by only disabling password logins for the `core` user).

cgwalters:
If we agree to take this approach, please link this PR to the JIRA officially.

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
travier: 👍🏻

common.yaml
```diff
@@ -63,13 +63,6 @@ postprocess:
      #!/usr/bin/env bash
      set -xeo pipefail
 
-     # Disable PasswordAuthentication in SSH
-     sed -i "s|^PasswordAuthentication yes$|PasswordAuthentication no|g" /etc/ssh/sshd_config
-     # Disable root login because don't do that.
-     sed -i "s|^PermitRootLogin yes$|PermitRootLogin no|g" /etc/ssh/sshd_config
-     # Enable ClientAliveInterval and set to 180 per https://bugzilla.redhat.com/show_bug.cgi?id=1701050
-     sed -i "s|^#ClientAliveInterval 0$|ClientAliveInterval 180|g" /etc/ssh/sshd_config
-
```
dustymabe, line 72: Did these sed statements stop working because the formatting in `/etc/ssh/sshd_config` changed?

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
dustymabe: This would allow logging in as root with a key, though? Seems like we should keep `PermitRootLogin no` if we're keeping the other stuff and fix tmt.

common.yaml
travier: See the `^` in the sed lines

travier: The last one likely worked but I've just moved it for consistency.

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
```diff
@@ -0,0 +1,8 @@
+# Disable PasswordAuthentication in SSH to preserve RHCOS 8 defaults
+# See: https://issues.redhat.com/browse/OCPBUGS-11613
+# See: https://github.com/openshift/os/issues/1216
+PasswordAuthentication no
```
jlebon, line 4: We have `50-disable-passwords.conf` in FCOS that also does this. Should we share it instead of redefining it here?

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
```diff
@@ -0,0 +1,8 @@
...
+
+# Enable ClientAliveInterval and set to 180
+# See: https://bugzilla.redhat.com/show_bug.cgi?id=1701050
+ClientAliveInterval 180
```
jlebon, line 8: Aside: might be good to check whether this is still needed in el9. E.g. we don't have this in FCOS and no one has complained AFAIK. But if it's an issue there too, we should upstream it to FCOS. If not, then we can drop it here too.

Not a blocker.

overlay.d/07el9/etc/ssh/sshd_config.d/41-rhcos-defaults.conf
travier: Looks like this hack https://github.com/teemtee/tmt/commit/cd9d3176da1355fb99cd9a085498a7b23de0ca08 will not work for RHCOS 9 anyway so it will need a change.

travier:
https://issues.redhat.com/browse/COS-2155

travier:
To clarify, I think we should merge this as is to keep this easier to backport. Improvements will happen in a followup PR.

jlebon (approved):
/approve
/lgtm

overlay.d/07el9/etc/ssh/sshd_config.d/40-rhcos-defaults.conf
jlebon, line 1: Just noting that 07el9 is included by `common.yaml`, so this should be affecting the exact same variants.

travier: Yes, I was planning on doing a large el9/common clean up now that we only have EL9 variants but saving that for a later PR.

travier:
Adding a test for this in https://github.com/openshift/os/pull/1255

travier:
/retitle OCPBUGS-11764: manifests: Use a file drop in for sshd config in EL9

PR #1254, opened 2023-04-12
manifests: Merge EL9 manifests and overlays

travier:
Now that we only have RHCOS based on RHEL 9 and SCOS based on C9S, we can merge the manifests again.


HuijingHei (approved):
LGTM

PR #1255, opened 2023-04-12
tests: Add OpenSSH config test

travier:
See: https://github.com/openshift/os/pull/1251
See: https://github.com/openshift/os/issues/1216
See: https://issues.redhat.com/browse/OCPBUGS-11764

### DIFF
--- a/tests/kola/ssh/config
+++ b/tests/kola/ssh/config
@@ -1,0 +1,24 @@
+#!/bin/bash
+## kola:
+##   exclusive: false
+#
+# Verify OpenSSH config options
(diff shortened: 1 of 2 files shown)


travier:
Will likely move / share this test with FCOS.

tests/kola/ssh/config
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
```
HuijingHei, line 2: Just one minor question, do we need to add `kola` section like:
```
## kola:
##   exclusive: false
```

PR #1261, opened 2023-04-17
kola-denylist: drop rhcos.network.*

jlebon:
This was fixed by https://github.com/coreos/coreos-assembler/pull/3431.


jlebon:
This is ready now that the cosa PR is merged!

PR #1266, opened 2023-04-18
OCPBUGS 11935: Pin container-selinux to ART candidate repo

thegreyd:
Fixing https://issues.redhat.com/browse/OCPBUGS-11935
[slack ref](https://redhat-internal.slack.com/archives/CB95J6R4N/p1681826692751499)

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -122,6 +122,7 @@ repo-packages:
       - nss-altfiles
   - repo: rhel-9.2-server-ose-4.13
     packages:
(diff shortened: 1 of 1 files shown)


sdodson:
/retitle OCPBUGS-11935: Pin container-selinux to ART candidate repo

jlebon:
There seems to be agreement internally to instead just fix the epoch issue so we naturally get the RPM from the rhaos repo and move back to RHEL once it's updated enough. Let's close this. We can revive it if things change.

/close

PR #1267, opened 2023-04-18
overlay.d: Add 30gcp-udev-rules overlay from FCOS

ravanelli:
- Add 30gcp-udev-rules from FCOS
 - See https://github.com/coreos/fedora-coreos-config/pull/2350 for more info.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -24,6 +24,7 @@ ostree-layers:
   - overlay/21dhcp-chrony
   - overlay/25azure-udev-rules
   - overlay/30rhcos-nvme-compat-udev
+  - overlay/30gcp-udev-rules
 
 arch-include:
   x86_64:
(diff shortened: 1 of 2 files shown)


travier:
LGTM but as we have no test for this, it's hard to now if it works. I've suggested in https://github.com/coreos/fedora-coreos-tracker/issues/1457#issuecomment-1513252754 that we should do the testing in two parts to make sure that we at least have the rules in the initrd before we merge this.

HuijingHei:
Build with the patch locally and check the result is expected.
```
[core@cosa-devsh ~]$ ls -l /usr/lib/udev/rules.d/*gce*
-rw-r--r--. 2 root root 1077 Jan  1  1970 /usr/lib/udev/rules.d/64-gce-disk-removal.rules
-rw-r--r--. 2 root root 2000 Jan  1  1970 /usr/lib/udev/rules.d/65-gce-disk-naming.rules
[core@cosa-devsh ~]$ ls -l /usr/lib/udev/google_nvme_id
-rw-r--r--. 2 root root 9186 Jan  1  1970 /usr/lib/udev/google_nvme_id
```

PR #1269, opened 2023-04-19
denylist: remove ext.config.kdump.crash denial

Shilpi-Das1:
The underlying issue for kdump crash test failure on ppc64le has been fixed in the upstream kernel [1]. We can remove this test from the denylist now since the fix has made it to all streams that we care about.

[1] coreos/coreos-assembler#2725 (comment)
Signed-off-by: Shilpi Das <shdas@redhat.com>

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -11,10 +11,6 @@
    - rhel-8.6
   arches:
   - s390x
-- pattern: ext.config.shared.kdump.crash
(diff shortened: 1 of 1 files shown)


ravanelli (approved):
/lgtm

ravanelli:
@Prashanth684 Are you able to add the missing labels?  `staff-eng-approved` and `bugzilla-valid-bug`?

PR #1271, opened 2023-04-20
OCPBUGS-12203: overlay: Inject SYSTEMD_JOURNAL_COMPRESS=LZ4 by default

travier:
With the latest fix in systemd [1], we can use an environment variable
to tell journald to use a specific compression algorithm for the
journal and thus don't need a custom build anymore.

See: systemd/systemd#27126 [1]
See: #1230
See: https://issues.redhat.com/browse/LOG-3832
See: https://bugzilla.redhat.com/show_bug.cgi?id=2183546


travier:
We need to wait for this to land in C9S to pass.

jamacku:
@travier @cgwalters @sdodson Is this PR ready to be merged? Thanks

cgwalters:
> @travier @cgwalters @sdodson Is this PR ready to be merged? Thanks

It seems like the relevant support still hasn't been merged into c9s?  Oh, I see you just did https://gitlab.com/redhat/centos-stream/rpms/systemd/-/commit/96f92a96d38f85ce28be052bc26903fba026c6f1 but that isn't shipped yet in a compose.

jamacku:
I expect a new build of systemd to be available in c9s soon.

sdodson:
@cgwalters Even if it hasn't landed in CS9 we want this to land before a new build shows up in RHCOS so that we don't immediately revert because we lost the hardcoding? We also want to get this into a boot image bump?

cgwalters:
Yes; what we can do here is to just snooze the test on c9s.  I've done that now.
/lgtm

travier:
4.13 cherrypick in https://github.com/openshift/os/pull/1288

PR #1279, opened 2023-05-10
OCPBUGS-11124, OCPBUGS-11411: overlay: Inject pcrphase service definition

mkowalski:
This PR removes `After=` section from the definition of
systemd-pcrphase. It is because currently it blocks possibility to SSH
into the node which for any reason has nodeip-configuration or
configure-ovs not succeeding.

The self-healing functionality of the latter creates a scenario in which
network-online.targed is not yet reached but we already want to access
the node for debugging purposes.

At the same time as by default systemd-pcrphase blocks user sessions and
depends on remote-fs, this creates a deadlock. In order to remediate
this situation, we are removing dependency on remote-fs here. It is
justified as OpenShift nodes are not meant to use remote home
directories.

We are also modifying dependencies of systemd-user-sessions.service so
that we are explicitly saying that we do not need network in order to
allow access to the node.

Fixes: OCPBUGS-11124
Fixes: OCPBUGS-11411

### DIFF
--- a/overlay.d/15rhcos-pcrphase-deps/etc/systemd/system/systemd-pcrphase.service
+++ b/overlay.d/15rhcos-pcrphase-deps/etc/systemd/system/systemd-pcrphase.service
@@ -1,0 +1,27 @@
+#  SPDX-License-Identifier: LGPL-2.1-or-later
+#
+#  This file is part of systemd.
+#
+#  systemd is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+# This is a manually copy-pasted unit definition with `After=` section removed.
+# It is because dependency on remote-fs.target causes lockdown in case of any
+# network issues, causing this service to block ability to login to the node.
+# Related issues: OCPBUGS-11124, OCPBUGS-11411
+
+[Unit]
+Description=TPM2 PCR Barrier (User)
+Documentation=man:systemd-pcrphase.service(8)
+Before=systemd-user-sessions.service
+AssertPathExists=!/etc/initrd-release
+ConditionSecurity=tpm2
+ConditionPathExists=/sys/firmware/efi/efivars/StubPcrKernelImage-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/systemd/systemd-pcrphase ready
+ExecStop=/usr/lib/systemd/systemd-pcrphase shutdown
--- a/overlay.d/15rhcos-pcrphase-deps/statoverride
+++ b/overlay.d/15rhcos-pcrphase-deps/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>
--- a/overlay.d/15rhcos-user-sessions-deps/etc/systemd/system/systemd-user-sessions.service
+++ b/overlay.d/15rhcos-user-sessions-deps/etc/systemd/system/systemd-user-sessions.service
@@ -1,0 +1,24 @@
+#  SPDX-License-Identifier: LGPL-2.1-or-later
+#
+#  This file is part of systemd.
+#
+#  systemd is free software; you can redistribute it and/or modify it
+#  under the terms of the GNU Lesser General Public License as published by
+#  the Free Software Foundation; either version 2.1 of the License, or
+#  (at your option) any later version.
+
+# This is a manually copy-pasted unit definition with `After=` section modified.
+# It is because in OpenShift installations we want to allow access to the system
+# even without network completely configured.
+# Related issues: OCPBUGS-11124, OCPBUGS-11411
+
+[Unit]
+Description=Permit User Sessions
+Documentation=man:systemd-user-sessions.service(8)
+After=nss-user-lookup.target home.mount
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/systemd/systemd-user-sessions start
+ExecStop=/usr/lib/systemd/systemd-user-sessions stop
--- a/overlay.d/15rhcos-user-sessions-deps/statoverride
+++ b/overlay.d/15rhcos-user-sessions-deps/statoverride
@@ -1,0 +1,2 @@
+# Config file for overriding permission bits on overlay files/dirs
+# Format: =<file mode in decimal> <absolute path to a file or directory>


overlay.d/15rhcos-pcrphase-deps/etc/systemd/system/systemd-pcrphase.service
```diff
@@ -0,0 +1,27 @@
...
+# network issues, causing this service to block ability to login to the node.
+# Related issues: OCPBUGS-11124, OCPBUGS-11411
+
+[Unit]
+Description=TPM2 PCR Barrier (User)
+Documentation=man:systemd-pcrphase.service(8)
+Before=systemd-user-sessions.service
+AssertPathExists=!/etc/initrd-release
+ConditionSecurity=tpm2
+ConditionPathExists=/sys/firmware/efi/efivars/StubPcrKernelImage-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
```
zaneb, line 21: What's the significance of this UUID? This looks painful to keep in sync with the RPM.

mkowalski: It's just a hardcoded value, seems like a global UUID used by EFI, https://systemd.io/BOOT_LOADER_INTERFACE/

jlebon:
OK, I have read some of the background on this. ISTM like we should just mask all the `systemd-pcrphase` stuff for now instead. Reading `systemd-pcrphase.service(8)`, it and other PCR-related services are only used when booting a UKI via systemd-stub. UKIs are on the roadmap eventually, so we may have to undo this masking at some point, but for now there's no point in trying keep it working if it will need to be adjusted further. I doubt we'll be the only ones wanting to e.g. be able to log into the host to debug things before networking is up.

mkowalski:
I am not sure if I'm missing something, but we can't do lots of operations on this service (i.e. I cannot mask it as well as we couldn't simply disable it). Maybe it's related to the fact that this is some systemd-originated service

```
[root@master-0 kubelet.service.d]# systemctl mask systemd-pcrphase.service
Failed to mask unit: File /etc/systemd/system/systemd-pcrphase.service already exists.
```

cgwalters:
Sorry I know we've debugged this in multiple places in those bugs, but are you really, 100% sure that this will fix the problem?

> This PR removes After= section from the definition of systemd-pcrphase. 

One thing I notice here is that indeed we have
```
[root@cosa-devsh ~]# rpm-ostree status
State: idle
Deployments:
● 4ab367a7d23cc53de1b3c5ddd0e1939c0458d912d3eb69df6d33afa0f6c34659
                  Version: 413.92.202303281804-0 (2023-03-28T18:07:15Z)
[root@cosa-devsh ~]# grep After= /usr/lib/systemd/system/systemd-pcrphase.service
After=remote-fs.target remote-cryptsetup.target
[root@cosa-devsh ~]#
```

*but* also:

```
[root@cosa-devsh ~]# grep After= /usr/lib/systemd/system/systemd-user-sessions.service
After=remote-fs.target nss-user-lookup.target network.target home.mount
[root@cosa-devsh ~]#
```

So `systemd-pcrphase.service` is just pulling in `remote-cryptsetup.target` in addition.

I am still not convinced that I was wrong over here https://issues.redhat.com/browse/OCPBUGS-11124?focusedId=22056799&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22056799

jlebon:
> I am not sure if I'm missing something, but we can't do lots of operations on this service (i.e. I cannot mask it as well as we couldn't simply disable it). Maybe it's related to the fact that this is some systemd-originated service
> 
> ```
> [root@master-0 kubelet.service.d]# systemctl mask systemd-pcrphase.service
> Failed to mask unit: File /etc/systemd/system/systemd-pcrphase.service already exists.
> ```

To address this part: is this with a custom RHCOS containing this PR? It worked for me locally on the latest RHCOS 4.13 build. If we go this way, I can provide more information about how we'd do this at compose time.

If you want to test whether it fixes the issue, you can also write an Ignition config/MC to mask the unit.

mkowalski:
> Sorry I know we've debugged this in multiple places in those bugs, but are you really, 100% sure that this will fix the problem?
> 
> > This PR removes After= section from the definition of systemd-pcrphase.
> 
> One thing I notice here is that indeed we have
> 
>[...]
> ```
> [root@cosa-devsh ~]# grep After= /usr/lib/systemd/system/systemd-user-sessions.service
> After=remote-fs.target nss-user-lookup.target network.target home.mount
> [root@cosa-devsh ~]#
> ```
> 
> So `systemd-pcrphase.service` is just pulling in `remote-cryptsetup.target` in addition.
> 
> I am still not convinced that I was wrong over here https://issues.redhat.com/browse/OCPBUGS-11124?focusedId=22056799&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-22056799

The more I look at this the more confused I am, but I am now looking at RHCOS 411.86.202302021552-0 and what I see is following

```
[Unit]
Description=Permit User Sessions
Documentation=man:systemd-user-sessions.service(8)
After=remote-fs.target nss-user-lookup.target network.target home.mount

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=/usr/lib/systemd/systemd-user-sessions start
ExecStop=/usr/lib/systemd/systemd-user-sessions stop
```

So we have OCP 4.11 which is **not** affected by this problem but `systemd-user-sessions.service` also has `After=remote-fs.target`. This is the reason why we never pointed fingers directly at `remote-fs`.

But the more I think about it, the more I feel that maybe we should simply edit `systemd-user-sessions.service` so that instead of `After=remote-fs.target nss-user-lookup.target network.target home.mount` it gets only `After=nss-user-lookup.target network.target home.mount`

My rationale overall is as follows - between RHEL8 and RHEL9 we didn't change definitions of already existing stuff. But we did introduce pcrphase service that has some dependencies. By removing those dependencies, we almost go back to the previous state.

Note that pcrphase service pulls `After=remote-fs.target remote-cryptsetup.target` and I am removing both of them. Even if somewhere along the way we took remote-fs as our target when we should have focused on remote-cryptsetup, pcrphase now becomes dependency-less

travier:
I agree with Colin and I think that the problem is somewhere else. If masking this unit really unblocks us then I'll say OK but this still needs more investigation.

---

You don't need to copy/paste an entire unit to change `After=` dependencies. You can write a drop-in in `foo.service.d/override.conf` with:

```
[Unit]
# Reset dependencies
After=
# Re-add only those we want
After=foo.service bar.target
```

mkowalski:
> You don't need to copy/paste an entire unit to change `After=` dependencies. You can write a drop-in in `foo.service.d/override.conf` with:

That's not true. As per upstream documentation https://github.com/systemd/systemd/blob/250e9fadbcc0ca90e697d7efb40855b054ed3b8f/man/systemd.unit.xml#L1786 you cannot set empty `After=` in a drop-in

travier:
> That's not true. As per upstream documentation [systemd/systemd@`250e9fa`/man/systemd.unit.xml#L1786](https://github.com/systemd/systemd/blob/250e9fadbcc0ca90e697d7efb40855b054ed3b8f/man/systemd.unit.xml#L1786) you cannot set empty `After=` in a drop-in

Wow, nice catch. I did not know that. Thanks and sorry.

travier:
Shouldn't masking the unit also work in this case?

jlebon:
Here's what I've been playing with

```yaml
variant: fcos
version: 1.4.0
systemd:
  units:
    # override the builtin openvswitch.service
    - name: openvswitch.service
      enabled: true
      contents: |
        [Unit]
        Description=Fake Open vSwitch
        # These match openvswitch.service
        Before=network.target network.service
        After=network-pre.target ovsdb-server.service ovs-vswitchd.service
        PartOf=network.target
        Requires=ovsdb-server.service
        Requires=ovs-vswitchd.service

        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStart=true

        [Install]
        WantedBy=multi-user.target
    - name: nodeip-configuration.service
      enabled: true
      contents: |
        [Unit]
        Description=Fake nodeip-configuration
        # These match the nodeip-configuration.service template
        Wants=NetworkManager-wait-online.service crio-wipe.service
        After=NetworkManager-wait-online.service ignition-firstboot-complete.service crio-wipe.service
        Before=kubelet.service crio.service ovs-configuration.service

        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStart=sleep infinity

        [Install]
        WantedBy=multi-user.target
    - name: ovs-configuration.service
      enabled: true
      contents: |
        [Unit]
        Description=Fake ovs-configuration
        # These match the ovs-configuration.service template
        Requires=openvswitch.service
        Wants=NetworkManager-wait-online.service
        After=NetworkManager-wait-online.service openvswitch.service network.service nodeip-configuration.service
        Before=network-online.target kubelet.service crio.service node-valid-hostname.service

        [Service]
        Type=oneshot
        RemainAfterExit=yes
        ExecStart=sleep infinity

        [Install]
        WantedBy=multi-user.target
```

So `nodeip-configuration.service` and `ovs-configuration` are set up to indefinitely hang. But testing on both the latest 4.12 and 4.13 releases, I get the same result: the serial console comes up after a short delay (maybe 5s?). Once in, I can tell that `systemd-user-sessions.service` and `sshd.service` are up and running.

So likely there's some other detail I'm missing in this mock test that actually makes it hang on 4.13 but not 4.12.

mkowalski:
> So `nodeip-configuration.service` and `ovs-configuration` are set up to indefinitely hang. But testing on both the latest 4.12 and 4.13 releases, I get the same result: the serial console comes up after a short delay (maybe 5s?). Once in, I can tell that `systemd-user-sessions.service` and `sshd.service` are up and running.

I may need to digest the details, but this is not the behaviour we see in OCP 4.13; what I do see when setting `nodeip-configuration.service` to hang indefinitely is that I do not have the serial console usable (so no matter if I had root password or not, I couldn't login anyway)

Not sure what's missing from the butane config, but I don't understand how possibly we could get a console with `nodeip-configuration.service` that hangs

2023-05-22-123621_1062x931_scrot

jlebon:
That's interesting. I'd like to dig into this. I'm mostly offline today but will take a look at it more deeply tomorrow.

jlebon:
OK, I think this is superseded by https://github.com/openshift/os/pull/1294.

PR #1280, opened 2023-05-11
OCPBUGS-11124, OCPBUGS-11411: overlay: Mask pcrphase service

mkowalski:
This PR masks systemd-pcrphase service which is not used at the moment. Its existence creates issues as it depends on the remote-fs and as a consequence blocks access to the system if network configuration is not correct.

Anyway in OpenShift we are not using remote home directories, so this would not be useful for us.

We are also modifying dependencies of systemd-user-sessions.service so that we are explicitly saying that we do not need network in order to allow access to the node.

Fixes: OCPBUGS-11124
Fixes: OCPBUGS-11411

### DIFF
--- a/manifest-rhel-9.2.yaml
+++ b/manifest-rhel-9.2.yaml
@@ -108,6 +108,9 @@ postprocess:
      ---
      EOF
 
+     # Tweak dependencies of systemd-user-sessions.service to not rely on network.
+     sed -i 's/^After=.*/After=nss-user-lookup.target home.mount/' /usr/lib/systemd/system/systemd-user-sessions.service
+
 # Packages that are only in RHCOS and not in SCOS or that have special
 # constraints that do not apply to SCOS
 packages:
--- a/overlay.d/15rhcos-pcrphase-mask/etc/systemd/system/systemd-pcrphase.service
+++ b/overlay.d/15rhcos-pcrphase-mask/etc/systemd/system/systemd-pcrphase.service
@@ -1,0 +1,1 @@
+/dev/null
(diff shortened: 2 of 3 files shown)


mkowalski:
/cc @jlebon 

Can you take a look if this way of masking a service will work in the assembly?

manifest-rhel-9.2.yaml
```diff
@@ -108,6 +108,11 @@ postprocess:
      ---
      EOF
 
+     # Tweak dependencies of systemd-user-sessions.service to not rely on network.
+     # Copy file to /etc because /usr is a read-only filesystem.
```
cgwalters, line 112: No.  In this case we're doing this at build time to *generate* the content in `/usr`.

cgwalters: (This is a fundamental difference between doing a PR for this repository versus e.g. https://github.com/openshift/machine-config-operator/ - the latter case is "configuration" that is lifecycled outside of the OS and lives in `/etc` or `/var` mostly; the content here goes in `/usr`)

In a [layering](https://docs.openshift.com/container-platform/4.12/post_installation_configuration/coreos-layering.html#coreos-layering) world, this is unified because it's equally easy to drop higher level content in `/usr` too.

mkowalski: Oh this is great, makes it even easier for me then. Just fixed and now I modify directly in `/usr`

jlebon:
I think this is also superseded by https://github.com/openshift/os/pull/1294.

mkowalski:
/close

Indeed https://github.com/openshift/os/pull/1294 is solving the issue

PR #1285, opened 2023-05-17
extensions: Remove explicit mention to appstream repo

travier:
This repo is already enabled in the main RHEL 9.2 manifest thus we don't need to explicitely re-enable it here.


c4rt0 (approved):
LGTM

PR #1289, opened 2023-05-19
ci: enable ci tests as we have updated to RHEL 9.2 GA repos

HuijingHei:
See https://github.com/openshift/release/pull/39361

PR #1288, opened 2023-05-19
[4.13] OCPBUGS-12206: overlay: Inject SYSTEMD_JOURNAL_COMPRESS=LZ4 by default

travier:
With the latest fix in systemd [1], we can use an environment variable to tell journald to use a specific compression algorithm for the journal and thus don't need a custom build anymore.

See: https://github.com/systemd/systemd/pull/27126 [1] See: https://github.com/openshift/os/pull/1230
See: https://issues.redhat.com/browse/LOG-3832
See: https://bugzilla.redhat.com/show_bug.cgi?id=2183546 (cherry picked from commit dd2059d2e5f3d1bcd194730bf8f3b5d89fac4a7f)

PR #1291, opened 2023-05-22
Fix prow CI errors for RHCOS jobs

mike-nguyen:
I didn't have permission to add on top of https://github.com/openshift/os/pull/1289, so this PR encompasses @HuijingHei's changes as well as fixes to CI.  See individual commits for more details.

https://github.com/openshift/release/pull/39586 would also fix CI but the cleanup here is needed regardless.


HuijingHei:
Thanks @mike-nguyen for fixing this.

/retest-required

travier:
I've backported those changes in https://github.com/openshift/os/pull/1288

PR #1292, opened 2023-05-23
Move non-C9S/RHEL packages to dedicated manifest

LorbusChris:
This is a first step towards separating a base SCOS from OKD v4.x layers

The `packages-openshift.yaml` could later be removed from the SCOS compose, and instead be consumed by `rpm-opstree ex rebuild`  in an ostree layering (derivative container) build.

In OKD, these packages are all sourced from either CentOS SIGs or OpenShift, removing them from the SCOS base will be a big step towards making SCOS an official CentOS deliverable, as the remaining packages all stem from C9S repos.

/cc @travier @sherine-k


sherine-k (approved):
That's great! Thanks!

travier:
/lgtm
Let's make sure this passes CI.

travier:
Maybe we could start calling the "base" SCOS manifest just `scos`, and the one with the OpenShift packages `scos-full` as we'll want to drop it in the end.

LorbusChris:
> Maybe we could start calling the "base" SCOS manifest just `scos`,

:+1: 

> and the one with the OpenShift packages `scos-full` or `scos-ocp` as we'll want to drop it in the end.

`scos-okd` for community release builds, `scos-ocp` for Prow CI builds?

travier:
/lgtm

I'd prefer we keep just `scos` for the variant that we'll want to have named like that in the end as it won't be only for OKD.

travier:
Test too too long 😕 
/retest

PR #1298, opened 2023-05-26
overlay: stop disabling `iscsid.socket`

jlebon:
The iSCSI daemon is now socket-activated so that it's only running when
needed rather than always enabled. We're breaking that by disabling
`iscsid.socket`.

This effectively reverts 929ac48 ("c9s: Disable iscsid.socket"). It's
not certain why this was done, but it was likely to workaround failing
tests. These tests should be fixed now[[1]] so we should be able to stop
doing this.

[1]: https://github.com/coreos/coreos-assembler/pull/3275


jlebon:
Requires: https://github.com/openshift/os/pull/1294

jlebon:
And now rebased!

PR #1302, opened 2023-06-01
[release-4.13] OCPBUGS-14357: Disable iscsi.service & re-enable iscsid.socket

travier:
Cherry-picked from:
 - https://github.com/openshift/os/pull/1294
 - https://github.com/openshift/os/pull/1298
 - https://github.com/openshift/os/pull/1305

Needs:
  - https://github.com/coreos/coreos-assembler/pull/3497
  - https://github.com/coreos/fedora-coreos-config/pull/2448

---

overlay: disable iscsi.service by default

`iscsi.service` has `Before=remote-fs-pre.target` *and*
`After=network-online.target`. This forces `remote-fs-pre.target`
to block on `network-online.target` and hence in OCP, on
`ovs-configuration.service` (which has `Before=network-online.target`).

So this transitively makes `systemd-user-sessions.service` block on
`network-online.target`.

This was an issue in Fedora as well and was discussed in a devel
thread[[1]]. `iscsi.service` was subsequently reworked[[2]][[3]] so that
it was only activated if iSCSI was actually used by the system.

On RHEL 8, `iscsi.service` and co. were directly enabled by RPM
scriptlets rather than using presets. In RHCOS, we explicitly make
presets canonical[[4]] so we shipped with `iscsi.service` disabled by
default. On RHEL 9, the units were fixed to use presets[[5]]. This is
why we started seeing this issue after moving to RHEL 9.

So all we need in theory is to have the Fedora patch backported to RHEL
9. However, since we don't really need the functionality from
`iscsi.service` by default in RHCOS, we can fast-track its
(re-)disablement and not wait for the `iscsi-starter.service` workaround.

Note that `iscsi.service` is only used to bring up iSCSI sessions
marked for autostart in `/var/lib/iscsi/nodes` and is separate from
`iscsid.service`, which is what actually manages the iSCSI connections.
In OpenShift, we rely on the latter only (e.g. configured iSCSI PVCs
are done by the kubelet directly calling out to `iscsiadm`). It's
also separate from iSCSI devices that use host bus adapters, which are
transparent to RHCOS/OCP.

Fixes: https://issues.redhat.com/browse/OCPBUGS-11124

[1]: https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HACVEJ3FMOIM2TOENOVH5CPOUNR7NCMS
[2]: https://src.fedoraproject.org/rpms/iscsi-initiator-utils/c/1e689cd0c6667eca838c85975a1b7a070209e5ad
[3]: https://src.fedoraproject.org/rpms/fedora-release/pull-request/246
[4]: https://github.com/coreos/fedora-coreos-config/blob/1553518214088a89d6a2360a6fcdddbd3915628a/manifests/ignition-and-ostree.yaml#L35-L44
[5]: https://bugzilla.redhat.com/show_bug.cgi?id=1930458

(cherry picked from commit b5c5a05853258678dca2b00e7a81ccdccf6700e3)

---

overlay: stop disabling `iscsid.socket`

The iSCSI daemon is now socket-activated so that it's only running when
needed rather than always enabled. We're breaking that by disabling
`iscsid.socket`.

This effectively reverts 929ac48 ("c9s: Disable iscsid.socket"). It's
not certain why this was done, but it was likely to workaround failing
tests. These tests should be fixed now[[1]] so we should be able to stop
doing this.

[1]: https://github.com/coreos/coreos-assembler/pull/3275

(cherry picked from commit 8c9427029cd6da75756a797f04d85c51caebd4ec)

---

tests/kola: delete ext.config.systemd.network-online test

This was moved to FCOS[[1]].

[1]: https://github.com/coreos/fedora-coreos-config/pull/2437

(cherry picked from commit 096f0aed32d3a7e43daeee6d176b93521da9ee2f)

---

Bump fedora-coreos-config submodule

See: https://github.com/coreos/fedora-coreos-config/pull/2448

```
$ git -C fedora-coreos-config shortlog --no-merges \
    9fae403ff93a090ef4f7436eda8a0d5387b9c862..ed7f4f22c6db1fbcabad678267a747cc70cbc53a
Dusty Mabe (1):
      tests/kola: wait longer in commonlib.sh is_service_active

Jonathan Lebon (1):
      tests/kola: upstream network-online login test from RHCOS


cgwalters (approved):
/lgtm

cgwalters:
Hm interesting, the newly added test is failing.

travier:
/hold
(waiting on f-c-c backports to land and be added to this RP)

dustymabe (approved):
Trivial LGTM

/lgtm

Issue #1099, opened 2022-12-13
Fix rt kernel test to work across versions


cgwalters:
Our hardcoded kernel versions already went away:

https://github.com/openshift/os/pull/951#issuecomment-1349511670

Need to either copy hardcoded kernel versions to a separate rpm repo (or stick them in a container which we extract - probably better) or make the version scraping dynamic.

PR #1306, opened 2023-06-05
kola/rpm-ostree/replace-rt-kernel: do not hardcode kernel versions

jmarrero:
closes https://github.com/openshift/os/issues/1099

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -14,9 +14,6 @@
   arches:
   - ppc64le
 
-- pattern: ext.config.rpm-ostree.replace-rt-kernel
-  tracker: https://github.com/openshift/os/issues/1099
-
 - pattern: iso-live-login.uefi-secure
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
--- a/tests/kola/rpm-ostree/replace-rt-kernel
+++ b/tests/kola/rpm-ostree/replace-rt-kernel
@@ -21,17 +21,18 @@ baseurl=
 case "${major}" in
     8)
         # TODO: why the heck does centos only support insecure http?!?
+        # TODO: Fix centos8 path
         baseurl=http://mirror.centos.org/centos/8-stream/
-        # TODO avoid hardcoding versions
-        target_kver=4.18.0-383.el8
-        target_kver_rt=4.18.0-394.rt7.179.el8
+        repo_url=??
+        target_kver=$(runv curl -sSLf ${baseurl}BaseOS/${basearch}/os/Packages/ | runv grep -o 'href=".*">' | runv grep -Po '(?<=kernel-core-).*?\.el8' | runv head -1)
+        target_kver_rt=$(runv curl -sSLf ${baseurl}RT/${basearch}/os/Packages/ | runv grep -o 'href=".*">' | runv grep -Po '(?<=kernel-rt-core-).*?\.el8' | runv tail -1)
         ;;
     9)
         # TODO: why the heck does centos only support insecure http?!?
         baseurl=http://mirror.stream.centos.org/9-stream/
-        # TODO avoid hardcoding versions
-        target_kver=5.14.0-196.el9
-        target_kver_rt=5.14.0-201.rt14.202.el9
+        repo_url=https://raw.githubusercontent.com/openshift/os/bf4383e4929ad9abb67d508a7ffa7cf45e8de5ca/c9s.repo
+        target_kver=$(runv curl -sSLf ${baseurl}BaseOS/${basearch}/os/Packages/ | runv grep -o 'href=".*">' | runv grep -Po '(?<=kernel-core-).*?\.el9' | runv head -1)
+        target_kver_rt=$(runv curl -sSLf ${baseurl}RT/${basearch}/os/Packages/ | runv grep -o 'href=".*">' | runv grep -Po '(?<=kernel-rt-core-).*?\.el9' | runv tail -1)
         ;;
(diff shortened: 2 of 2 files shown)


jmarrero:
This script works:
```
#!/bin/bash

set -euo pipefail

basearch=$(arch)
baseurl=https://mirror.stream.centos.org/9-stream/
target_kver=$(curl --silent ${baseurl}BaseOS/${basearch}/os/Packages/ | grep -o 'href=".*">' | grep -Po '(?<=kernel-core-).*?(?=\.el)' | head -1)
target_kver_rt=$(curl --silent ${baseurl}RT/${basearch}/os/Packages/ | grep -o 'href=".*">' | grep -Po '(?<=kernel-rt-core-).*?(?=\.el)' | tail -1)

echo -e "my kernel versions are\n${target_kver}\n${target_kver_rt}"
```

But still figuring out what is wrong in Kola, I must be missing something obvious.

jmarrero:
So apparently kernel-rt now needs: https://mirror.stream.centos.org/9-stream/RT/x86_64/os/Packages/realtime-setup-2.2-6.el9.x86_64.rpm which pulls in a bunch of deps  (tuna,tuned,etc) ultimately also needs python. I am wondering if just loading the centos stream repos for the test instead of doing a curl calls for each dep makes more sense here. 
@dustymabe @cgwalters what do you guys think?

dustymabe:
> So apparently kernel-rt now needs: [mirror.stream.centos.org/9-stream/RT/x86_64/os/Packages/realtime-setup-2.2-6.el9.x86_64.rpm](https://mirror.stream.centos.org/9-stream/RT/x86_64/os/Packages/realtime-setup-2.2-6.el9.x86_64.rpm) which pulls in a bunch of deps (tuna,tuned,etc) ultimately also needs python. I am wondering if just loading the centos stream repos for the RT part of the test instead of doing a curl calls for each dep makes more sense for that part. For just testing a kernel downgrade we can use the curl calls. @dustymabe @cgwalters what do you guys think?

Maybe we can use the yum repo directly? Looks like support was added in https://github.com/coreos/rpm-ostree/pull/3636 and there is an example in:

https://github.com/coreos/rpm-ostree/blob/0c3df78222650cf4174333dfa2d2286903aae264/ci/test-container.sh#L118-L124

dustymabe:
> > > So apparently kernel-rt now needs: [mirror.stream.centos.org/9-stream/RT/x86_64/os/Packages/realtime-setup-2.2-6.el9.x86_64.rpm](https://mirror.stream.centos.org/9-stream/RT/x86_64/os/Packages/realtime-setup-2.2-6.el9.x86_64.rpm) which pulls in a bunch of deps (tuna,tuned,etc) ultimately also needs python. I am wondering if just loading the centos stream repos for the RT part of the test instead of doing a curl calls for each dep makes more sense for that part. For just testing a kernel downgrade we can use the curl calls. @dustymabe @cgwalters what do you guys think?
> > 
> > 
> > Maybe we can use the yum repo directly? Looks like support was added in [coreos/rpm-ostree#3636](https://github.com/coreos/rpm-ostree/pull/3636) and there is an example in:
> > [coreos/rpm-ostree@`0c3df78`/ci/test-container.sh#L118-L124](https://github.com/coreos/rpm-ostree/blob/0c3df78222650cf4174333dfa2d2286903aae264/ci/test-container.sh#L118-L124)
> 
> I might be missing the point on doing this for the test since we still need to download the repo.

Right. We need to download the repo file, but then we don't have to do any complicated parsing (curl, etc).

just curl the yum repo definition for CentOS stream and then `rpm-ostree override replace --experimental --from repo=reponame kernel` or something like that.


> 
> Also RHEL8 is proving a pain to test, I get OOM on my VM. Kola gives me
> 
> ```
> Error: Parsing config.ign: Butane config has unrecognized spec version and prerelease is already set to "experimental"; cannot work around
> 2023-06-06T21:00:58Z cli: Parsing config.ign: Butane config has unrecognized spec version and prerelease is already set to "experimental"; cannot work around
> error: failed to execute cmd-kola: exit status 1
> ```
> 
> Was the request of this rooted on a specific need? OK if we make this RHEL9 only for the first pass?

The original PR was here: https://github.com/openshift/os/pull/1087

I'm fine with doing this for 4.14+ (i.e. new stuff) for now, but @cgwalters may have a different opinion.

kola-denylist.yaml
```diff
@@ -16,6 +16,8 @@
 
 - pattern: ext.config.rpm-ostree.replace-rt-kernel
   tracker: https://github.com/openshift/os/issues/1099
+  osversion:
+  - rhel-8.6
```
dustymabe, line 20: IIUC we don't sync the denylist across branches here like we do for FCOS (though I think we should probably do that in the future). I think that means a rhel-8.6 would never get attempted to run against this code (since 4.14 is RHEL9 only) and you can just delete this.

jmarrero:
> Right. We need to download the repo file, but then we don't have to do any complicated parsing (curl, etc).

Ahh in this case the complicated curl/parsing is to get the kernel-version and not hardcode it.

jmarrero:
Strange failure on 
```
rhelver=$(. /usr/lib/os-release && echo "${RHEL_VERSION}")
```
which works locally and manually on a VM. retrying...

travier:
It's fine for me if this is made RHEL 9 only

jmarrero:
Would not update the PR after I force pushed for some reason.... new PR here: https://github.com/openshift/os/pull/1309

jmarrero:
updated branch.

PR #1309, opened 2023-06-07
kola/rpm-ostree/replace-rt-kernel: do not hardcode kernel versions

jmarrero:
closes https://github.com/openshift/os/issues/1099

### DIFF
--- a/tests/kola/rpm-ostree/replace-rt-kernel
+++ b/tests/kola/rpm-ostree/replace-rt-kernel
@@ -2,6 +2,9 @@
 ## kola:
 ##   tags: "needs-internet"
 ##   timeoutMin: 30
+##   # We've seen some OOM when 1024M is used in similar tests:
+##   # https://github.com/coreos/fedora-coreos-tracker/issues/1506
+##   minMemory: 2048
 ##   description: Verify replacing the current kernel with an 
 ##     older centos kernel and replacing with kernel-rt.
 
@@ -15,68 +18,51 @@ runv () {
 }
 
 basearch=$(arch)
-rhelver=$(. /usr/lib/os-release && echo "${RHEL_VERSION}")
-major=$(echo "${rhelver}" | cut -f 1 -d .)
+major=$(. /usr/lib/os-release && echo "${CPE_NAME}" | grep -Eo '[0-9]{1,2}')
 baseurl=
 case "${major}" in
-    8)
-        # TODO: why the heck does centos only support insecure http?!?
-        baseurl=http://mirror.centos.org/centos/8-stream/
-        # TODO avoid hardcoding versions
-        target_kver=4.18.0-383.el8
-        target_kver_rt=4.18.0-394.rt7.179.el8
-        ;;
     9)
-        # TODO: why the heck does centos only support insecure http?!?
-        baseurl=http://mirror.stream.centos.org/9-stream/
-        # TODO avoid hardcoding versions
-        target_kver=5.14.0-196.el9
-        target_kver_rt=5.14.0-201.rt14.202.el9
+        baseurl=https://mirror.stream.centos.org/9-stream/
+        repo_url=https://raw.githubusercontent.com/openshift/os/master/c9s.repo
         ;;
-    *)  fatal "Unhandled RHEL_VERSION=${rhelver}"
+    *)  fatal "Unhandled major RHEL/SCOS VERSION=${major}"
         ;;
 esac
 
+# setup repos
+runv rm -rf /etc/yum.repos.d/*
+runv curl -sSLf "${repo_url}" -o /etc/yum.repos.d/cs.repo
+runv curl -sSLf https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
+runv sed -i 's|^gpgkey.*|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official|' /etc/yum.repos.d/cs.repo
+
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
 "")
-    echo "Testing overriding with previous CentOS Stream kernel"
-    kver=$(uname -r)
-    if test "${kver}" = "${target_kver}"; then
-        fatal "Already in ${target_kver}"
-    fi
-    tmpdir=$(mktemp -d -p /var/tmp)
-    pushd "${tmpdir}"
-    runv curl -sSLf --remote-name-all "${baseurl}/BaseOS/${basearch}/os/Packages/kernel"{,-core,-modules,-modules-extra}"-${target_kver}.${basearch}.rpm"
-    runv rpm-ostree override replace ./*.rpm
-    popd
-    rm "${tmpdir}" -rf
+    echo "Testing overriding with CentOS Stream kernel"
+    runv rpm-ostree override replace --experimental --from repo=baseos kernel{,-core,-modules,-modules-extra,-modules-core}
     runv /tmp/autopkgtest-reboot 1
     ;;
 1)
-    case $(uname -r) in
-        "${target_kver}.${basearch}")
+    case $(rpm -qi kernel-core | grep Vendor) in
+        *"CentOS")
             echo "ok kernel override"
             ;;
         *)
             runv uname -r
-            runv rpm -q kernel
-            fatal "Failed to apply kernel override to ${target_kver}"
+            runv rpm -qi kernel-core
+            fatal "Failed to apply kernel override"
             ;;
     esac
 
     echo "Testing overriding with CentOS Stream RT kernel"
     case $basearch in
         x86_64)
-            tmpdir=$(mktemp -d -p /var/tmp)
-            pushd "${tmpdir}"
             runv rpm-ostree override reset -a
-            runv curl -sSLf --remote-name-all "${baseurl}/NFV/${basearch}/os/Packages/kernel-rt"{-core,-modules,-modules-extra,-kvm}"-${target_kver_rt}.${basearch}.rpm"
+            kernel_pkgs=("kernel-rt-core" "kernel-rt-modules" "kernel-rt-modules-extra" "kernel-rt-modules-core")
             args=()
-            for x in ./*.rpm; do
-                args+=(--install ./"${x}")
+            for x in ${kernel_pkgs}; do
+                args+=(--install "${x}")
             done
-            runv rpm-ostree override remove kernel{,-core,-modules,-modules-extra} "${args[@]}"
-            rm "${tmpdir}" -rf
+            runv rpm-ostree override remove kernel{,-core,-modules,-modules-extra,-modules-core} "${args[@]}"
             runv /tmp/autopkgtest-reboot 2
             ;;
         *) echo "note: no kernel-rt for $basearch"; exit 0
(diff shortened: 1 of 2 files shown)


cgwalters:
Continuing this thread https://github.com/openshift/os/pull/1306#issuecomment-1580948065 on this PR:

Agree eventually we should stop parsing HTML for the *previous kernel* test but doing that is going to require handling the "repoquery" stuff either in rpm-ostree or temporarily layering dnf or something else

As far as the `--from` thing for the kernel-rt override, I *think* today that that is actually a no-op in this case and we'll just use the latest from whatever repo - we're not repository-locking.  It seems actually a bit more confusing to me to provide it because it won't do anything.

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -69,13 +65,14 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
...
+            runv curl -L https://centos.org/keys/RPM-GPG-KEY-CentOS-Official > /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
+            runv sed -i 's|^gpgkey.*|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official|' /etc/yum.repos.d/cs.repo
             runv rpm-ostree override reset -a
-            runv curl -sSLf --remote-name-all "${baseurl}/NFV/${basearch}/os/Packages/kernel-rt"{-core,-modules,-modules-extra,-kvm}"-${target_kver_rt}.${basearch}.rpm"
-            args=()
-            for x in ./*.rpm; do
-                args+=(--install ./"${x}")
-            done
-            runv rpm-ostree override remove kernel{,-core,-modules,-modules-extra} "${args[@]}"
+            args=(--install kernel-rt)
```
cgwalters, line 74: We don't actually do `kernel-rt` in OCP because it's a metapackage that drags in too much stuff, see https://github.com/openshift/machine-config-operator/blob/dadba5bda3ad2051050a93e7dd0ea9e7f1324b21/pkg/daemon/update.go#L1174

So I'd probably copy that for now.

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -19,64 +22,54 @@ rhelver=$(. /usr/lib/os-release && echo "${RHEL_VERSION}")
...
-    fi
-    tmpdir=$(mktemp -d -p /var/tmp)
-    pushd "${tmpdir}"
-    runv curl -sSLf --remote-name-all "${baseurl}/BaseOS/${basearch}/os/Packages/kernel"{,-core,-modules,-modules-extra}"-${target_kver}.${basearch}.rpm"
-    runv rpm-ostree override replace ./*.rpm
-    popd
-    rm "${tmpdir}" -rf
+    runv rm -rf /etc/yum.repos.d/*
+    runv curl -v "${repo_url}" -o /etc/yum.repos.d/cs.repo
+    runv curl -v https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
```
cgwalters, line 45: This isn't a big deal right now, but in the general case you always want to use `curl -L` to follow redirects (here and everywhere). 

And we also want `-f` to fail if there's a HTTP error (otherwise we'll output the error HTML successfully into what should be a GPG key).

We had these flags in the `curl -sSLf` that was removed below.

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -19,64 +22,49 @@ rhelver=$(. /usr/lib/os-release && echo "${RHEL_VERSION}")
...
 
+# setup repos
+runv rm -rf /etc/yum.repos.d/*
+runv curl -sSLf "${repo_url}" -o /etc/yum.repos.d/cs.repo
+runv curl -sSLf https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
+runv sed -i 's|^gpgkey.*|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official|' /etc/yum.repos.d/cs.repo
+
 case "${AUTOPKGTEST_REBOOT_MARK:-}" in
 "")
     echo "Testing overriding with previous CentOS Stream kernel"
```
dustymabe, line 42: minor
```suggestion
    echo "Testing overriding with CentOS Stream kernel"
```

dustymabe (approved):
LGTM

tests/kola/rpm-ostree/replace-rt-kernel
jmarrero: fixed.

dustymabe (approved):
/lgtm

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -15,68 +18,52 @@ runv () {
...
-        target_kver=4.18.0-383.el8
-        target_kver_rt=4.18.0-394.rt7.179.el8
-        ;;
     9)
         # TODO: why the heck does centos only support insecure http?!?
         baseurl=http://mirror.stream.centos.org/9-stream/
-        # TODO avoid hardcoding versions
-        target_kver=5.14.0-196.el9
-        target_kver_rt=5.14.0-201.rt14.202.el9
+        repo_url=https://raw.githubusercontent.com/openshift/os/bf4383e4929ad9abb67d508a7ffa7cf45e8de5ca/c9s.repo
```
LorbusChris, line 27: Is this specific commit a random one? Should this be https://raw.githubusercontent.com/openshift/os/master/c9s.repo instead?

jmarrero: It should, fixed.

jmarrero:
I tested this manually on SCOS and can't get it to fail. If this fails again might try even more memory.

LorbusChris:
> I tested this manually on SCOS and can't get it to fail. If this fails again might try even more memory.

I'm also thinking this is a memory issue. In the SCOS community builds we give the pods running COSA 4gb ram and 1 full CPU.

travier:
Let's bump this to 4GB here for this test only, master only as we don't run SCOS on other branches: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/os/openshift-os-master.yaml#L78

jmarrero:
lowering to 2GB and making this RHCOS only for now. I'll add this test for SCOS to my backlog.

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -2,6 +2,10 @@
 ## kola:
 ##   tags: "needs-internet"
 ##   timeoutMin: 30
+##   # We've seen some OOM when 1024M is used in similar tests:
+##   # https://github.com/coreos/fedora-coreos-tracker/issues/1506
+##   minMemory: 2048
+##   distros: rhcos
```
travier, line 8: We don't have an scos switch here, `rhcos` is aliased to `scos`. You'll have to keep the denylist entry above but just for SCOS so `c9s`.

dustymabe:
what failures are we seeing with SCOS? Could it be that there is nothing to override because we are now using the latest version of the kernel in c9s and it matches the already baked in kernel in the OSTree?

tests/kola/rpm-ostree/replace-rt-kernel
dustymabe: It would be nice if we could drop the denylist entry.. Maybe we use [`is_scos()`](https://github.com/coreos/fedora-coreos-config/blob/8807c8d9b3be1a6958bdb7016c70e5bdcf15bec6/tests/kola/data/commonlib.sh#L62-L66) inside the test?

If the reason it's failing is because of https://github.com/openshift/os/pull/1309#issuecomment-1587660800 then we could just do the rt test on SCOS, which is still a pretty good test.

jmarrero:
```
Jun 12 14:01:32 qemu0 kola-runext-replace-rt-kernel[2030]: Resolving dependencies...done
Jun 12 14:01:32 qemu0 kola-runext-replace-rt-kernel[2030]: error: Bus owner changed, aborting. This likely means the daemon crashed; check logs with `journalctl -xe`.
```
`error: Bus owner changed, aborting.` is the error. If it was something with the repos I can tweak the test easily. Sadly this error I have seen it in low memory scenarios but for some reason even with 4GB I am seeing the test fail like this on SCOS.

tests/kola/rpm-ostree/replace-rt-kernel
```diff
@@ -15,68 +18,51 @@ runv () {
...
-    *)  fatal "Unhandled RHEL_VERSION=${rhelver}"
+    *)  fatal "Unhandled major RHEL/SCOS VERSION=${major}"
         ;;
 esac
 
+# setup repos
+runv rm -rf /etc/yum.repos.d/*
+runv curl -sSLf "${repo_url}" -o /etc/yum.repos.d/cs.repo
+runv curl -sSLf https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
+runv sed -i 's|^gpgkey.*|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official|' /etc/yum.repos.d/cs.repo
```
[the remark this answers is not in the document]
LorbusChris: To me it looks like rpm-ostree may be failing when trying to verify the repos in c9s.repo. The keys for all of them have to be available. Blanket replacing all of them with one key won't work.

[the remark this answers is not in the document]
dustymabe: hmm. but this test as written passes on RHCOS (@jmarrero, correct me if I am wrong).

According to @jmarrero it fails on SCOS though and we aren't sure why yet.

dustymabe:
> `error: Bus owner changed, aborting.` is the error.

What dod the journal logs say?

tests/kola/rpm-ostree/replace-rt-kernel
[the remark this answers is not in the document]
LorbusChris: And we do have `repo_gpgcheck=0` on the repos in c9s.repo. I might've pulled a false alarm here.

LorbusChris:
My previous hold was probably a false alarm, but I'm still unsure why this fails on SCOS
/hold cancel

jmarrero:
I opened https://github.com/openshift/os/issues/1315 to investigate the SCOS failure, CI is not saving the journal for this test (unless I am looking in the wrong place, there is a link to the logs in the issue). On a SCOS VM running the test manually I could not get it to fail, so I need to build scos and run the test against it locally to see if it fails and debug further.

tests/kola/rpm-ostree/replace-rt-kernel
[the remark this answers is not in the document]
jmarrero: Correct this passes on RHCOS both on CI & locally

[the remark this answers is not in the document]
cgwalters: The problem here is that those reference `distribution-gpg-keys` which we don't ship in the OS by default.  I think @LorbusChris 's suggestion makes sense...

But I don't understand actually why we're removing the repo files by default?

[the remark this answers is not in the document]
jmarrero: It is because this is a test for RHCOS only at the moment, I am just making sure there nothing on yum.repos.d before downloading the centos repo. When I get to the SCOS test, the plan is to just use the provided repos/keys and just turn on the rt repo if it's disabled.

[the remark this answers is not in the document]
LorbusChris: see also https://github.com/openshift/os/pull/1316

PR #1313, opened 2023-06-09
extensions/Dockerfile: Use fedora:38 as builder

LorbusChris:


extensions/Dockerfile
```diff
@@ -20,7 +20,7 @@ RUN if [[ -n "${VARIANT}" ]]; then MANIFEST="manifest-${VARIANT}.yaml"; EXTENSIO
 ## Creates the repo metadata for the extensions & builds the go binary.
 ## This uses Fedora as a lowest-common-denominator because it will work on
 ## current p8/s390x.  See https://github.com/openshift/os/issues/1000
-FROM quay.io/fedora/fedora:37 as builder
+FROM quay.io/fedora/fedora:latest as builder
```
travier, line 23:
```suggestion
FROM quay.io/fedora/fedora:38 as builder
```

travier (commented):
We keep things to a fixed release so that it sticks for release branches

extensions/Dockerfile
LorbusChris: Makes sense. Updated

travier (approved):
/lgtm

Issue #1274, opened 2023-05-08
RHCOS 4.13-9.2 ppc64le will occassionally fail running the %pre scriptlets for openvswitch


mike-nguyen:
These errors are showing up during the rpm-ostree compose:
```
14:30:05  openvswitch3.1.prein: usermod.rpmostreesave: /etc/passwd.6: lock file already used
14:30:05  openvswitch3.1.prein: usermod.rpmostreesave: cannot lock /etc/passwd; try again later.
```

This results in worker nodes not coming up in a cluster install.

mike-nguyen:
This was addressed by: https://github.com/openshift/os/pull/1275. It only prevents us from shipping with broken groups for openvswitch

cgwalters:
We have a new problem now which is is rhel9 (?) builds picked up https://src.fedoraproject.org/rpms/openvswitch/c/a17c9d439da4f7e3bfec0ce4c3b178232d28d3fb?branch=rawhide it sounds like, and I believe today the problem is that use of sysusers.d just clashes with our previously hardcoded bits here https://github.com/openshift/os/blob/master/passwd#L27 and most importantly here: https://github.com/openshift/os/blob/1f2c0eb7e370d2412db15fa28556f419ddf73c5d/group#L45 (note openvswitch has no groups).

Basically we can do one of two things:

- Have the user *and* group data injected via sysusers
- Hardcode the user data in `/usr`

But it doesn't make sense to do both.  At this point, we could try dropping the hardcoded user/group files from this repo and rely on sysusers (i.e. per machine state).  *Or* we could hardcode the hugetlbfs group.

Now honestly, I think the *real* fix here is to move openvswithch to use `DynamicUser=yes` and open the hugetlbfs bits as an earlier privileged operation instead of relying on group access.

dcbw:
> _Or_ we could hardcode the hugetlbfs group.

@cgwalters note that openvswitch being in hugetlbfs *only* happens on x86-64 and ARM. Not on PPC or S390.

dcbw:
@cgwalters I'm curious what's actually clashing here though. sysusers.d(5) says it'll do the things it's asked if the group/user doesn't exist yet:

```
       g
           Create a system group of the specified name should it not exist yet. Note that u implicitly creates a matching group. The group will be created with no password set.

       m
           Add a user to a group. If the user or group do not exist yet, they will be implicitly created.
```

[...]

PR #1317, opened 2023-06-17
drop hardcoded openvswitch user/group definitions

dustymabe:
The RPM is now using systemd-sysusers fragments [1] so we can drop the hardcoded definitions. One problem here, though, is that the hugetlbfs group never gets added to the openswitch user [2] so let's add a workaround for that for now.

One side effect of this is that it does change the previously defined UID/GIDs from 800/801 to different values (dynamically generated at build time). I assume this is OK because of some of the discussion in [1].

[1] https://github.com/openshift/os/issues/1274#issuecomment-1593690794
[2] https://github.com/openshift/os/issues/1274#issuecomment-1595860275

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -189,6 +189,16 @@ postprocess:
      # but we have containers that expect it to be mounted so for now let's continue
      # generating it.
      ln -sr /usr/share/zoneinfo/UTC /etc/localtime
+  - |
+     #!/usr/bin/env bash
+     set -xeo pipefail
+     # Add the hugetlbfs group to the openvswitch user if the openvswitch-hugetlbfs.conf
+     # sysusers fragment exists. This is a workaround for a bug somewhere in the stack
+     # that we need to investigate further.
+     # https://github.com/openshift/os/issues/1274#issuecomment-1595860275
+     if [ -f /usr/lib/sysusers.d/openvswitch-hugetlbfs.conf ]; then
+         usermod -a -G hugetlbfs openvswitch
+     fi
 
 remove-files:
   # We don't ship man(1) or info(1)
--- a/group
+++ b/group
@@ -42,5 +42,3 @@ nfsnobody:x:65534:
 kube:x:994:
 sshd:x:74:
 chrony:x:992:
-openvswitch:x:800:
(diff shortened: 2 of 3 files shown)


dustymabe:
> One side effect of this is that it does change the previously defined UID/GIDs from 800/801 to different values (dynamically generated at build time). I assume this is OK because of some of the discussion in [1].

If this isn't desirable I think we can simply add back in the hardcoding and leave the workaround script in place and it will still solve the problem.

dustymabe:
looks like the version of openvswitch in CentOS stream isn't quite up to date with what is in RHEL so the tests are going to fail:

```
$ curl https://mirror.stream.centos.org/SIGs/9-stream/nfv/x86_64/openvswitch-2/Packages/o/ 2>/dev/null | grep -o -P 'openvswitch3.1-3.1.*?.rpm'  | uniq
openvswitch3.1-3.1.0-3.el9s.x86_64.rpm
openvswitch3.1-3.1.0-17.el9s.x86_64.rpm
```



> > 
> 
> If this isn't desirable I think we can simply add back in the hardcoding and leave the workaround script in place and it will still solve the problem.

I guess I'll open that PR too.

dustymabe:
> looks like the version of openvswitch in CentOS stream isn't quite up to date with what is in RHEL so the tests are going to fail:
> 
> ```
> $ curl https://mirror.stream.centos.org/SIGs/9-stream/nfv/x86_64/openvswitch-2/Packages/o/ 2>/dev/null | grep -o -P 'openvswitch3.1-3.1.*?.rpm'  | uniq
> openvswitch3.1-3.1.0-3.el9s.x86_64.rpm
> openvswitch3.1-3.1.0-17.el9s.x86_64.rpm
> ```
> 
> > > 
> > 
> > 
> > If this isn't desirable I think we can simply add back in the hardcoding and leave the workaround script in place and it will still solve the problem.
> 
> I guess I'll open that PR too.

https://github.com/openshift/os/pull/1318

dustymabe:
ok nevermind I just had this in the wrong manifest that applied to RHEL only.

PR #1318, opened 2023-06-18
add usermod workaround for openvswitch hugetlbfs group

dustymabe:
The RPM is now using systemd-sysusers fragments [1] and the RPM scriptlets no longer successfully add the `hugetlbfs` group to the `openvswitch` user [2]. Let's add a workaround for now while we investigate.

[1] https://github.com/openshift/os/issues/1274#issuecomment-1593690794
[2] https://github.com/openshift/os/issues/1274#issuecomment-1595860275


HuijingHei:
`ext.config.files.openvswitch-hugetlbfs-groups` is failed on c9s, maybe it also needs the workaround?
```
    --- FAIL: non-exclusive-test-bucket-0/ext.config.files.openvswitch-hugetlbfs-groups (2.78s)
            cluster.go:162: Error: Unit kola-runext-9.service exited with code 1
```

dustymabe:
> `ext.config.files.openvswitch-hugetlbfs-groups` is failed on c9s, maybe it also needs the workaround?

ahh. yeah I didn't add it to a shared manifest. Just pushed up a commit moving it to a shared manifest so hopefully it works now.

c4rt0 (commented):
LGTM

cgwalters (approved):

LorbusChris:
It looks like https://src.fedoraproject.org/rpms/systemd/blob/rawhide/f/sysusers.generate-pre.sh and https://pkgs.devel.redhat.com/cgit/rpms/systemd/tree/sysusers.generate-pre.sh?h=rhel-9.2.0 have diverged, and the latter is missing multiple updates to the script.

Specifically, this change that is missing in RHEL looks like the culprit to me: https://src.fedoraproject.org/rpms/systemd/c/f27d461663bec17ad64422682f260f0020ccc7f7?branch=rawhide

HuijingHei:
Have no idea how to debug the repo error:
```
error: Updating rpm-md repo 'rhel-9.2-server-ose-4.14': cannot update repo 'rhel-9.2-server-ose-4.14': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried; Last error: Curl error (28): Timeout was reached for http://base-4-14-rhel92.ocp.svc.cluster.local/rhel-9.2-server-ose-4.14/repodata/repomd.xml [Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds]
error: failed to execute cmd-fetch: exit status 1
```

dustymabe:
> http://base-4-14-rhel92.ocp.svc.cluster.local

I just don't know why it's trying a repo in a `svc.cluster.local` domain. Maybe that's the problem?

cgwalters:
> I just don't know why it's trying a repo in a svc.cluster.local domain.

It's to avoid entitling the Prow hosts; we pull from a proxy mirror today.

PR #1322, opened 2023-06-23
OCPBUGS-15375: extensions: remove unused code as MCO directly uses image

jmarrero:


mike-nguyen:
/retitle OCPBUGS-15375: extensions: remove unused code as MCO directly uses image

PR #1326, opened 2023-06-26
Enable sandboxed-container extension on s390x

danilocesar:
The next steps from the OpenSynergy group is to enable s390x builds for sandoboxed-container extension as part of the plan to support PeerPods on s390x.

Fixes: #1206


danilocesar:
@mike-nguyen and @cgwalters can you take a look at this? We halted the work on #1207 and Eduardo left the group, so I'm picking up the effort to make it happen for the next release. Could you please review it?

PR #1327, opened 2023-07-05
extensions: add crun-wasm to enable wasm support in OCP

sohankunkerkar:
This package enables wasm-supported workloads in OCP. 

xref: https://issues.redhat.com/browse/RFE-4177

### DIFF
--- a/extensions-c9s.yaml
+++ b/extensions-c9s.yaml
@@ -6,6 +6,10 @@ repos:
   - sig-virtualization
 
 extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  wasm:
+    packages:
+      - crun-wasm
   # https://github.com/coreos/fedora-coreos-tracker/issues/1504
   ipsec:
     packages:
--- a/extensions-rhel-9.2.yaml
+++ b/extensions-rhel-9.2.yaml
@@ -3,6 +3,10 @@
 # and https://github.com/coreos/fedora-coreos-tracker/issues/401
 
 extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  wasm:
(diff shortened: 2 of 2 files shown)


extensions-c9s.yaml
```diff
@@ -6,6 +6,10 @@ repos:
   - sig-virtualization
 
 extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  crun-wasm:
```
haircommander, line 10: should be wasm per update to https://github.com/openshift/machine-config-operator/pull/3776/files

sohankunkerkar: Fixed

jmarrero (approved):

marmijo:
After this merged, `s390x` and `ppc64le` builds started failing in 4.14-9.2. It looks like the crun-wasm package is only built for `x86_64` and `aarch64`, so I opened a PR to limit the new package to only those two arches.
https://github.com/openshift/os/pull/1329

LorbusChris:
This also makes SCOS builds failing, as `crun-wasm` isn't available in any of the specified repos. Is it available elsewhere for C9S @sohankunkerkar ?

sohankunkerkar:
> This also makes SCOS builds fail as `crun-wasm` isn't available in any of the specified repos. Is it available elsewhere for C9S @sohankunkerkar ?

Oh, my bad. It's not available for C9S. We have specifically built it for rhel-9 only: https://brewweb.engineering.redhat.com/brew/buildinfo?buildID=2561445

sohankunkerkar:
@LorbusChris FYI: https://github.com/openshift/os/pull/1330

PR #1329, opened 2023-07-06
extensions: limit crun-wasm package to x86_64 and aarch64

marmijo:
The `crun-wasm` package is [built for only these two arches](https://brewweb.engineering.redhat.com/brew/packageinfo?packageID=84555). Specify `x86_64` and `aarch64` only so that `s390x` and `ppc64le` can be unblocked.

xref: https://github.com/openshift/os/pull/1327


cgwalters (approved):
/lgtm

PR #1330, opened 2023-07-06
extensions: remove wasm extension from C9S

sohankunkerkar:
`crun-wasm` isn't built for C9S, so skip this to avoid SCOS build failure.

### DIFF
--- a/extensions-c9s.yaml
+++ b/extensions-c9s.yaml
@@ -7,12 +7,14 @@ repos:
 
 extensions:
   # https://issues.redhat.com/browse/RFE-4177
-  wasm:
-    architectures:
-      - x86_64
-      - aarch64
-    packages:
-      - crun-wasm
+  # wasm:
+  #   architectures:
+  #     - x86_64
+  #     - aarch64
+  #   repos:
+  #     - appstream
+  #   packages:
+  #     - crun-wasm
   # https://github.com/coreos/fedora-coreos-tracker/issues/1504
   ipsec:
     packages:


LorbusChris:
Thanks for filing this!

Please do something like https://github.com/openshift/os/commit/aa88c1cba02c888627b1d95ffd6eea9e541be5a8.
and ideally also open a ticket to get `crun-wasm` built for C9S asap.

/hold

sohankunkerkar:
> Thanks for filing this!
> 
> Please do something like [aa88c1c](https://github.com/openshift/os/commit/aa88c1cba02c888627b1d95ffd6eea9e541be5a8). and ideally also open a ticket to get `crun-wasm` built for C9S asap.
> 
Sure, but `crun-wasm` isn't there in RHEL9 yet and the package is built specifically for OCP users.

cgwalters:
Yeah, it'd need to be built by an OCP process.  But I think what we're running straight into here is that basically building RPMs and OCP stinks; there's tons of automation around container images that are shared between CI/ART but not rpms.  

Long term what I think would make the most sense is to have gitlab.com/redhat/ocp/rpms with infrastructure that mirrors that on gitlab.com/redhat/centos-stream or so.  (But this runs into a lot of interesting open questions around integrating Prow and our CI with gitlab...)

LorbusChris:
I've been doing builds of those RPMs (i.e. ones not in Stream/RHEL) on https://cbs.centos.org through the CentOS Cloud SIG, but manual maintenance is obviously not ideal.

I have opened RFEs to get support for packit.dev build automation on both the CBS and the Stream koji instances:
- https://github.com/packit/packit/issues/1779
- https://github.com/packit/packit/issues/2002

The Packit team is looking to prioritize these, please leave a comment there if you'd find this useful too.

@sohankunkerkar could you comment the lines out instead of removing them for now until we have an RPM build?

extensions-c9s.yaml
```diff
@@ -6,13 +6,15 @@ repos:
...
 
 extensions:
-  # https://issues.redhat.com/browse/RFE-4177
-  wasm:
-    architectures:
-      - x86_64
-      - aarch64
-    packages:
-      - crun-wasm
+ # https://issues.redhat.com/browse/RFE-4177
```
travier, line 9:
```suggestion
  # https://issues.redhat.com/browse/RFE-4177
```

LorbusChris:
Both SCOS and RHCOS 9.2 build-test-qemu both timed out. Trying again
/retest

PR #1332, opened 2023-07-19
OCPBUGS-15843: Enable rhcos-fips-finish.service in diskless cases

zaneb:
Allow the user to enable FIPS by passing fips=1 on the kernel command line, even in diskless cases such as the live ISO.

The rhcos-fips.service (which reads the MachineConfig and sets up fips=1 on the kernel command line before triggering a reboot) works only once the OS is written to disk. The same is not true of rhcos-fips-finish.service, which runs after fips=1 is passed at boot time. The former is not needed if fips=1 is already passed on the kernel command line, but the latter is required to set up the crypto policies.

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
@@ -31,10 +31,9 @@ install() {
     # for a true FIPS boot: https://bugzilla.redhat.com/show_bug.cgi?id=1778940
     echo "# RHCOS FIPS mode installation complete" > "$initdir/etc/system-fips"
 
-    # We don't support FIPS in diskless cases currently
-    target=ignition-diskful.target
     # note we `|| exit 1` here so we error out if e.g. the units are missing
     # see https://github.com/coreos/fedora-coreos-config/issues/799
-    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips.service || exit 1
-    systemctl -q --root="$initdir" add-requires "$target" rhcos-fips-finish.service || exit 1
+    # We don't support reconfiguring the bootloader for FIPS in diskless cases
+    systemctl -q --root="$initdir" add-requires ignition-diskful.target rhcos-fips.service || exit 1
+    systemctl -q --root="$initdir" add-requires initrd.target rhcos-fips-finish.service || exit 1
 }
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh
@@ -65,6 +65,10 @@ firstboot() {
 }
 
 finish() {
+    if [ "$(</proc/sys/crypto/fips_enabled)" -ne 1 ]; then
+        fatal "FIPS mode is not enabled."
+    fi
+
     # This is analogous to Anaconda's `chroot /sysroot fips-mode-setup`. Though
     # of course, since our approach is "Ignition replaces Anaconda", we have to
     # do it on firstboot ourselves. The key part here is that we do this


overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/module-setup.sh
```diff
@@ -36,5 +36,5 @@ install() {
     # note we `|| exit 1` here so we error out if e.g. the units are missing
     # see https://github.com/coreos/fedora-coreos-config/issues/799
     systemctl -q --root="$initdir" add-requires "$target" rhcos-fips.service || exit 1
```
travier, line 38: Can you inline the variable from 2 lines above and remove the now stale comment?

travier:
This will skip https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh#L67. Don't we need that to effectively enable FIPS mode in the Live rootfs?

cgwalters:
Yes.

Bigger picture, I'd like to discuss with the RHEL team about effectively folding this logic into the upstream [dracut fips module](https://github.com/dracutdevs/dracut/tree/master/modules.d/01fips).  Or at least, do it downstream.

Basically if we make the official interface to "enable fips" be "set fips=1 in the kernel cmdline" then we can also switch to enabling it in the initramfs in firstboot for the Anaconda case too.  Anaconda has [almost exact duplicate code](https://github.com/rhinstaller/anaconda/blob/1f93219a10ba1c5bb03711961a55499ddf1884c2/pyanaconda/modules/security/installation.py#L167) to run the `fips-mode-setup` before rebooting, but I believe we can just consistently enable in the initramfs on the first boot.

overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips-finish.service
```diff
@@ -3,7 +3,7 @@ Description=Finish FIPS mode setup
 DefaultDependencies=false
 # we run in the second boot after the one where we set fips=1
 ConditionKernelCommandLine=ignition.firstboot
-ConditionKernelCommandLine=fips
+ConditionKernelCommandLine=fips=1
```
cgwalters, line 6: I guess...but it's not clear to me that we want to support any setup other than (in Rust-like notation):

- `None` (value not present)
- `Some(fips=1)`

It's not supported to set `fips=0` as far as I know.  So I'd argue perhaps we should continue with `ConditionKernelCommandLine=fips`, but then in the code parse its value and abort if it has any value other than `1`.

zaneb: I put this in just to make sure that if somebody puts something unsupported like `fips=0` (or just `fips`) it won't have any effect here.

zaneb:
> This will skip https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh#L67. Don't we need that to effectively enable FIPS mode in the Live rootfs?

Can you clarify why it will be skipped? We're skipping it now (in the live ISO/PXE); the purpose of this patch is to not skip it any more (when the user has specified `fips=1`).

> Anaconda has [almost exact duplicate code](https://github.com/rhinstaller/anaconda/blob/1f93219a10ba1c5bb03711961a55499ddf1884c2/pyanaconda/modules/security/installation.py#L167) to run the fips-mode-setup before rebooting, but I believe we can just consistently enable in the initramfs on the first boot.

+1 (in the medium term). Right now there's at least 3 things that have to be in sync (kernel cmdline, /etc/crypto-policies, /etc/system-fips). Reducing that to 1 (that triggers the others) would be good for everybody.

cgwalters:
So one thing that would be a great followup to this is to add some kola test for fips, both in the diskful and diskless case.

/lgtm

cgwalters:
Failure is not related

PR #1339, opened 2023-08-04
Bump fedora-coreos-config

coreosbot-releng:

```
Aashish Radhakrishnan (1):
      denylist: Snooze ext.config.toolbox for rawhide

Colin Walters (2):
      initramfs: Ensure we `Requires=ostree-prepare-root.service`
      initramfs: Remove manual sysroot mounting

Dusty Mabe (12):
      overrides: drop systemd-253.4-1.fc38 pin
      overlay.d/05core: move coreos-rootflags to 35coreos-ignition dracut module
      overlay.d/05core: add coreos-rootflags back to diskful sysroot.mount
      overlay.d/05core: drop RHEL8 path from coreos-gpt-setup.sh
      denylist: drop denials for ntp tests
      denylist: snooze some networking tests in rawhide
      denylist: add branched, next-devel streams to denials for rawhide
      overlay.d/05core: drop Requires on root device for coreos-rootflags.service
      overlay.d/05core: simplify logic in diskful-generator
      overlay.d/05core: enable coreos-rootflags.service when needed
      overlay.d/05core: drop RHEL8 workarounds for live-generator
      overlay.d/05core: support booting ISO completely from RAM

Michael Armijo (3):
      overrides: unpin kernel-6.3.12-200.fc38
      overrides: pin kernel-6.3.12-200 on `ppc64le` only
      denylist: extend snooze for ext.config.root-reprovision.*
```


marmijo (approved):

jmarrero (approved):
/approve

mike-nguyen:
I don't think there is a way to make the bot rebase so I re-ran the workflow. It picked up the CI change.

dustymabe:
> So I assume we want the Live ISO from RAM feature here, but it's worth noting there's a nontrivial rework of the initramfs `sysroot.mount` handling that also got rolled in here, which in theory we don't need for RHEL9 today (it will just make some things easier to debug on failures). We'll need to be on guard for possible regressions.

+1 - Since we already had a 4.14 branch I was assuming the ISO from RAM stuff we'd backport to 4.14 independent of the other stuff.

PR #1346, opened 2023-08-24
azure-checkin: Order `Afterburn checkin` finish before `coreos-ignition-unique-boot`

HuijingHei:
There is race condition about `checkin` and `coreos-ignition- unique-boot`. Removal of the virtual CD might cause Unique Boot Filesystem service failed, as could not find `/dev/sr0`, so make sure checkin finished before we start it.

Fixes https://issues.redhat.com/browse/OCPBUGS-17643


travier:
/retitle OCPBUGS-17643: azure-checkin: Order Afterburn checkin finish before coreos-ignition-unique-boot
/lgtm

PR #1350, opened 2023-09-01
azure-checkin: order checkin starting after `coreos-ignition-unique-boot`

HuijingHei:
This reverts commit e6500062fc421e32a638b87da4caed50b1616481. 

Sorry for the confusion, and should do more testing before merged.

---

azure-checkin: order checkin starting after `coreos-ignition-unique-boot`

There is race condition about checkin and coreos-ignition-unique-boot.
Removal of the virtual CD might cause Unique Boot Filesystem
service failed, as could not find `/dev/sr0`, so need to wait until
after coreos-ignition-unique-boot.
Fixes https://issues.redhat.com/browse/OCPBUGS-17643

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
@@ -30,12 +30,6 @@ After=ignition-fetch.service
 # across reboots, so defer it until after the kargs reboot.
 After=coreos-kargs-reboot.service
 
-# There is race condition about checkin and coreos-ignition-unique-boot.
-# Removal of the virtual CD might cause Unique Boot Filesystem 
-# service failed, as could not find `/dev/sr0`, so make sure checkin 
-# finished before we start it.
-Before=coreos-ignition-unique-boot.service
-
 # Run before switching root.
 Before=initrd.target
 
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/30rhcos-afterburn-checkin/rhcos-afterburn-checkin.service
@@ -30,6 +30,12 @@ After=ignition-fetch.service
 # across reboots, so defer it until after the kargs reboot.
 After=coreos-kargs-reboot.service
 
+# There is race condition about checkin and coreos-ignition-unique-boot.
+# Removal of the virtual CD might cause Unique Boot Filesystem 
+# service failed, as could not find `/dev/sr0`, so need to wait until 
+# after coreos-ignition-unique-boot.
+After=coreos-ignition-unique-boot.service
+
 # Run before switching root.
 Before=initrd.target
 


HuijingHei:
I think this is safe to merge, but let me do more testing to give more confidence.

HuijingHei:
Need more testing with this.

/holdon

HuijingHei:
Testing result on Azure CVM:
@jinyunma  helps to create 2 clusters, all the nodes are managed to boot.
hhei create 3 clusters, all the nodes are managed to boot.

Through the testing results, I think the PR is ready to review.

HuijingHei:
cc @travier @cgwalters @dustymabe

travier:
That looks good to me as well. This should keep the disk available while the check happens and then we're good. We don't really need it but it's easier of devices stay in a consistent state while the check happens.

The device check should be fast enough that the order should not matter.

/lgtm

HuijingHei:
Thanks @travier for the review!

travier:
/retitle OCPBUGS-17643: azure-checkin: order checkin starting after `coreos-ignition-unique-boot`

cgwalters:
@HuijingHei try this code in cosa 

```
diff --git a/mantle/platform/qemu.go b/mantle/platform/qemu.go
index 086e12496..ab8620a2d 100644
--- a/mantle/platform/qemu.go
+++ b/mantle/platform/qemu.go
@@ -1838,6 +1838,22 @@ func (builder *QemuBuilder) Exec() (*QemuInstance, error) {
 		return nil, fmt.Errorf("failed to connect over qmp to qemu instance")
 	}
 
+	// Hacky code to test https://github.com/openshift/os/pull/1346
+	if timeout, ok := os.LookupEnv("COSA_TEST_CDROM_UNPLUG"); ok {
+		timeout, err := time.ParseDuration(timeout)
+		if err != nil {
+			return nil, err
+		}
+		go func() {
+			time.Sleep(timeout)
+			_, err := inst.qmpSocket.Run([]byte(`{ "execute": "eject", "arguments": { "device": "ide1-cd0", "force": true } }`))
+			if err != nil {
+				plog.Errorf("failed to eject: %s", err)
+			}
+			plog.Info("ejected CD")
+		}()
+	}
+
 	return &inst, nil
 }
 
```

Then run it like

```
env COSA_TEST_CDROM_UNPLUG=30s cosa run -- -cdrom /path/to/some/testing.iso
```

Change `30s` here to be e.g. `1s` so the device gets unplugged early in boot.

cgwalters:
OK right, in chat one distinction here may be that in Azure the *physical device* goes away, as opposed to having the virtual media ejected.

Hmm, does it appear in Azure as a USB CD-ROM?  (a minute passes while I go to find a random journal log from an OCP cluster launch in Azure...)

```
Sep 07 22:01:39.559822 localhost kernel: sr 0:0:0:2: [sr0] scsi-1 drive
Sep 07 22:01:39.562958 localhost kernel: cdrom: Uniform CD-ROM driver Revision: 3.20
Sep 07 22:01:39.565860 localhost kernel: sr 0:0:0:2: Attached scsi CD-ROM sr0
```

OK right, it's SCSI.  I didn't think that could be hot unplugged but maybe I'm wrong?  Also indeed I still see `sr0` even on subsequent boots.

So actually, now I'm confused how `/dev/sr0` is going away temporarily; what is causing that?

travier:
If the disk is unplugged / ejected then you likely won't find it after the initial boot / after the check in.

HuijingHei:
> If the disk is unplugged / ejected then you likely won't find it after the initial boot / after the check in.

Got it, thanks!

Issue #1352, opened 2023-09-01
[4.11] Fedora 36 secondary architecture repos can't be reached any longer


aaradhak:
The 4.11 [builds](https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com/job/build-arch/1674/) are failing in s390x and ppc64le arches because the fedora36 image used in the tests can't install rpms anymore. It looks like the the secondary arch repos didn't get moved into the right place at EOL or maybe the metalink is pointing to the wrong place for them. We are using the container from the registry registry.fedoraproject.org/fedora:36

```
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]: Fedora 36 - ppc64le                             207  B/s | 196  B     00:00
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]: Errors during downloading metadata for repository 'fedora':
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]:   - Status code: 404 for https://dl.fedoraproject.org/pub/fedora-secondary/releases/36/Everything/ppc64le/os/repodata/repomd.xml (IP: 38.145.60.22)
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]:   - Status code: 404 for https://d2lzkl7pfhq30w.cloudfront.net/pub/fedora-secondary/releases/36/Everything/ppc64le/os/repodata/repomd.xml (IP: 18.160.0.78)
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]:   - Status code: 404 for http://dl.fedoraproject.org/pub/fedora-secondary/releases/36/Everything/ppc64le/os/repodata/repomd.xml (IP: 38.145.60.23)
04:47:23  Aug 31 08:47:13 qemu0 kola-runext-dhcp-propagation[2087]: Error: Failed to download metadata for repo 'fedora': Cannot download repomd.xml: Cannot download repodata/repomd.xml: All mirrors were tried
04:47:23  Aug 31 08:47:14 qemu0 kola-runext-dhcp-propagation[2042]: Error: error building at STEP "RUN dnf -y install systemd dnsmasq iproute iputils && dnf clean all && systemctl enable dnsmasq": error while running runtime: exit status 1
[...]

PR #1343, opened 2023-08-16
[release-4.11] ci: use new redirector

mike-nguyen:
The old redirector is being decomissioned and there are firewall rules preventing access.

PR #1353, opened 2023-09-01
[release-4.11] denylist: skip ext.config.shared.chrony.dhcp-propagation

aaradhak:
This test is failing on s390x and ppc64le because the fedora36 image used in the tests can't install rpms anymore. Snooze failing test till 2023-09-22

Ref: https://github.com/openshift/os/issues/1352

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -27,3 +27,9 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
+- pattern: ext.config.shared.chrony.dhcp-propagation
+  tracker: https://github.com/openshift/os/issues/1352
+  snooze: 2023-09-22
+  arches:
+  - s390x
+  - ppc64le


kola-denylist.yaml
```diff
@@ -27,3 +27,8 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
+- pattern: ext.config.shared.chrony.dhcp-propagation
+  tracker: https://github.com/openshift/os/issues/1352
+  arches:
+  - s390x
+  - ppc64le
```
dustymabe, line 34: maybe make this a snooze so we'll get reminded to check back in on it?

dustymabe:
CI might not work until https://github.com/openshift/os/pull/1343 is merged - cc @mike-nguyen

kola-denylist.yaml
aaradhak: Right, I forgot about the snooze part

kola-denylist.yaml
```diff
@@ -27,3 +27,9 @@
   tracker: https://github.com/coreos/coreos-assembler/issues/2725
   arches:
   - ppc64le
+- pattern: ext.config.shared.chrony.dhcp-propagation
+  tracker: https://github.com/openshift/os/issues/1352
+  snooze: 2023-09-22
+  arches:
+  - s390x
+  - ppc64le
```
marmijo, line 35: Maybe we should add `warn: true` to this snooze as well.

dustymabe: Unfortunately we didn't really prioritize backporting that new feature to older branches of COSA so it won't do much. See https://github.com/coreos/coreos-assembler/pull/3581

marmijo: I see, thanks for the info! :+1:

aaradhak: Oh didnt know about this new feature.

jschintag:
Since this is only modifying the denylist, overriding CI

/override rhcos-86-build-test-metal
/override rhcos-86-build-test-qemu

travier:
/lgtm
The snooze makes sense as we expect this to be fixed with https://github.com/openshift/os/issues/1352

HuijingHei:
Try:
/override rhcos-86-build-test-metal
/override rhcos-86-build-test-qemu

travier:
@HuijingHei Note that my request for Mrunal is not about overriding the CI here, it's about the `jira/valid-bug` label requirement.

HuijingHei:
> @HuijingHei Note that my request for Mrunal is not about overriding the CI here, it's about the `jira/valid-bug` label requirement.

Sorry, I might misunderstood, does this mean we do not need to override the CI ?

jschintag:
As per the latest status of https://pagure.io/releng/issue/11655, the test works again. The old 4.11 build was successfully rebuild.
I guess this means  we can close this PR as it is no longer needed?

PR #1357, opened 2023-09-07
Bump fedora-coreos-config

coreosbot-releng:

```
Jan Schintag (1):
      kola secex tests: Increase test timeout to 5min.
```


jmarrero (approved):
/lgtm

jlebon:
This is blocking the pipeline and none of the Prow tests here test secex (the only change). Let's just get this in.

PR #1359, opened 2023-09-11
OCPBUGS-15843: initramfs/40rhcos-fips: Workaround dracut unmounting /boot

cgwalters:
We enabled FIPS in our ISO, but that triggers a new corner case because of how we handle `/boot` there.

Work around the lack of
https://github.com/dracutdevs/dracut/commit/ab26ad2c2ab4a5884e392951998d40829f130387 in RHEL 9.2.


cgwalters:
(So, so much painful debugging went into this...I was really fixated on the theory that this had something to do with the bwrap shenanigans happening in `rhcos-fips-finish.sh`, but no...)

jmarrero (approved):
/lgtm

travier:
Looks reasonable.
/lgtm

cgwalters:
Ah yes was just missing a 
`ConditionPathExists=/run/ostree-live`

PR #1367, opened 2023-09-15
image.yaml: aws; add old defaults for new fields

prestist:
see https://github.com/coreos/coreos-assembler/pull/3607

With the above PR, cosa's image-default.yaml adds 3 configuration fields. The defaults provided differ from before. Overide the new configuration fields to maintain old defaults.

### DIFF
--- a/image-rhel-9.2.yaml
+++ b/image-rhel-9.2.yaml
@@ -15,3 +15,9 @@ vmware-os-type: rhel8_64Guest
 # VMware hardware versions: https://kb.vmware.com/s/article/1003746
 # Supported VMware versions: https://lifecycle.vmware.com/
 vmware-hw-version: 15
+
+# see https://github.com/coreos/coreos-assembler/pull/3607
+# Defaults for AWS
+aws-imdsv2-only: false
+aws-volume-type: "gp2"
+aws-x86-boot-mode: "legacy-bios"
--- a/image-c9s.yaml
+++ b/image-c9s.yaml
@@ -18,3 +18,9 @@ vmware-os-type: rhel8_64Guest
 # VMware hardware versions: https://kb.vmware.com/s/article/1003746
 # Supported VMware versions: https://lifecycle.vmware.com/
 vmware-hw-version: 15
+
+# see https://github.com/coreos/coreos-assembler/pull/3607
+# Defaults for AWS
+aws-imdsv2-only: false
+aws-volume-type: "gp2"
+aws-x86-boot-mode: "legacy-bios"


image-rhel-9.2.yaml
```diff
@@ -15,3 +15,9 @@ vmware-os-type: rhel8_64Guest
 # VMware hardware versions: https://kb.vmware.com/s/article/1003746
 # Supported VMware versions: https://lifecycle.vmware.com/
 vmware-hw-version: 15
+
+# see https://github.com/coreos/coreos-assembler/pull/3607
+# Defaults for AWS
+aws-imdsv2-only: false
+aws-volume-type: "gp2"
+aws-x86-boot-mode: "uefi"
```
dustymabe, line 23: I think this is the wrong value:

```suggestion
aws-x86-boot-mode: "bios"
```

dustymabe (approved):
/lgtm

image-rhel-9.2.yaml
jlebon: I think the legacy behaviour is to not specify a boot mode parameter at all. According to https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-boot.html, this means that it'll be based on the instance type. I'm not sure if we set `legacy-bios` and try to boot it on a Graviton instance whether it'll break.

Probably cleaner anyway to do something similar to what we're doing for IMDSv2: https://github.com/coreos/coreos-assembler/pull/3607/files#diff-e051905e06e5eaa099fca1faa4b15ecfb54b08e54fe1763aecdaa30c77da4e34R378-R380 (i.e. leave it unset in the initial `params`, and only set the field if needed). Keeping the unset behaviour could be done here with e.g. `aws-x86-boot-mode: "default"`.

jlebon: Oh right, Graviton are the aarch64 instance types, so I think we're fine with `legacy-bios`, yup!

dustymabe:
But if you read the docs it says it defaults to bios for all x86 and uefi for aarch64 so I think this should be safe.

LorbusChris:
Do we need the same change in https://github.com/openshift/os/blob/master/image-c9s.yaml for C9S-based SCOS?

LorbusChris:
adding
/hold
until my question is answered

dustymabe:
> Do we need the same change in [`master`/image-c9s.yaml](https://github.com/openshift/os/blob/master/image-c9s.yaml?rgh-link-date=2023-09-15T23%3A15%3A06Z) for C9S-based SCOS?

I don't know if SCOS builds AMIs? Either way can't hurt to add. I'll add it but will skip the CI run since CI won't really test this change anyway.

PR #1372, opened 2023-09-25
rhsm: add `rhsm.conf` which is from `subscription-manager`

HuijingHei:
`rhsm.conf` (from `subscription-manager`) is missing on RHCOS, which causes `rpm-ostree install <pkg>` failed with entitlement build, as the SSL CA cert path is wrong to set
`/etc/rhsm-host-host/ca/redhat-uep.pem`.

```
STEP 3/5: RUN rpm-ostree install libreswan

error: Updating rpm-md repo 'rhel-9-for-x86_64-baseos-rpms': cannot update repo 'rhel-9-for-x86_64-baseos-rpms': Cannot download repomd.xml: Curl error (77): Problem with the SSL CA cert (path? access rights?) for https://cdn.redhat.com/content/dist/rhel9/9/x86_64/baseos/os/repodata/repomd.xml [error setting certificate file: /etc/rhsm-host-host/ca/redhat-uep.pem]
error: build error: building at STEP "RUN rpm-ostree install libreswan": while running runtime: exit status 1
```

https://issues.redhat.com/browse/RHEL-1451 meant to track from `librhsm` side, but was closed because of risky.

Fixes: https://issues.redhat.com/browse/OCPBUGS-11181

### DIFF
--- a/overlay.d/05rhcos/etc/rhsm/rhsm.conf
+++ b/overlay.d/05rhcos/etc/rhsm/rhsm.conf
@@ -1,0 +1,110 @@
+# Red Hat Subscription Manager Configuration File:
+
+# Unified Entitlement Platform Configuration
+[server]
+# Server hostname:
+hostname = subscription.rhsm.redhat.com
+
+# Server prefix:
+prefix = /subscription
+
+# Server port:
+port = 443
+
+# Set to 1 to disable certificate validation:
+insecure = 0
+
+# an http proxy server to use
+proxy_hostname =
+
+# The scheme to use for the proxy when updating repo definitions, if needed
+# e.g. http or https
+proxy_scheme = http
+
+# port for http proxy server
+proxy_port =
+
+# user name for authenticating to an http proxy, if needed
+proxy_user =
+
+# password for basic http proxy auth, if needed
+proxy_password =
+
+# host/domain suffix blocklist for proxy, if needed
+no_proxy =
+
+[rhsm]
+# Content base URL:
+baseurl = https://cdn.redhat.com
+
+# Repository metadata GPG key URL:
+repomd_gpg_url =
+
+# Server CA certificate location:
+ca_cert_dir = /etc/rhsm/ca/
+
+# Default CA cert to use when generating yum repo configs:
+repo_ca_cert = %(ca_cert_dir)sredhat-uep.pem
+
+# Where the certificates should be stored
+productCertDir = /etc/pki/product
+entitlementCertDir = /etc/pki/entitlement
+consumerCertDir = /etc/pki/consumer
+
+# Manage generation of yum repositories for subscribed content:
+manage_repos = 1
+
+# Refresh repo files with server overrides on every yum command
+full_refresh_on_yum = 0
+
+# If set to zero, the client will not report the package profile to
+# the subscription management service.
+report_package_profile = 1
+
+# The directory to search for subscription manager plugins
+pluginDir = /usr/share/rhsm-plugins
+
+# The directory to search for plugin configuration files
+pluginConfDir = /etc/rhsm/pluginconf.d
+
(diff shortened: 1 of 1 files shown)


dustymabe:
I tried to context load all the information in from https://issues.redhat.com/browse/RHEL-1451 and https://github.com/rpm-software-management/librhsm/pull/10/files but things are still a bit fuzzy to me.

I really don't like copying in `rhsm.conf` into our configs like this. Did we consider an option of asking the owning team to break this out into a subpackage instead?

HuijingHei:
> Did we consider an option of asking the owning team to break this out into a subpackage instead?

Sounds good to me, Colin also had the same suggestion.

Create bug https://issues.redhat.com/browse/RHEL-9409 to track the issue, so close this now.

cgwalters:
Thank you for continuing to work on this!

cgwalters:
https://issues.redhat.com/browse/RHEL-9409?focusedId=23154499&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-23154499
And cc @dustymabe - seems like it's either this or that right?

dustymabe:
I think it's fine to include it here (i.e. this PR) if there isn't another good option. What I'm mostly trying to avoid is our file here getting out of date and then we see problems as a result. We could easily detect and prevent that case by having a test that checks our included file against some sort of baseline (could be upstream git or some other thing). 

Another option is to meet with the team as suggested as a possibility to try to explain what problem we are hitting and why this solves the problem.

HuijingHei:
Try to summary here:

To enable host subscriptions to work in containers, see [fix](https://github.com/openshift/os/pull/876) from @cgwalters 

When running entitlement build to install package, it will try to access `rhel.repo` and need to confirm the certification first,

- for ubi image, the SSL CA cert path is set to default `/etc/rhsm/ca` (according to `rhsm.conf` from `subscription-manager`), if detect running in container, then replace cert path to `/etc/rhsm-host/ca` according to https://github.com/candlepin/subscription-manager/blob/subscription-manager-1.29.38-1/src/rhsm/config.py#L363

- for rhcos image, we do not have `rhsm.conf`(because we do not ship `subscription-manager`), and rpm-ostree has the dependency `librhsm`, if detect running in container (check `/etc/rhsm-hot` dir exists), set `conf_file` to `/etc/rhsm-host/rhsm.conf` according to code https://github.com/rpm-software-management/librhsm/blob/master/rhsm/rhsm-context.c#L349, then get SSL CA cert path using the same dir as `rhsm.conf` which is `/etc/rhsm-host/ca`, but finally check `rhsm.conf` path has prefix `/etc/rhsm-host` see https://github.com/rpm-software-management/librhsm/blob/master/rhsm/rhsm-context.c#L542, will update ca cert dir from `/etc/rhsm` to `/etc/rhsm-host`, which means replace `/etc/rhsm-host/ca` -> `/etc/rhsm-host-host/ca`, the path is not correct and fail.

To fix this:

- [~~RHEL-1451~~](https://issues.redhat.com/browse/RHEL-1451) meant to track from librhsm side, see [PR](https://github.com/rpm-software-management/librhsm/pull/10), check if `/etc/rhsm-host/rhsm.conf` does not exist, will leave `/etc/rhsm-host/ca` as it is, but was closed because of risky.

- [~~RHEL-9409~~](https://issues.redhat.com/browse/RHEL-9409) `Break out rhsm.conf into subpackage subscription-manager-rhsm-conf` according to [comment](https://github.com/openshift/os/pull/1372#issuecomment-1733959967), which was closed as can't do.

- According to pointer from @cgwalters , add `subscription-manager` is also an option, the side effect is we might make RHCOS image too big.

- Add `rhsm.conf` file as overlay, but is risky according to [comment](https://github.com/openshift/os/pull/1372#issuecomment-1737987558) from @dustymabe 

- Any other suggestions?

Issue #1047, opened 2022-11-01
transition `oc adm release new` metadata for operating systems


cgwalters:
See https://github.com/coreos/coreos-assembler/pull/3153#discussion_r1010438142

Basically, we should rework `oc adm release new` to natively search for ostree-container images, instead of having this `machine-os` thing.  Further, I think we should suppport multiple in the release image.

Then we can drop the OCP-specific metadata on the container.

marmijo:
We are removing the `io.openshift.build.version-display-names` and `io.openshift.build.versions` labels (specifically where the key is set to `machine-os`) from the RHEL 10-based images.

These labels are currently added by COSA and are consumed by the release controller and `oc adm release info`. They are responsible for:

  - Displaying the `machine-os` version in release information.
  - Driving the RPM diff generation seen on the release controller pages (e.g., the package changes list at the bottom of a release page).

Currently, the release tooling does not support multiple components claiming to be `machine-os`. Including these labels on both the RHEL 9.6 (current default) and RHEL 10.1+ images causes `oc adm release info` to emit warnings regarding multiple versions and results in incorrect or conflicted reporting in the release controller.

To unblock the inclusion of RHCOS 10.1+ images in the 4.21 payload immediately without breaking the reporting for the currently supported release (RHEL 9.6), we are removing these labels from the RHEL 10.1+ builds as a stop-gap measure. We will retain them on RHCOS 9.6 to ensure the current presentation remains accurate. A complete solution to enable the tooling to report on both streams simultaneously will be determined later.

PR #1374, opened 2023-09-27
COS-2473: Re-enable `ostree-container-inject-openshift-cvo-labels` knob

jlebon:
This was previously enabled (#1048) and then disabled again (#1084) because `oc` doesn't know how to handle multiple images with those labels in the release payload. We'll need to solve this eventually if we want to be able to ship multiple OS images in the payload (that's tracked in #1047), but we don't need to block on this if we can remove the legacy `machine-os-content` at the same time.

See also: https://github.com/openshift/driver-toolkit/issues/101
See also: https://github.com/openshift/machine-config-operator/pull/3364

### DIFF
--- a/image-rhel-9.2.yaml
+++ b/image-rhel-9.2.yaml
@@ -21,3 +21,6 @@ vmware-hw-version: 15
 aws-imdsv2-only: false
 aws-volume-type: "gp2"
 aws-x86-boot-mode: "legacy-bios"
+
+# add the requisite OCP metadata to our container image
+ostree-container-inject-openshift-cvo-labels: true


jlebon:
This would go at the same time as https://github.com/openshift/driver-toolkit/pull/135.
We need to make sure DTK, ART, and RHCOS teams are aligned before doing this change.

cgwalters:
I'm ok with trying to just change everything at once, but I think we need to give a heads up to stakeholders.

jlebon:
> I'm ok with trying to just change everything at once, but I think we need to give a heads up to stakeholders.

Sounds good. Anyone other than those listed in the previous comment? I guess there's the MCO team too.

image-c9s.yaml
```diff
@@ -24,3 +24,6 @@ vmware-hw-version: 15
 aws-imdsv2-only: false
 aws-volume-type: "gp2"
 aws-x86-boot-mode: "legacy-bios"
+
+# add the requisite OCP metadata to our container image
+ostree-container-inject-openshift-cvo-labels: true
```
LorbusChris, line 29: see https://github.com/openshift/os/blob/master/image-c9s.yaml#L9
at this point we could probably drop one of the two files.

LorbusChris (changes requested):

cgwalters:
I rebased https://github.com/openshift/machine-config-operator/pull/3364 FWIW, so I think we can get down to just two things that need to change at the same time, not 3.

image-c9s.yaml
jlebon: Ahh whoops, didn't realize it was already there. I just dropped that hunk for now rather than try to include a larger change in this PR.

jlebon:
Just so we don't spend too much time playing the coordination game, let's go ahead and merge this. It seems OK if 4.15 is rocky for a bit until the other PRs merge.

/hold cancel

PR #1375, opened 2023-09-28
c9s.repo: Switch Cloud SIG repo to 4.15

LorbusChris:


c9s.repo
```diff
@@ -55,8 +55,8 @@ enabled=1
 gpgkey=file:///usr/share/distribution-gpg-keys/centos/RPM-GPG-KEY-CentOS-SIG-Virtualization
 
 [sig-cloud-okd]
```
jcapiitao, line 57: Should we rename this to `sig-cloud-okd-4.15` ? in order to be consistent with the manifests (i.e https://github.com/openshift/os/commit/b157aa70682fa5ee94439f01c5caa10ec88daa47)

LorbusChris: I have no strong opinion on that, it might indeed be more obvious that way.

We don't use the `rhel-9.2-server-ose-4.14` for official SCOS builds (instead we pull the RPMs from the `artifacts` payload image), but it's still here for CI builds of SCOS in Prow. Once we can automate builds of `openshift-clients` on CBS (e.g. via Packit.dev), we'll be able to drop the RHEL repo here too.

jcapiitao: Ok, so the `rpm-artifacts-copy` task is kind of "hack" to be able to build SCOS outside Red Hat firewall IIUC for now.
But long term goal is to be able to build `openshift-clients` on CBS and providing it through `sig-cloud-okd` repo, in other words CentOS Stream would be the first target for development.

travier:
/lgtm
(For when this is ready)

c9s.repo
LorbusChris: yes exactly. Maybe we can even build directly on CentOS Stream instead of CBS.

There are a couple of RPMs that are contained in products that are not RHEL (i.e. openvswitch, cri-o, openshift-lcients), that we currently don't build on the Stream Koji instance.
If packit.dev automation comes to Stream before it comes to CBS, which seems likely, I'll try to pursue using Stream for these builds.

PR #1371, opened 2023-09-22
Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Aashish Radhakrishnan (2):
      denylist: snooze ext.config.docker.basic on rawhide
      denylist: drop ext.config.kdump.crash test

Jonathan Lebon (4):
      05core/dracut: enable `iscsi` module
      kola/expected-contents: only check for `iscsi_ibft` on supported arches
      manifests: add crun-wasm.yaml
      Revert "coreos-teardown-initramfs: use NetworkManager to teardown networking"

Michael Armijo (1):
      overrides: fast-track glibc-2.37-10.fc38

Steven Presti (1):
      manifest: add bash-color-prompt for >= f39 fixes coreos/fedora-coreos-tracker#1567

Timothée Ravier (3):
      tests/kola/docker: Add a basic docker run test
      ci & tests: Switch to quay.io/fedora/fedora
      tests/kola/docker/basic: Only run on FCOS

gursewak1997 (5):
      upgrade/extended: Get latest release from update server This test fails when it is run during the time period between the stream build and the release. Updated the test to get the last_release from update server and check if there's an edge pointing to it.
      upgrade/extended: Fix last_release_index extraction method Initially, I was getting the last_release_index from the update json file itself under key org.fedoraproject.coreos.releases.age_index whereas we are using the actual index value of the versions as edge values in edges array
      denylist: extend snooze for coreos.ignition.ssh.key
      denylist: bump snooze for ext.config.kdump.crash on aarch64 This test is still failing.
      denylist: snooze kdump.crash for ppc64le in rawhide
```

PR #1376, opened 2023-09-28
common.yaml: stop using `initramfs-args` in favour of `dracut.conf.d` dropins

jlebon:
The field is deprecated:
https://github.com/coreos/rpm-ostree/pull/3834

See individual commit messages for details.
Closes: #1377


cgwalters (approved):
Awesome series.  I assume so but worth verifying: Did you compare the before/after of the `lsinitrd`?

Also tangentially related to:

> It just stops uselessly generating ifcfg files in the initrd.

Could this be related to https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/NetworkManager-clean-initrd-state.yaml which I started to look at and then recoiled in horror and closed the tab...

jlebon:
I hadn't checked closely the diff, but I did now and one thing that jumps out is that `ifcfg` was still getting pulled in. I opened https://github.com/coreos/fedora-coreos-config/pull/2642 and https://github.com/dracutdevs/dracut/pull/2529 for that.

jlebon:
To be clear, this needs to wait until https://github.com/coreos/fedora-coreos-config/pull/2640 lands in RHCOS.

/hold

jlebon:
> Could this be related to [openshift/machine-config-operator@`master`/templates/common/_base/files/NetworkManager-clean-initrd-state.yaml](https://github.com/openshift/machine-config-operator/blob/master/templates/common/_base/files/NetworkManager-clean-initrd-state.yaml?rgh-link-date=2023-09-28T21%3A23%3A53Z) which I started to look at and then recoiled in horror and closed the tab...

I'm not sure what that script is doing. It seems to be doing something that shouldn't normally be necessary and doesn't talk about how it intersects with the teardown scripts we have and the existing behaviour of not stomping on user configs. Probably deserves a chat with the team that introduced it.

jlebon:
> To be clear, this needs to wait until [coreos/fedora-coreos-config#2640](https://github.com/coreos/fedora-coreos-config/pull/2640) lands in RHCOS.
> 
> /hold

This landed as part of https://github.com/openshift/os/pull/1371.

/hold cancel

PR #1345, opened 2023-08-17
Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Andrew Jeddeloh (1):
      overlay: drop coreos-root in favor of tmpfiles

Colin Walters (2):
      buildroot: Add clippy
      transposefs: Only autosave-xfs for much larger filesystems

Dusty Mabe (10):
      denylist: add denials for tests that layer packages on rawhide
      denylist: drop ext.config.toolbox denial
      denylist: drop package layer denials
      overrides: fast-track kernel-6.4.11-200.fc38
      denylist: mark snoozed tests as warn: true
      tests/kola: delete trailing whitespace in root-bash test
      overlay.d: remove trailing whitespace in ignition-ostree-populate-var.sh
      overlay.d: add 30lvmdevices overlay
      overlay.d/README: fix azure overlay title
      tests/kola: don't run fwupd-refresh-timer test on !fcos

Huijing Hei (3):
      overlay/sysusers: remove `10-groups-basic.conf` which comes from `setup` package
      tmpfiles: Move root bash files copy to tmpfiles.d
      tests: add test to verify `/var/roothome/.bash*` files exist

Michael Armijo (1):
      denylist: snooze `coreos.ignition.ssh.key` on kola-azure

Timothée Ravier (9):
      Enable fwupd refresh timer on Fedora 39
      overlay/README: Update for 17fedora-modularity overlay
      test: Disable 'source' ShellCheck warnings
      test: Remove unneeded whitespace at EOL
      overlay.d/usr/libexec/coreos-check-modularity: Remove whitespace at EOL
      ci: Add action to check for whitespace at EOL
      tests/logrotate-service: Make platform independent
      tests/kola/data/commonlib.sh: ShellCheck fixes
      overlay.d/*: ShellCheck fix for /proc/cmdline reading

YASMIN VALIM (1):
      manifests: name tzdata package in packages list

gursewak1997 (1):
      denylist: bump snooze for ext.config.kdump.crash on aarch64
```

PR #1379, opened 2023-10-05
[release-4.14] OCPBUGS-17154: Bump fedora-coreos-config

jlebon:
For https://github.com/coreos/fedora-coreos-config/pull/2660 for
https://issues.redhat.com/browse/OCPBUGS-17154.

```
Jonathan Lebon (1):
      Revert "coreos-teardown-initramfs: use NetworkManager to teardown networking"
```


jlebon:
I'm not sure why the diff is larger than expected.

Hmm, it looks like the current submodule commit of f-c-c in the release-4.14 branch isn't actually what was in the 4.14 branch right before https://github.com/coreos/fedora-coreos-config/pull/2660 merged.

I think things might've gotten messed up during 4.14 branching somehow. But anyway, I think we do also want the other stuff in here?

dustymabe:
I'm pretty sure things like adding LVM devices was already done i.e. in https://github.com/openshift/os/pull/1345#issuecomment-1699235342 and then verified by QA. 

Things certainly got messed up if we regressed and those "fixes" got pulled out of 4.14.

jlebon:
/hold

Until we figure out what's going on here.

jlebon:
Debugged this with @dustymabe. The f-c-c submodule on this branch was never following the rhcos-4.14 branch of f-c-c. We did the following:
- forced-pushed to the rhcos-4.14 branch of the f-c-c repo the exact commit that the release-4.14 branch was sitting on
- cherry-picked again my backport of https://github.com/coreos/fedora-coreos-config/pull/2660 on top
- verified that no commit which were thought to be backported in the rhcos-4.14 (but really would've never taken place) were all already in the ref

/hold cancel

jlebon:
> * expected [Jira Issue OCPBUGS-17154](https://issues.redhat.com//browse/OCPBUGS-17154) to depend on a bug targeting a version in 4.15.0 and in one of the following states: MODIFIED, ON_QA, VERIFIED, but no dependents were found

Hmm, this is already in 4.15 as part of https://github.com/openshift/os/pull/1371. Can we just override this check or do we need to make a Jira ticket after the fact to satisfy it?

jlebon:
Actually, let's hold this until we determine whether we want this in GA or as an early z update.

/hold

Discussions about that in https://issues.redhat.com/browse/OCPBUGS-17154.

cgwalters:
> /hold

I moved to draft because it's a stricter check actually.

HuijingHei:
Also probably fix https://issues.redhat.com/browse/OCPBUGS-16724, check `c4.2xlarge` nodes failed to start with the same error:
`Error: failed to set networking: Timeout was reached`

jlebon:
@sdodson mentioned he's good with this if we are.
I'm +1 on doing this. Marked as ready for review to get CI going again.

@cgwalters If you agree, can you drop the `hold` label?

/assign @cgwalters

PR #1382, opened 2023-10-12
docs/faq: add question about unpredictable block device names

jlebon:
This has come up multiple times now.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -292,6 +292,80 @@ Currently, non-default multipath configurations for the primary disk cannot be s
 
 If the device is connected to the host via a HBA then it'll show up transparently as a local disk and should work fine. We do not currently support booting from an iSCSI device where the OS is the initiator. iSCSI on secondary disks should be fine.
 
+## Q: How do I configure a secondary block device via Ignition/MC if the name varies on each node?
+
+First, verify that there isn't a `/dev/disk/by-*` symlink which works for your needs. If not, a few approaches exist:
+- If this is a fresh install and you're using the live environment to install RHCOS, as part of the install flow you can inspect the machine (by hand, or scripted) to imperatively figure out what the block device should be according to your own heuristics (e.g. "the only multipath device there is", or "the only NVMe block device"). You can then e.g. "render" the final Ignition config with the device path to pass to `coreos-installer` or directly partition it and optionally format it and use a consistent partition (and optionally filesystem) label that will be available to use in the generic Ignition config.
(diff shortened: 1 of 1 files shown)


jlebon:
I guess it's probably worth linking to https://github.com/openshift/machine-config-operator/issues/1720 in there.

docs/faq.md
```diff
@@ -292,6 +292,79 @@ Currently, non-default multipath configurations for the primary disk cannot be s
...
+        EnvironmentFile=/etc/secondary-dev.env
+        What=$VAR_LIB_FOOBAR_DEV
+        Where=/var/lib/foobar
+        Type=xfs
+
+        [Install]
+        RequiredBy=local-fs.target
+storage:
+  files:
+    - path: /etc/find-secondary-device
```
cgwalters, line 342: This seems better in `/usr/local`?  Though of course that goes right into making user customizations more "first class..."

docs/faq.md
```diff
@@ -292,6 +292,79 @@ Currently, non-default multipath configurations for the primary disk cannot be s
 
 If the device is connected to the host via a HBA then it'll show up transparently as a local disk and should work fine. We do not currently support booting from an iSCSI device where the OS is the initiator. iSCSI on secondary disks should be fine.
 
+## Q: How do I configure a secondary block device via Ignition/MC if the name varies on each node?
+
+First, verify that there isn't a `/dev/disk/by-*` symlink which works for your needs. If not, a few approaches exist:
+- If this is a fresh install and you're using the live environment to install RHCOS, as part of the install flow you can inspect the machine (by hand, or scripted) to imperatively figure out what the block device should be according to your own heuristics (e.g. "the only multipath device there is", or "the only NVMe block device"). You can then e.g. "render" the final Ignition config with the device path to pass to `coreos-installer` or directly partition it and optionally format it and use a consistent partition (and optionally filesystem) label that will be available to use in the generic Ignition config.
```
cgwalters, line 298: I'm pretty sure Assisted Installer ships some logic to do this.

docs/faq.md
cgwalters: Or we could instead document doing this from a privileged container which seems better.

cgwalters (approved):
Offhand looks like sane raw material that we can hopefully add to the official docs at some point.

docs/faq.md
jlebon: I probably should've added a comment there, but the issue is that `/var` isn't yet mounted when this runs. Busting out a the container stack is probably out of the question too.

Of course, we could rejig this so that it runs not as part of `local-fs.target` but afterwards. The problem with that is that one would have to be aware of this when ordering other systemd services that do stuff on top of the mount point.

docs/faq.md
jlebon: Do you have a link offhand? Definitely makes sense to link to AI docs if that's what the reader is actually using.

docs/faq.md
jlebon: Added a comment to clarify this now.

travier (approved):

Issue #1384, opened 2023-10-17
[rhel-9.4 variant] some kola tests fail with Found legacy sysroot.readonly flag


marmijo:
When building a rhel-9.4 variant, all kola reprovision tests  fail on aarch64, ppc64le, and s390x with the same error message. EDIT: some other kola tests are failing with this same error message on x86_64. Those are listed below in a separate comment.
```
ostree-prepare-root[2206]: Found legacy sysroot.readonly flag, not configured in ostree/prepare-root.conf
```
It looks like ostree-prepare-root is potentially the issue here.
The test runs just fine on 4.15-9.2 multi-arch builds and output:
```
[    7.910212] ostree-prepare-root[1116]: preparing sysroot at /sysroot
[    7.910503] ostree-prepare-root[1116]: Resolved OSTree target to: /sysroot/ostree/deploy/rhcos/deploy/96d020aae31deb0f28ccb464c9437f80673a8554307286af991ddc66e498b0d5.0
[    7.910925] ostree-prepare-root[1116]: filesystem at /sysroot currently writable: 1
[    7.911107] ostree-prepare-root[1116]: sysroot.readonly configuration value: 1
[   11.585987] systemd[1]: ostree-prepare-root.service: Deactivated successfully.
``` 

But the 4.15-9.4 multi-arch reprovision tests look like this:
```
[   74.622657] ostree-prepare-root[1876]: Resolved OSTree target to: /sysroot/ostree/deploy/rhcos/deploy/c9441e1047b7e82ae08cd2328eeeb1c486f9e5104fde9fc5a53558850d2fca1c.0
[   74.623804] ostree-prepare-root[1876]: Found legacy sysroot.readonly flag, not configured in ostree/prepare-root.conf
[   74.623885] ostree-prepare-root[1876]: sysroot.readonly configuration value: 1 (fs writable: 1)
[   74.623989] ostree-prepare-root[1876]: composefs: Mounting with no digest or signature check
[   74.625318] ostree-prepare-root[1876]: composefs: optional support failed: No such file or directory
[   77.326671] systemd[1]: ostree-prepare-root.service: Deactivated successfully.
```
This is currently affecting:

- coreos.boot-mirror.luks
- coreos.boot-mirror
- test.tap
- luks.tang
- luks.sss.t2
- luks.sss.t1
- ext.config.shared.root-reprovision.swap-before-root
- ext.config.shared.root-reprovision.raid1
[...]

PR #1385, opened 2023-10-17
manifest-rhel-9.4: fake out RHEL_VERSION

jlebon:
During this transitionary period, we're inheriting from the C9S release package which doesn't define `RHEL_VERSION`. Since we're trying here to pretend to be a normal RHEL-based RHCOS, let's have a valid `RHEL_VERSION` string too.

Fixing this here means that consumers of this file don't need to worry about this weird case where RHCOS is pretending to be RHEL but is really CentOS Stream.

Fixes: #1384


cgwalters (approved):
/lgtm

marmijo (approved):
LGTM

PR #1386, opened 2023-10-17
manifest-rhel9.4: create os-release.rhel

marmijo:
Several RHCOS kola tests are failing because the file `os-release.rhel` cant be found. 
Change the extension of the os-release file in order to better configure this variant 
to look like a RHEL based RHCOS instead of a C9S based one.

### DIFF
--- a/manifest-rhel-9.4.yaml
+++ b/manifest-rhel-9.4.yaml
@@ -37,7 +37,7 @@ postprocess:
      set -xeo pipefail
 
      # Tweak /usr/lib/os-release
-     grep -v "OSTREE_VERSION" /etc/os-release > /usr/lib/os-release.stream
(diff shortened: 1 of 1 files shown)


jlebon (approved):

jlebon:
@marmijo If you still have it, it might be worth commenting here the error message from the test failures you got so it shows up in searches in the future.

marmijo:
The following tests were the failing ones:

- ext.config.version.rhel-major-version
- ext.config.version.rhel-matches-rhcos-build

```
+ variant=rhcos
+ case "${variant}" in
++ source /usr/lib/os-release.rhel
/usr/local/bin/kola-runext-rhel-major-version: line 17: /usr/lib/os-release.rhel: No such file or directory
```

PR #1388, opened 2023-10-18
denylist: add kola tests that are failing on rhel-9.4 variant

marmijo:
The following kola tests are failing on the new RHEL-9.4 variant:

- ext.config.rpm-ostree.replace-rt-kernel
- ext.config.shared.content-origins
- ext.config.version.rhel-matches-rhcos-build

Let's denylist them for now so we can get boot images built in the devel pipeline.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -10,8 +10,26 @@
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
   - c9s
+  - rhel-9.4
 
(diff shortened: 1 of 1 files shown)


marmijo:
I force-pushed an update to denylist `iso-live-login.uefi-secure` and `iso-as-disk.uefi-secure` which are also failing on the new variant because it is based on c9s content.

marmijo:
Oops, I force pushed to the wrong branch.

PR #1389, opened 2023-10-20
OCPBUGS-22929: [release-4.11]: drop compat-openssl10

marmijo:
Drop compat-openssl10 from RHCOS builds based on RHEL8.

See https://issues.redhat.com/browse/COS-2461

PR #1390, opened 2023-10-20
[release-4.12] drop compat-openssl10

marmijo:
Drop compat-openssl10 from RHCOS builds based on RHEL8.

See https://issues.redhat.com/browse/COS-2461

### DIFF
--- a/manifest-rhel-8.6.yaml
+++ b/manifest-rhel-8.6.yaml
@@ -147,8 +147,6 @@ packages:
  # We include the generic release package and tweak the os-release info in a
  # post-proces script
  - redhat-release
- # RHEL7 compatibility
- - compat-openssl10
  # SCOS package name does not include a version number
  - openvswitch2.17
  # https://github.com/openshift/os/issues/1036


cgwalters:
What I'd like to be able to do is run the [payload tests](https://docs.ci.openshift.org/docs/release-oversight/payload-testing/) on this PR but I don't think it'll work because we're not generating a release image etc. in this

travier:
To be able to land this, we'll need to do the following paperwork:
- Create a bug for 4.13 about this, immediately closed as CURRENTRELEASE as it's fixed in 4.13 already.
- Clone the bug for 4.12, set it to depend on the 4.13 bug and use it for this PR
- Clone the 4.12 bug for 4.11, set it to depend on the 4.12 bug and use it for https://github.com/openshift/os/pull/1389

HuijingHei (approved):
LGTM

travier:
Updated in https://github.com/openshift/os/pull/1418 as I can not push here.

PR #1397, opened 2023-11-17
NO-JIRA: denylist: skip PXE tests on ppc64le

jlebon:
These tests are broken with the latest GRUB BOOTP binaries in cosa, which is now on f39.


jlebon:
CI here doesn't test this. The real test will be in the prod pipeline. Let's merge it to find out faster.

PR #1406, opened 2023-12-14
catch coreos.ignition.failure err on 4.16 pipeline

c4rt0:
After failing to reproduce `coreos.ignition.failure` locally I am removing it from denylist in order to catch errors in the pipeline. The goal of this PR is to allow it in RHCOS 4.16, catch the errors from the log and re-introduce `coreos.ignition.failure` to kola-denylist while working on solution.

See: https://github.com/coreos/coreos-assembler/issues/3670

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -4,49 +4,37 @@
 - pattern: ext.config.rpm-ostree.replace-rt-kernel
   tracker: https://github.com/openshift/os/issues/1315
   osversion:
-  - c9s
+    - c9s
 
 - pattern: iso-live-login.uefi-secure
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
-  - c9s
-  - rhel-9.4
+    - c9s
+    - rhel-9.4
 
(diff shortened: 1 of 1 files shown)


jschintag:
Why not point the pipeline to your own branch of this repo instead, using the dev pipeline?

c4rt0:
> Why not point the pipeline to your own branch of this repo instead, using the dev pipeline?

Oh, that's a good suggestion. Let me investigate and find out how to do it.

jschintag:
>Oh, that's a good suggestion. Let me investigate and find out how to do it.

I think you would either need your own repository for pipeline config or point the dev one at your fork of this repo.
So Pipeline -> custom config -> custom openshift/os -> profit

HuijingHei:
> > Oh, that's a good suggestion. Let me investigate and find out how to do it.
> 
> I think you would either need your own repository for pipeline config or point the dev one at your fork of this repo. So Pipeline -> custom config -> custom openshift/os -> profit

I think it would be better to add this to doc.

PR #1408, opened 2023-12-15

coreosbot-releng:

```
Dusty Mabe (1):
      overlay.d: add 40grub overlay

Michael Armijo (1):
      denylist: remove ext.config.kdump.crash

jbtrystram (1):
      The test for iscsi boot fails on somes arches
```


marmijo:
CI is failing on this PR because `iso-offline-install-iscsi.bios` is timing out,  but this PR contains a commit to denylist that test, so we can probably ignore the failure.

/retitle NO-JIRA: Bump fedora-coreos-config

marmijo:
Actually, I think CI is failing because it's not denylisted in RHCOS yet. I opened https://github.com/openshift/os/pull/1410 to do that.

PR #1410, opened 2023-12-21
NO-JIRA: denylist: add iso-offline-install-iscsi.bios

marmijo:
This test is timing out in RHCOS and FCOS. Let's denylist it to unblock the pipeline for now while we investigate https://github.com/coreos/fedora-coreos-tracker/issues/1638


aaradhak (approved):

aaradhak (approved):

mike-nguyen:
/retitle NO-JIRA: denylist: add iso-offline-install-iscsi.bios

PR #1413, opened 2024-01-05
COS-1935: overlay: include 40grub overlay from FCOS configs

dustymabe:
See https://github.com/coreos/fedora-coreos-config/pull/2769

This won't really have any effect right now but will enable someone to use the generated ociarchive container images in the OSBuild workflow that calls bootupctl with the --with-static-configs argument.


jmarrero (approved):
/lgtm

PR #1415, opened 2024-01-08
OCPBUGS-26477: Enable ima signatures

lack:
As requested in OCPBUGS-26477, adding this flag to common.yaml will cause any IMA signatures to be installed from the RPMs to the RHCOS image.

They can then be used to locally cryptographically attest to the authenticity of the Red Hat-provided file contents at runtime.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -38,6 +38,8 @@ arch-include:
 
 documentation: false
 
+ima: true
+
 postprocess:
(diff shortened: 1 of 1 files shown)


travier:
/hold
This an RFE that needs discussion before moving to implementation. We would need to add tests for it as well.

PR #1417, opened 2024-01-14

coreosbot-releng:

```
Jonathan Lebon (2):
      40ignition-ostree: avoid jq try-catch
      tests/bootupd: skip test on ppc64le and s390x if bootupd not installed

Michael Armijo (1):
      denylist: update outdated entries

Renata Ravanelli (1):
      Add bootupd package for s390x - Bootupd is now available for s390x as part of our ongoing efforts to integrate osbuild. Let's include it

gursewak1997 (1):
      denylist: snooze root-reprovision.linear
```


jlebon:
This has the fix for the bootupd test on s390x and ppc64le.
/approve
/lgtm

PR #1418, opened 2024-01-15
OCPBUGS-22928: [release-4.12] manifests: drop compat-openssl10

travier:
Drop compat-openssl10 from RHCOS builds based on RHEL8 and make sure we don't include it via an exclude.

See: https://issues.redhat.com/browse/COS-2461


jlebon:
/hold

Still needs to be tested by other teams before we get this in.

marmijo (approved):

sdodson:
This change is unique to 4.12 as it's the last supported 8.x based release and thus not really a backport.

PR #1419, opened 2024-01-15
OCPBUGS-22929: [release-4.11]: manifests: drop compat-openssl10

travier:
Drop compat-openssl10 from RHCOS builds based on RHEL8 and make sure we don't include it via an exclude.

See: https://issues.redhat.com/browse/COS-2461


jlebon:
/hold

Still needs to be tested by other teams before we get this in.

travier:
/close
Given that 4.11 is soon EOL: https://access.redhat.com/support/policy/updates/openshift

PR #1421, opened 2024-01-17
rhcos-packages: Make bootupd no arch dependent

ravanelli:
We now have bootupd support across all architectures.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -242,6 +242,8 @@ packages:
  - kernel-modules-extra
  # Audit
  - audit
+ # Bootloader updater
+ - bootupd
  # Containers
  - containernetworking-plugins
  # Pinned due to cosa on Fedora not honoring RHEL 8 modules as expected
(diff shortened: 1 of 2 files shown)


common.yaml
```diff
@@ -242,6 +242,8 @@ packages:
  - kernel-modules-extra
  # Audit
  - audit
+ # Bootloader updater
+ - bootupd
```
jlebon, line 246: We already include the same `bootupd.yaml` manifest in this file. We just need to move it out from the `arch-include` and into the `include` section.

BTW, to confirm the no-op s390x support will also correctly no-op on the `bootupctl backend generate-update-metadata` step?

jlebon:
/hold

We can't do this yet because bootupd is too ancient on 9.2 to ship on ppc64le and s390x, which is still the default variant built and shipped. So for now this needs to go in 9.4-specific manifests.

dustymabe:
newer version of this in https://github.com/openshift/os/pull/1439

I think 4.16 is 9.4 now so what was holding up this PR shouldn't be an issue any longer?

travier:
/close
Closing in favor of https://github.com/openshift/os/pull/1439

PR #1424, opened 2024-01-19
NO-JIRA: manifests: default to rhel-9.4 variant and test it in Prow

jlebon:
This mirrors the ART switchover to defaulting to the 9.4 builds in CI
and nightlies.

Note the pipecfgs already explicitly specify variants for both the
4.16-9.2 and 4.16-9.4 variants. So this change is mostly symbolic (pun
intended), but does affect the default variant built by developers.

While we're here, do some other subtle cleanups. I think it was
confusing how e.g. `extensions-rhel-9.4.yaml` symlinked back to
`extensions-rhel-coreos-9.yaml`, like it was going back up one
abstraction layer. Instead, I think it's clearer to have the concrete
implementations symlinks among each other if they happen to be
identical. We do this now with `extensions-rhel-9.2.yaml` being a
symlink to its 9.4 counterpart. Similarly for `image.yaml` versions.

git makes it seem like it's a larger change than it is because of the
type changes. But it's basically just some renames and symlinking. No
content was changed.


travier:
You need to update all the symlinks for images, extensions, etc. There is a small dedicated cosa command to do it `update-variant`.

jlebon:
> You need to update all the symlinks for images, extensions, etc.

Thanks, good catch!

> There is a small dedicated cosa command to do it `update-variant`.

That didn't quite work actually because of the way some of the symlinks were set up. I've fixed this now so that the command works and the symlinks are clearer.

jlebon:
Ahh yes, we need to also update Prow configuration.

travier (approved):
/lgtm
/hold

jlebon:
OK updated this but it now requires https://github.com/openshift/release/pull/48522.

jlebon:
So to make this explicit, this PR is now more important than how it started: it makes CI actually test the rhcos-9.4 variant rather 9.2.

jlebon:
Now that https://github.com/openshift/release/pull/48522 is merged, let's try it again!

/retest

ci/prow-entrypoint.sh
```diff
@@ -261,13 +261,13 @@ main() {
             cosa_build
             kola_test_qemu
             ;;
-        "rhcos-92-build-test-qemu")
+        "rhcos-9-build-test-qemu")
```
c4rt0, line 264: Should it not be the:
`rhcos-94-build-test-qemu`
?

ci/prow-entrypoint.sh
```diff
@@ -261,13 +261,13 @@ main() {
...
             setup_user
             cosa_init "rhel-coreos-9"
             cosa_build
             kola_test_qemu
             ;;
-        "rhcos-92-build-test-metal")
+        "rhcos-9-build-test-metal")
```
c4rt0, line 270: Should it not be the:
`rhcos-94-build-test-qemu`
?

mike-nguyen:
I forgot there's no 9.4 repo mirror because there's no 9.4 repo yet. We're using a mix of centos stream and rhel-9.2 content for 9.4.

jlebon:
> I forgot there's no 9.4 repo mirror because there's no 9.4 repo yet. We're using a mix of centos stream and rhel-9.2 content for 9.4.

Right yeah, not sure why I thought that'd work. Will adapt.

jlebon:
OK cool, CI is green on this now! :heavy_check_mark:

ci/prow-entrypoint.sh
jlebon: I changed it in https://github.com/openshift/release/pull/48522 to just call `rhcos-9-build-test-qemu`.

jlebon: Thanks for reviewing!

c4rt0: Ah, not that I knew it but now it makes sense :)

c4rt0 (approved):
lgtm

mike-nguyen:
/lgtm

thanks @jlebon for fixing our CI!

dustymabe (approved):
/lgtm

travier:
I'm wondering how this passed CI as we're having failures in https://github.com/openshift/os/pull/1443 that are related to the fact that our 9.4 manifests are using the `rhcos` `VARIANT_ID` and not the  `scos` one and thus try to run Secure Boot tests: https://github.com/openshift/os/blob/master/ci/prow-entrypoint.sh#L124

travier:
Ah, we had test snooze that expired.

PR #1426, opened 2024-01-22
NO-JIRA: denylist: re-deny `iso-offline-install-iscsi.bios`

jlebon:
It's known failing in the pipelines, so there's no use snoozing it. Let's keep it off until we've debugged and fixed it.


jlebon:
Let's just get this in. Prow doesn't run testiso tests and the pipeline is broken anyway.

/override ci/prow/images

Issue #1428, opened 2024-01-23
[4.16] Clean up legacy RHEL 8 workarounds


travier:
Clean up legacy workaround that should now be fixed in RHEL 9.2+

- Remove https://github.com/openshift/os/blob/master/overlay.d/05rhcos/usr/libexec/rhcos-rebuild-selinux-policy & associated unit
	- Probably close https://github.com/openshift/os/issues/1381
- https://github.com/openshift/os/blob/master/overlay.d/15rhcos-rhel8-workarounds/usr/lib/systemd/system/rpm-ostreed.service.d/startlimit.conf
- Update kola tests to remove all RHEL 8 mentions: https://github.com/openshift/os/tree/master/tests/kola

PR #1429, opened 2024-01-23
COS-2612: overlay: RHEL 8 workarounds cleanups

travier:
overlay: Remove SELinux policy rebuild workaround

This workaround is not needed anymore with newer ostree and
policy-coreutils releases.

See: https://github.com/openshift/os/issues/1428

---

overaly: Remove RHEL 8 workaround for rpm-ostree

We are now shipping an updated rpm-ostree release.
This is also not included in any manifest right now so effectively dead
code.

See: https://github.com/openshift/os/issues/1428

---

Fixes: https://github.com/openshift/os/issues/1428


jlebon (approved):
/lgtm
/approve

PR #1438, opened 2024-02-07
NO-JIRA: common.yaml: include passwd RPM in manifest

dustymabe:
This package will drop out of the FCOS manifests in Fedora 40+ because the shadow-utils RPM now provides it [1]. Let's name it here so we continue to include it in RHCOS until it is also removed from RHEL (assuming that will happen).

[1] https://src.fedoraproject.org/rpms/shadow-utils/c/91360f25a8c8b810d59bec2803a2477a2647c775?branch=rawhide

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -261,6 +261,8 @@ packages:
  - authselect
  # https://bugzilla.redhat.com/show_bug.cgi?id=1900759
  - qemu-guest-agent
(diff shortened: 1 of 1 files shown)


jlebon:
CI is being worked on in https://github.com/openshift/os/pull/1424.

/approve
/lgtm

dustymabe:
CI is failing on a depsolve issue. It's being tracked internally.

PR #1439, opened 2024-02-08
NO-JIRA: make bootupd not arch specific

dustymabe:
It's available for all architectures now.


travier:
/hold
We are indeed building 9.4 by default now but this will break our 9.2 (backup builds). It's unlikely that we'll need those now but just in case, let's hold a short while until we remove all 9.2 traces from here.

dustymabe:
can we just tag in newer bootupd into 4.16 9.2 tags like we did for 4.15?

This is currently blocking OSBuild because we can't build images for ppc64le.

travier:
/unhold
Let's not block osbuild work for that

PR #1443, opened 2024-02-18

coreosbot-releng:

```
jbtrystram (1):
      05core: skip sysroot.mount for remote kdump
```


travier:
This is failing on `--qemu-firmware=uefi-secure basic` so Secure Boot which we don't have in C9S (9.4).

travier:
https://github.com/openshift/os/pull/1448

PR #1444, opened 2024-02-19
OCPCLOUD-2484,OCPCLOUD-2481: Adds acr and gcr image credential providers

theobarberbany:
This change adds the azure acr-image-credential-provider and gcp gcr-image-credential-provider packages that kubelet will use to retrieve authentication credentials


theobarberbany:
/hold

Until we have CI RPMs in the artifacts image, so we don't break OKD builds

LorbusChris:
OKD composes are currently halted. While I'd still prefer for the RPMs to land in `artifacts` first, OKD does not block this. Feel free to unhold.
/lgtm

PR #1449, opened 2024-02-26
OCPBUGS-29437: install the smaller openshift-kubelet package

sdodson:
The packaging has recently been refactored to offer a kubelet only package, which is all that RHCOS should require. This removes extraneous server binaries kube-apiserver, kube-controller-manager, and kube-scheduler


jlebon:
/approve

CI should be fixed by https://github.com/openshift/os/pull/1448.
/retest

travier (approved):
/lgtm

PR #1451, opened 2024-02-27
NO-JIRA: ci/prow-entrypoint: put kola logs in ARTIFACT_DIR

jlebon:
Right now, we're not getting any logs from the kola tests which makes it extremely hard to debug test failures.

The `cosa kola` wrapper has knowledge of `ARTIFACT_DIR` too, but since we're running kola multiple times, let's just set up the right path with distinct subdirs per run right here.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -124,11 +124,11 @@ kola_test_qemu() {
     # Skip Secure Boot tests on SCOS for now
     # See: https://github.com/openshift/os/issues/1237
     if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
-        cosa kola --basic-qemu-scenarios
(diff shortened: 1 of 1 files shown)


jlebon:
OK cool, looks like this is working!
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_os/1451/pull-ci-openshift-os-master-scos-9-build-test-metal/1762522166621900800/artifacts/test/artifacts/
- - https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_os/1451/pull-ci-openshift-os-master-scos-9-build-test-qemu/1762522166655455232/artifacts/test/artifacts/

HuijingHei (approved):
LGTM

PR #1452, opened 2024-02-28
NO-JIRA: denylist: re-snooze `basic` on rhel-9.4

jlebon:
Commit 81e5134 relies on logic in the Prow scripts to skip the Secure
Boot basic test, but that logic does not exist in the prod pipelines. So
we still need to snooze the basic test for now.

Once https://github.com/coreos/coreos-assembler/pull/3652 is in, we
should be able to clean this up.

Fixes 81e5134 ("kola-denylist: Do not snooze basic tests anymore").


marmijo (approved):

travier:
Arg shoot. Thanks for the fix.

PR #1453, opened 2024-02-29
[release-4.15] OCPBUGS-29995: Bump-fedora-coreos-config

coreosbot-releng:

```
jbtrystram (1):
      05core: skip sysroot.mount for remote kdump
```


jlebon:
@jbtrystram One detail we forgot: for PRs to !master, we should have the workflow prepend the branch name in brackets like here.

PR #1455, opened 2024-03-04
OCPBUGS-30149: Revert "common.yaml: enable cliwrap"

jlebon:
This reverts commit ff686c634bf298ea69785e27cae4de11bb87be6e.

With cliwrap, rpm-ostree runs rpm with less privileges. We special-case `--query`, so e.g. `rpm -qf` still runs with dropped privileges.

But anyway, we actually want to reduce this to wrapping *just* `dnf` for now (see discussions in
https://github.com/coreos/fedora-coreos-config/pull/2887).

Resolves: https://issues.redhat.com/browse/OCPBUGS-30149

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -33,10 +33,6 @@ conditional-include:
     # And remove some cruft from grub2
     include: fedora-coreos-config/manifests/grub2-removals.yaml
 
-# we want derivations to be able to type `dnf install ...` for a unified
-# experience
-cliwrap: true
-
 documentation: false
 
 postprocess:


jmarrero (approved):
/lgtm
/approve

jlebon:
While CI is running on this, let me sanity-check that's what we're hitting in https://issues.redhat.com/browse/OCPBUGS-30149.

jlebon:
It's interesting though; it seems the `EPERM` only shows up when querying a file that doesn't exist. Otherwise, it works as usual. E.g.:

```
[root@cosa-devsh ~]# rpm -qf /etc/audit/rules.d/audit.rules
audit-3.1.2-2.el9.x86_64
[root@cosa-devsh ~]# rpm -qf /etc/audit/rules.d/enoent
error: file /etc/audit/rules.d/enoent: Permission denied
```

So even with this, it's possible that the MCO might still get an error (but this time about the file not existing) which wouldn't be related to this change.

mike-nguyen:
On the previous working build it gives us the correct error when querying for a file that doesn't exist.

```
[root@cosa-devsh ~]# rpm -qf /etc/audit/rules.d/enoent
error: file /etc/audit/rules.d/enoent: No such file or directory
```

sinnykumari:
> So even with this, it's possible that the MCO might still get an error (but this time about the file not existing) which wouldn't be related to this change.

Interesting, from the bug `/etc/audit/rules.d/mco-audit-quiet-containers.rules` do exist on the node and rpm -qf  gives permission denied error. We will get to know better after MCO code runs with the fix.

PR #1458, opened 2024-03-11
COS-2258: prow-entrypoint: kola arg drop `--basic-qemu-scenarios`

prestist:
With the pr https://github.com/coreos/coreos-assembler/pull/3652 the argument `basic-qemu-scenarios` are migrated to formal kola tests Additionally update `development` documentation to exercise basic tests using the glob pattern.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -123,12 +123,22 @@ kola_test_qemu() {
 
     # Skip Secure Boot tests on SCOS for now
     # See: https://github.com/openshift/os/issues/1237
-    if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
-        cosa kola --basic-qemu-scenarios --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+    # Due to the changes in https://github.com/coreos/coreos-assembler/pull/3652
+    # we need to check if the basic.nvme is available in the list of tests
+    local args=""
+    local manifest="src/config/manifest.yaml"
+    if cosa kola list --json | jq -r '.[].Name' | grep -q "basic.nvme"; then
+        if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
+            args+="--denylist-test *.uefi-secure"
+        fi
     else
-        cosa kola --basic-qemu-scenarios --skip-secure-boot --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+        if ! rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
+            cosa kola --basic-qemu-scenarios --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+        else
+            cosa kola --basic-qemu-scenarios --skip-secure-boot --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+        fi
     fi
-    cosa kola run --parallel 2 --output-dir ${ARTIFACT_DIR:-/tmp}/kola
+    cosa kola run ${args} --parallel 2 --output-dir ${ARTIFACT_DIR:-/tmp}/kola
 }
 
 # Build metal, metal4k & live images and run kola tests
--- a/docs/development-rhcos.md
+++ b/docs/development-rhcos.md
@@ -103,7 +103,7 @@ and later. For older versions, see the internal documentation.
 - You may then run tests on the image built with [`kola`][kola]:
   ```
   # Run basic QEMU scenarios
-  $ cosa kola --basic-qemu-scenarios
+  $ cosa kola run basic*
   # Run all kola tests (internal & external)
   $ cosa kola run --parallel 2
   ```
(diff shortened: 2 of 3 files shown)


ci/prow-entrypoint.sh
```diff
@@ -124,9 +124,9 @@ kola_test_qemu() {
     # Skip Secure Boot tests on SCOS for now
     # See: https://github.com/openshift/os/issues/1237
     if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
-        cosa kola --basic-qemu-scenarios --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+        cosa kola --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
     else
-        cosa kola --basic-qemu-scenarios --skip-secure-boot --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+        cosa kola --skip-secure-boot --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
     fi
```
jlebon, line 130: This would still do a separate kola run. We need to instead translate it to just changing the arguments that would be passed to `cosa kola run` below (line 131) for whether or not to denylist the Secure Boot tests.

prestist: Sorry I don't quite understand :/ Can we chat tmr ?

jlebon: So the current behaviour (without your patch) is to do two runs: first the basic scenarios, and then the full kola run. This is similar to the previous behaviour in coreos-ci-lib.

With the cosa change, we don't want to do that first run anymore since the basic tests are part of the full kola run. So that if-statement is just about whether or not we will need to denylist the Secure Boot tests in the full run.

But actually, for this PR to pass CI (and CI in https://github.com/coreos/coreos-assembler/pull/3652), we could do a similar thing to what we did in coreos-ci-lib, i.e. keep the current logic of two runs if old cosa, and otherwise a single run.

prestist: Ah yes, that makes sense, for one reason or another the first comment was bouncing off my noggin.

prestist: What do we do about the notes? *.md files that are edited, should I add a comment about it being 1 thing in one version and another in the new version?

ci/prow-entrypoint.sh
```diff
@@ -123,10 +123,18 @@ kola_test_qemu() {
 
     # Skip Secure Boot tests on SCOS for now
     # See: https://github.com/openshift/os/issues/1237
-    if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
-        cosa kola --basic-qemu-scenarios --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+    # Due to the changes in https://github.com/coreos/coreos-assembler/pull/3652
+    # we need to check if the basic.nvme is available in the list of tests
+    if cosa kola list --json | jq -r '.[].Name' | grep -q "basic.nvme"; then
+        if ! rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
+            cosa kola --denylist-test *.uefi-secure --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
```
jlebon, line 130: We don't want to run the full suite twice. We should just set a variable here that has the `--denylist` flag which we pass to `cosa kola run` below.

ci/prow-entrypoint.sh
```diff
@@ -123,12 +123,21 @@ kola_test_qemu() {
 
     # Skip Secure Boot tests on SCOS for now
     # See: https://github.com/openshift/os/issues/1237
-    if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
-        cosa kola --basic-qemu-scenarios --output-dir ${ARTIFACT_DIR:-/tmp}/kola-basic
+    # Due to the changes in https://github.com/coreos/coreos-assembler/pull/3652
+    # we need to check if the basic.nvme is available in the list of tests
+    local args=""
+    if cosa kola list --json | jq -r '.[].Name' | grep -q "basic.nvme"; then
+        if ! rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-release"; then
```
jlebon, line 130: While reviewing this, I noticed https://github.com/openshift/os/pull/1448/files#r1526385732. I guess we might as well fix it here. So we can drop the `!` and define `manifest=src/config/manifest.yaml` just above this.

And similarly in the else block below, either add a `!` or swap the if-statement body with the else body.

jlebon (commented):
Found an issue which isn't new but would be nice to fix. Otherwise LGTM!

ci/prow-entrypoint.sh
travier: I think we should drop this workaround now that we have proper filtering and use the kola denylist instead.

travier:
See also: https://github.com/openshift/os/pull/1464

LorbusChris:
The special casing for SCOS might not be required going forward (assuming Secure Boot on SCOS has actually been fixed), see https://github.com/openshift/os/pull/1464

ci/prow-entrypoint.sh
prestist: @jlebon  ok, done :)

@travier could you elaborate? Do you mean we should update the kola tests to not run on centos?

jlebon:
Nice work!

/lgtm

jlebon:
This change is quite safe and is internal to us.

PR #1464, opened 2024-03-18
prow-entrypoint: Resume Secure Boot testing on SCOS

LorbusChris:
The issue with the shim RPM [1] was resolved by [2].

[1] https://issues.redhat.com/browse/RHEL-4391
[2] https://gitlab.com/redhat/centos-stream/rpms/shim-unsigned-x64/-/commit/d4af6470909fca24b647d2a30498ede37c852826

/hold
for testing

/cc @travier @jlebon

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -121,24 +121,13 @@ cosa_build() {
 kola_test_qemu() {
     cosa buildextend-qemu
 
-    # Skip Secure Boot tests on SCOS for now
-    # See: https://github.com/openshift/os/issues/1237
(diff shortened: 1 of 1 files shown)


jlebon:
It's still failing in CI at least.

dustymabe:
if the new RPM still isn't working it may be affected by the same problem we hit in FCOS: https://github.com/coreos/fedora-coreos-tracker/issues/1694

To fix that the shim RPM needed to be rebuilt.

LorbusChris:
It looks like there hasn't been a new build on the Stream Koji: https://kojihub.stream.centos.org/koji/packageinfo?packageID=3318

PR #1467, opened 2024-03-20
RHEL-29856: Temporarily use mirrored repos for c9s, and pin NetworkManager in rhel-9.4

jlebon:
c9s.repo: temporarily use mirrored repos

Unlike the pungi compose output repos, the mirrors have older versions
of RPMs, which we want right now while we're in this weird period where
we'd like to be able to pin to some older content because the newer
stuff will never make it to rhel-9.4 anyway.

---

manifest-rhel-9.4: pin to NetworkManager-1.46.0-1.el9

The latest NM currently in c9s is hitting issues. But anyway, that NM is
not going to end up in 9.4. So let's just temporarily pin to the latest
version that is.

We should be able to clean this up once the rhel-9.4 variant stops
consuming from c9s.

See also: https://issues.redhat.com/browse/RHEL-29856

### DIFF
--- a/c9s.repo
+++ b/c9s.repo
@@ -1,6 +1,8 @@
 [baseos]
 name=CentOS Stream 9 - BaseOS
-baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os
+baseurl=http://mirror.stream.centos.org/9-stream/BaseOS/$basearch/os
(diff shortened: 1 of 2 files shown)


mike-nguyen:
/retitle RHEL-29856: Temporarily use mirrored repos for c9s, and pin NetworkManager in rhel-9.4

PR #1471, opened 2024-03-26
NO-JIRA: faq: add details about RAID installations

jlebon:
We have OpenShift docs for some of this, but it's nice to have it summarized succinctly in this format.

Notably new is the Hybrid RAID path, which still needs to be documented in the OCP docs.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -440,3 +440,52 @@ rpm-ostree uninstall kernel-rt-core kernel-rt-kvm kernel-rt-modules kernel-rt-mo
   --install kernel-rt-modules-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm \
   --install kernel-rt-modules-extra-4.18.0-305.34.2.rt7.107.el8_4.x86_64.rpm
 ```
+
+## Q: How do I install RHCOS on top of a RAID device?
+
+### Hardware RAID
+
+This is transparent to RHCOS and shows up as a unified block device. You should be able to target `coreos-installer install` at that device as usual.
+
+### Software RAID
+
+RHCOS supports software RAID1 via high-level sugar: https://docs.openshift.com/container-platform/4.15/installing/install_config/installing-customizing.html#installation-special-config-mirrored-disk_installing-customizing
+
+### Fake RAID/Hybrid RAID/Intel VROC
+
+Some systems support what is known as Fake or Hybrid RAID, where some of the work of maintaining the RAID is offloaded to the hardware, but otherwise it appears just like software RAID to the OS.
+
+To install to these devices, configure them as necessary in the firmware and/or using `mdadm` as documented.
+
+#### Intel VROC
+
+To configure an Intel VROC-enabled RAID1, first create the IMSM container, e.g.:
+
+```
+mdadm -CR /dev/md/imsm0 -e imsm -n2 /dev/nvme0n1 /dev/nvme1n1
+```
+
+Then, create the RAID1 inside of that container. Due to a gap in RHCOS, we create a dummy RAID0 volume in front of the real RAID1 one that we then delete:
+
+```
+# create dummy array
+mdadm -CR /dev/md/dummy -l0 -n2 /dev/imsm0 -z10M --assume-clean
+# create real RAID1 array
+mdadm -CR /dev/md/coreos -l1 -n2 /dev/imsm0
+
+# stop member arrays and delete dummy one
+mdadm -S /dev/md/dummy
+mdadm -S /dev/md/coreos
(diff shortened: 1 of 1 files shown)


jlebon:
The Hybrid RAID instructions there still need to be validated.

jbtrystram (commented):
Looks sane to me, however i am not able to reproduce and verify.
Quick question: in the mdadm raid option, you still need to have a non-raid EFI partition right?  how would you boot otherwise?

travier (approved):

travier:
Feels a bit like magic to me :)

jlebon:
> Quick question: in the mdadm raid option, you still need to have a non-raid EFI partition right? how would you boot otherwise?

No, it's on top of the RAID. The RAID setup is understood by the firmware.

mtkaczyk (commented):
I didn't mention is earlier because there was no time and need too but..

In normal environment we should refer to the MD devices via named links, which are resist in "/dev/md/" directory.

Assumptions that:
 /dev/md127 - container
/dev/md126 - vol0
/dev/md125 - vol1
should work in 99% cases but if there is other MD raid array (/dev/md127 is occupied) then md device number could be different than we hardcoded here.

Should we care? I  Maybe we should warn here? For reference see:
https://www.intel.com/content/dam/support/us/en/documents/memory-and-storage/linux-intel-vroc-userguide-333915.pdf

chapter `4.3.1 Block Device of Intel® VROC RAID Volume`

I don't like how it is implemented in mdadm but any changes in this area are harmful for users.. They sticked to it.

travier:
/title NO-JIRA: faq: add details about RAID installations

travier:
/retitle NO-JIRA: faq: add details about RAID installations

travier:
I'll let Jonathan see if something is needed for https://github.com/openshift/os/pull/1471#pullrequestreview-1978879863 and remove the hold

mtkaczyk:
> One thing is I think now when we re-assemble after deleting the dummy, we can't just use `mdadm -As`, right? We need to do it more directly so that we can specify the md name IIUC. I went with `mdadm -A /dev/md/coreos /dev/md/imsm0`. Let me know if that doesn't look correct.

Personally, I don't like `assemble (-A)` because it is really complicated and it is almost impossible to not have regressions there but in general it should work. I tested the command on my platform and it appears to work.

If you will use `mdadm -As` or `mdadm -I /dev/md/imsm0`, raid device md link will get `_0` suffix, this is something not wanted here (confusing for users).

```
lrwxrwxrwx  1 root root    8 Apr  8 10:26 imsm0 -> ../md127
lrwxrwxrwx  1 root root    8 Apr  8 10:26 vol_0 -> ../md125
lrwxrwxrwx  1 root root    8 Apr  8 10:26 vol1_0 -> ../md126
```

The standard flow, used in system and installers (more reliable, deeply tested) is:
```
mdadm -Eb /dev/md/imsm0 > mdadm_tmp.conf
mdadm -I /dev/md/imsm -c tmp_mdadm.conf
```
We are creating temporary config file, and then passing it to `Incremental` command. With that, mdadm knows that metadata can be considered as trusted and we should not append suffix.

Both ways are correct, second is less risky as it is commonly used. There are no issues in you command.

Btw. if we can inject this conf file to `/etc/mdadm.conf` for both initramfs and regular system, then they will be named same after booting to system. I think that currently they will receive `_0` suffix. This is not an issue unless customers are complying, just FYI :)

jlebon:
Ack, thanks for the confirmation. :)

The temporary config file approach seems interesting. Though since we're modifying the container while it's offline, will `mdadm -I ... -c tmp_mdadm.conf` still work even though the config file will mention the old dummy RAID0 that was deleted?

Yeah, re. the installed system, I don't think customers or the system need to care in practice of the naming used in the actual booted system. But it's good to be aware of the config-related commands.

Anyway, if the docs as written look good to you (and especially if you've tested it), then I'd say let's go with that. Unfortunately, I'm not in a position to retest changes to the procedure very easily so I'm hesitant to tweak it further.

/hold cancel

mtkaczyk:
> The temporary config file approach seems interesting. Though since we're modifying the container while it's offline, will `mdadm -I ... -c tmp_mdadm.conf` still work even though the config file will mention the old dummy RAID0 that was deleted?

Config is just a hint, which says "hey, this array can be deeply trusted because it is described in config".
In config each IMSM array is defined as separated entity, so simply if there is no array metadata on drives (array is deleted) mdadm will not try to assemble this raid volume and config entry will be ignored.

If UUID is still same after deleting one volume then old config will work, if UUID changed you can regenerate it.

PR #1472, opened 2024-03-27
NO-JIRA:  docs/faq: Update RHEL/RHCOS & OCP version table #1472

travier:


mike-nguyen:
/retitle NO-JIRA:  docs/faq: Update RHEL/RHCOS & OCP version table #1472

PR #1473, opened 2024-03-27
COS-2608: Switch from c9s content to RHEL 9.4 beta content

mike-nguyen:
RHEL 9.4 customer beta is now publicly available.

### DIFF
--- a/manifest-rhel-9.4.yaml
+++ b/manifest-rhel-9.4.yaml
@@ -16,13 +16,12 @@ include:
 
 # Starting from here, everything should be specific to RHCOS based on RHEL 9.4 content
 
-# CentOS Stream 9 repos + internal repos for now
 repos:
-  - baseos
-  - appstream
-  - rhel-9.2-fast-datapath
+  - rhel-9.4-baseos
+  - rhel-9.4-appstream
+  - rhel-9.4-fast-datapath
   # Include RHCOS 9 repo for oc, kubelet
-  - rhel-9.2-server-ose-4.16
+  - rhel-9.4-server-ose-4.16
 
 # We include hours/minutes to avoid version number reuse
 automatic-version-prefix: "416.94.<date:%Y%m%d%H%M>"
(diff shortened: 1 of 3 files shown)


mike-nguyen:
Need to check if the 9.4 beta content is mirrored in CI.

manifest-rhel-9.4.yaml
```diff
@@ -113,22 +112,15 @@ postprocess:
...
+ # post-process script
+ - redhat-release
 
-# Packages pinned to specific repos in SCOS 9
+# Packages pinned to specific repos in RHCOS 9
 repo-packages:
-  # We always want the kernel from BaseOS
-  - repo: baseos
-    packages:
-      - kernel
```
travier, line 127: Any specific reason for removing this part?

travier: So this should be good.

mike-nguyen:
Seeing 
```
BdsDxe: loading Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x2,0x0)
BdsDxe: failed to load Boot0001 "UEFI Misc Device" from PciRoot(0x0)/Pci(0x2,0x0): Access Denied
BdsDxe: No bootable option or device was found.
BdsDxe: Press any key to enter the Boot Manager Menu.
```
Which looks related to:
- https://github.com/openshift/os/issues/1237
- https://github.com/openshift/os/pull/1452

I tried building scos on master (not on this PR) and it had the same failure.

sdodson:
As far as I know secureboot shim for 9.4 hasn't been signed yet. I'll see
if I can find a reference tomorrow but fine with snoozing the test for
another three weeks with goal of tracking this down by end of next sprint.

mike-nguyen:

The secure boot tests for c9s should be snoozed or skipped based on issues mentioned above.  Lets override the scos-9-qemu test and address it in a separate PR. This change does not touch anything scos related.

PR #1474, opened 2024-04-03
manifests: include kubernetes.yaml

jbtrystram:
FCOS had shipped ZRAM-generator for a while but it was not shared with RHCOS. Moved the package to a different sub-manifest that would regroup kubernetes-specific things.
The swap on ZRAM would still need to be actually enabled through a config file [1] but that should be done by the MCO.

[1] https://github.com/systemd/zram-generator/blob/main/man/zram-generator.conf.md

ref: https://github.com/coreos/fedora-coreos-config/pull/2937
ref: https://issues.redhat.com/browse/COS-2507


dustymabe:
I'm thinking there is a better way: https://github.com/coreos/fedora-coreos-config/pull/2937#issuecomment-2034922693

PR #1475, opened 2024-04-03
manifest-rhel-9.4: use runc from rhaos repos

haircommander:
as the one in RHEL is missing a required patch, and it was decided we'd use the RHAOS one instead [xref](https://redhat-internal.slack.com/archives/C999USB0D/p1712067961154739)

### DIFF
--- a/manifest-rhel-9.4.yaml
+++ b/manifest-rhel-9.4.yaml
@@ -135,3 +135,6 @@ repo-packages:
       - cri-tools
       - openshift-clients
       - openshift-kubelet
+      # Carry runc from the rhaos branches, as it has a patch needed for OCPBUGS-30806
+      # and it is too late in the RHEL development cycle to update it in time.
+      - runc 


haircommander:
(this can wait on any 9.4 bump pieces needed)

mike-nguyen:
/hold 

While we figure out which path we want to go

travier:
Ideally the runc version in the RHAOS repo is higher than the one in RHEL and gets precedence. But having it forced here also makes sense until we resolve the mismatch. Let's maybe note somewhere that we need to revisit that in the next rebase to RHEL 9.6.

haircommander:
in [slack ](https://redhat-internal.slack.com/archives/C999USB0D/p1712067961154739) it seems consensus is to move towards dropping this and carrying it in RHEL (while in the meantime bumping rhcos version above the RHEL version)

haircommander:
(bumping the runc version solved this in 4.16, so I have removed the bug)

PR #1476, opened 2024-04-04
COS-2746: common.yaml: add dnf

jlebon:
As per discussions in
https://github.com/coreos/fedora-coreos-tracker/issues/1687, we will add dnf to RHCOS so it can be used for derivation.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -176,8 +176,6 @@ check-groups:
 exclude-packages:
   # https://bugzilla.redhat.com/show_bug.cgi?id=1798278
   - subscription-manager
-  # And this one shouldn't come in
-  - dnf
   # https://github.com/coreos/rpm-ostree/pull/1789/files/a0cd999a8acd5b40ec1024a794a642916fbc8ff8#diff-fc2076dc46933204a7a798f544ce3734
   # People need to use `rpm-ostree kargs` instead.
   - grubby
@@ -253,6 +251,8 @@ packages:
  - crypto-policies-scripts
  # For semanage
  - policycoreutils-python-utils
+ # https://github.com/coreos/fedora-coreos-tracker/issues/1687
+ - dnf
 
 packages-x86_64:
   # Temporary add of open-vm-tools. Should be removed when containerized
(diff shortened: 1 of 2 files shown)


jlebon:
Testing this, the biggest roadblock of course is entitlements. This is a known issue tracked at least within the MCO team and not something we need to address at the OS level. But anyway, hacking around that by just adding some internal repos, I sanity-checked derivations using dnf work fine:

```
$ cat src/config/Containerfile.test
FROM overridden
COPY rhel-9.4.repo /etc/yum.repos.d
RUN dnf install -y ltrace && ostree container commit
$ cosa buildextend-layered test
...
$ cosa run
...
[core@cosa-devsh ~]$ sudo -i
[root@cosa-devsh ~]# bootc switch --transport oci-archive $(ls /mnt/workdir/builds/latest/x86_64/*test*.ociarchive)
layers already present: 0; layers needed: 53 (1.2 GB)
 18.49 MiB [████████████████████] (0s) Fetched layer sha256:dfb4c7216493
Queued for next boot: ostree-unverified-image:oci-archive:/mnt/workdir/builds/latest/x86_64/rhcos-416.94.202404041902-0-layered-test.x86_64.ociarchive
  Version: 416.94.202404041902-0
  Digest: sha256:590e833b104874332eb2d5d26c79b3a72de62213c2266f69c39d81a0cac96b1d
[root@cosa-devsh ~]# reboot
...
[core@cosa-devsh ~]$ ltrace
ltrace: too few arguments
Try `ltrace --help' for more information.
```

c4rt0 (approved):
/lgtm

jlebon:
OK interesting. CI is failing because `network-online.target` is getting activated by `dnf-makecache.timer`. That timer unit isn't actually functioning on OSTree systems (it has `ConditionPathExists=!/run/ostree-booted`) but systemd still respects its requirements. I've added a preset for now to disable it outright, but ISTM like that timer should just drop that dependency. It triggers every hour, so it'll eventually converge.

travier:
I've confirmed that this should be the only unit added.
/lgtm

PR #1478, opened 2024-04-09
NO-JIRA: ci: Gate secure boot tests on centos-stream-release

travier:
centos-stream-release now replaces centos-release. Update the package name until CentOS Stream 9 gets the new shim release and finally passes Secure Boot tests.

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -122,6 +122,10 @@ kola_test_qemu() {
     # we need to check if the basic.nvme is available in the list of tests
     local args=""
     local manifest="src/config/manifest.yaml"
+    if [[ -f "src/config.json" ]]; then
+        variant="$(jq --raw-output '."coreos-assembler.config-variant"' 'src/config.json')"
+        manifest="src/config/manifest-${variant}.yaml"
+    fi
     if cosa kola list --json | jq -r '.[].Name' | grep -q "basic.nvme"; then
         if rpm-ostree compose tree --print-only "${manifest}" | jq -r '.packages[]' | grep -q "centos-stream-release"; then
             args+="--denylist-test *.uefi-secure"
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -123,11 +123,11 @@ kola_test_qemu() {
     local args=""
     local manifest="src/config/manifest.yaml"
     if cosa kola list --json | jq -r '.[].Name' | grep -q "basic.nvme"; then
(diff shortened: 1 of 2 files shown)


jlebon (approved):

dustymabe (approved):
/lgtm

jbtrystram (approved):
/lgtm

dustymabe:
`kola-basic/firmware-uefi-secure` is failing in this PR too.. are the settings update in this PR not applied when CI runs for it?

travier:
OK, I'll take another look at this tomorrow

mike-nguyen:
We use `centos-release` to skip some of the secureboot tests.  We will need to change them to `centos-stream-release` [here](https://github.com/openshift/os/blob/8f8cb936a474c2d3bedc2c9c08d867f120a4ee4e/ci/prow-entrypoint.sh#L126) and 
[here](https://github.com/openshift/os/blob/8f8cb936a474c2d3bedc2c9c08d867f120a4ee4e/ci/prow-entrypoint.sh#L130)

mike-nguyen:
In the raw build-log.txt I found

```
parse error: Invalid numeric literal at line 1, column 5
failed to execute cmd-kola: signal: broken pipe
```

Caused by `cosa kola list --json` which will return the command itself `cosa kola list --json` followed by the json output inside a coreos-assembler container.

```
[coreos-assembler]$ cosa kola list --json | head
kola -p qemu list --json --output-dir tmp/kola
[
	{
		"Name": "basic",
		"Platforms": [
			"aws",
			"azure",
			"do",
			"esx",
			"gcp",
```

versus just kola list --json which just returns the JSON.

```
[coreos-assembler]$ kola list --json | head
[
	{
		"Name": "basic",
		"Platforms": [
			"aws",
			"azure",
			"do",
			"esx",
			"gcp",
			"openstack",
```

It doesn't look like it wasn't possible to get into the conditional that adds `--denylist-test *.uefi-secure`

jlebon:
Should be fixed by https://github.com/coreos/coreos-assembler/pull/3775.

jlebon:
> Should be fixed by [coreos/coreos-assembler#3775](https://github.com/coreos/coreos-assembler/pull/3775).

Actually no. That's one issue, but there were two other issues here. I've pushed up some commits inline to fix those.

travier:
Fixes: https://github.com/openshift/os/pull/1458
We really need to land https://github.com/coreos/coreos-assembler/pull/3652

jbtrystram:
/retitle NO-JIRA: ci: Gate secure boot tests on centos-stream-release

travier:
Next is https://github.com/coreos/coreos-assembler/pull/3775

PR #1463, opened 2024-03-17
NO-JIRA: Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Aashish Radhakrishnan (2):
      denylist: snooze ext.config.var-mount.scsi-id on next & next-devel
      Kdump currently fails on kernel-next, for ppc64le

Dusty Mabe (2):
      manifests: add makedumpfile RPM on F41+
      manifests: remove ostree-bls.yaml

HuijingHei (1):
      tests: check default dropins file permissions

Jonathan Lebon (2):
      Add `dnf5` on Fedora 41+
      denylist: only warn if `coreos.boot-mirror` tests fail on ppc64le

Michael Armijo (2):
      denylist: extend snooze for two tests
      denylist: only denylist `coreos.boot-mirror*` on ppc64le

Timothée Ravier (1):
      manifests: Add back more wireless card firmwares

gursewak1997 (1):
      denylist: extend snooze for var-mount.scsi-id

jbtrystram (1):
      Manifests: move zram to system-configuration
```

PR #1479, opened 2024-04-11
COS-2507: Manifests: disable default ZRAM config 

jbtrystram:
In #2937 we moved zram-generator to the system-configuration.yaml
manifest so it's shared with RHCOS.
However in the centOS stream 9 zram-generator-0.3.2 [1] ships a
default config file, resulting in a swap device being created
by default.
This ends up breaking kubelet, until kubernetes 1.30
is released. Removing the default config from the package until
c10s

[1] https://gitlab.com/redhat/centos-stream/rpms/rust-zram-generator/-/blob/c9s/zram-generator.conf?ref_type=heads

----
Builds on #1463

### DIFF
--- a/zram-no-defaults.yaml
+++ b/zram-no-defaults.yaml
@@ -1,0 +1,5 @@
+remove-from-packages:
+  # zram-generator-0.3.2 (shipped in centOS 9) provides a default
+  # zram-generator config, we want to disable it
+  - - zram-generator
+    - "/usr/lib/systemd/zram-generator.conf"
--- a/common.yaml
+++ b/common.yaml
@@ -32,6 +32,13 @@ conditional-include:
   - if: basearch != "s390x"
     # And remove some cruft from grub2
     include: fedora-coreos-config/manifests/grub2-removals.yaml
+    #zram default config is in a subpackage in c10s
+    # Meanwhile, remove the default config from the package
+  - if: osversion == "c9s"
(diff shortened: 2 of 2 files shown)


fedora-coreos-config
c4rt0, line 1: Not sure if this was intentional?

jbtrystram: yes it is. The fedora-config bump brings the zram-generator package

c4rt0: Ok - it's clear now. In that case it /LGTM

c4rt0 (approved):
/LGTM

c4rt0:
Probably it's a good idea to change the title of this PR to something like:
`NO-JIRA: Manifests: disable default ZRAM config`

zram-no-defaults.yaml
```diff
@@ -0,0 +1,5 @@
+remove-from-packages:
+  # zram-generator-0.3.2 (shipped in centOS 9) provides a default
+  # zram-generator config, we want to disable it
+  - - zram-generator
+    - "/usr/lib/systemd/zram-generator.conf"
```
travier, line 5: Nit for a future PR: Missing empty line at the end of the file

travier: I should add https://github.com/coreos/fedora-coreos-config/blob/testing-devel/ci/find-whitespace to this repo.

PR #1485, opened 2024-04-15
NO-JIRA: ci/prow-entrypoint.sh: don't use `grep -q` at the end of pipeline

jlebon:
Unlike `grep`, `grep -q` will exit 0 as soon as a match is found. This will cause whatever is writing into `grep` to hit `SIGPIPE`. And if it's not equipped to handle that signal, it'll be terminated and the overall if-condition will always fail due to `-o pipefail`.

See also https://github.com/ostreedev/ostree/pull/3203.


prestist (approved):
Thank you, this lgtm.

travier (approved):
/lgtm

travier:
/retitle NO-JIRA: ci/prow-entrypoint.sh: don't use `grep -q` at the end of pipeline

PR #1487, opened 2024-04-16

coreosbot-releng:

```
c4rt0 (3):
      denylist: extend snooze for failing kola tests
      denylist: remove ext.config.kdump.crash
      manifests: update nitpick fedora-coreos.yaml

jbtrystram (1):
      manifests: add kexec-utils subpackage for kexec-tools
```


c4rt0 (approved):
/lgtm

PR #1488, opened 2024-04-16
NO-JIRA:docs/faq.md: update iSCSI-related bits

jlebon:
We now support booting from iSCSI. Expand on the different options there.


jbtrystram (approved):
/lgtm

PR #1489, opened 2024-04-17
NO-JIRA: ci: Remove Secure Boot test workaround

travier:
Update kola denylist with the new tests names.

See: https://github.com/openshift/os/issues/1237
See: https://github.com/coreos/coreos-assembler/pull/3652


kola-denylist.yaml
jlebon, line 1: Don't we need to add `basic.uefi-secure` to the denylist for `c9s`?

PR #1490, opened 2024-04-18
COS-2508: docs: explain how to enable swap

jbtrystram:
swap support is now beta2 in kubernetes (as 1.28) and should be stable in 1.30.
Document how to enable the feature on OCP.

Follow https://github.com/openshift/os/pull/1479

### DIFF
--- a/docs/swap.md
+++ b/docs/swap.md
@@ -1,0 +1,87 @@
+# Enable SWAP on ZRAM
+
+Kubernetes 1.30 will introduce support for swap on nodes, here are the steps required to enable it on OCP. 
+Currently (4.16) this requires enabling Techpreview.
+
+## 0 - Requirements
+
+The cluster os image must be at least `416.94.202404120955-0`
+
+## 1 - Enable tech preview
+
+edit the featuregate cluster settings: 
+
+`oc edit featuregate cluster`:
+
+```
+apiVersion: config.openshift.io/v1
+kind: FeatureGate
+metadata:
+  name: cluster
+spec:
+  featureSet: TechPreviewNoUpgrade
+```
+
+For more details see [enabling tech preview](https://docs.openshift.com/container-platform/4.15/nodes/clusters/nodes-cluster-enabling-features.html).
+
+Wait for the config to propagate and all the nodes to become ready again.
+
+## 3 Allow kubelet to start with swap
+
+Create the following file: `99-kubelet-swap-config.yaml`
+```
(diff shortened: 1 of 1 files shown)


jlebon:
It's tempting to add more in-depth docs here, but I would suggest not going too deep into that. That level of depth I think belongs better in the OCP docs. Otherwise we end up with competing docs that need to be maintained in parallel. We have the FAQ, but that's meant to be more shorter answers that ideally link to more comprehensive docs (OCP/RHEL/upstream).

Specifically for this, e.g. we could reformulate it as a FAQ question like:

> ## Q: Does RHCOS support swap?
> 
> It's supported as Tech Preview in 4.16. To enable it, first enable [the tech preview feature gate](link), then enable the `swapBehavior: LimitedSwap` option in a [KubeletConfig](link). swap can be a separate partition (via Ignition), or you can use [swap on zram](link).

Once we have more in-depth docs in the official docs, we can link to that.

jbtrystram:
> It's tempting to add more in-depth docs here, but I would suggest not going too deep into that. That level of depth I think belongs better in the OCP docs. 

100% agree. My idea was to have a draft so someone from the docs could take over. 
Now i realize I should have opened this in the docs repo...

docs/swap.md
```diff
@@ -0,0 +1,87 @@
...
+
+## 4 Verify swap is enabled
+
+From a debug terminal : 
+```
+# zramctl 
+NAME       ALGORITHM DISKSIZE DATA COMPR TOTAL STREAMS MOUNTPOINT
+/dev/zram0 lzo-rle         4G   4K   80B   12K       4 [SWAP]
+```
+
```
aaradhak, line 86: minor : may be we can remove these extra lines at the eof?

jlebon:
> > It's tempting to add more in-depth docs here, but I would suggest not going too deep into that. That level of depth I think belongs better in the OCP docs.
> 
> 100% agree. My idea was to have a draft so someone from the docs could take over. Now i realize I should have opened this in the docs repo...

Normally the docs team adds those and might already be working on it (cc @jldohmann). But yeah, definitely useful to have this written out even just here in PR form as something to start with for the official docs.

travier:
LGTM overall. Agree with Jonathan. Let's make this proper OCP docs.
/hold

jbtrystram:
@jldohmann was this added to the OCP docs ? I see a couple of PRs but nothing related to ZRAM swap

PR #1496, opened 2024-04-25

coreosbot-releng:

```
Aashish Radhakrishnan (3):
      Move to Fedora Linux 40
      Updating all container references to fedora 40
      tests/kola: use toolbox container image for selinux.podman-tmpfs-context

Dusty Mabe (3):
      overlay.d/15fcos: update link to FAQ
      overlay.d/05core: don't order coreos-enable-network after basic.target
      overlay.d/05core: drop Before=dracut-initqueue from coreos-multipath-wait.target
```


jbtrystram (commented):
There is nothing pressing for RHCOS in there, let's wait for the changes to accumulate a bit more

PR #1497, opened 2024-04-30

coreosbot-releng:

```
Michael Armijo (1):
      denylist: extend snooze for `coreos.ignition.ssh.key`
```


ravanelli (approved):
/lgtm

PR #1502, opened 2024-05-06
NO-JIRA: variants: simplify

jlebon:
This repo is really confusing to work with because of all the various
tiers of variants we have.

In practice, our production pipelines always specify a concrete variant
to build because the switchover between e.g. 9.2 and 9.4 happens on the
ART side, not RHCOS side. And even in CI, since the script that gets
called by Prow lives here, we can easily control which concrete variant
gets built.

So overall, we don't gain much from trying to have symbolic versionless
variants, but it adds cognitive overhead trying to understand it all.

This patch greatly simplifies things by getting rid of the `scos` and
`rhel-coreos-9` variants. Now, we *only* have concrete variants.
Document them in the README.

The only symbolic links left are the canonical variantless ones, which
determine the default variant that gets built if no `--variant` switch
is passed to `cosa init`.

This is also prep for #799, which will add more concrete variants that
do not bake in the OpenShift components.


c9s.repo
```diff
@@ -1,4 +1,4 @@
-[baseos]
+[c9s-baseos]
```
travier, line 1: From memory that breaks the extension container builds

travier (commented):
LGTM apart from the repo name change for C9S

travier:
CI fixes in https://github.com/openshift/release/pull/51750

c9s.repo
jlebon: Seems to work fine locally. It's also tested in CI here. Possibly whatever broke on it before is now fixed?

dustymabe (approved):
/lgtm

marmijo (approved):

dustymabe:
some fallout from this is that older branches `ext.config.rpm-ostree.replace-rt-kernel`test are pulling from the `master` branch and then sed replacing `[baseos]` to enable the repo and now they are failing:

```
=== RUN   ext.config.rpm-ostree.replace-rt-kernel
systemctl status kola-runext.service:
× kola-runext.service
     Loaded: loaded (/etc/systemd/system/kola-runext.service; static)
     Active: failed (Result: exit-code) since Thu 2024-05-09 22:09:20 UTC; 381ms ago
   Duration: 2.818s
    Process: 1939 ExecStart=/usr/local/bin/kola-runext-replace-rt-kernel (code=exited, status=1/FAILURE)
   Main PID: 1939 (code=exited, status=1/FAILURE)
        CPU: 154ms

May 09 22:09:17 qemu0 systemd[1]: Started kola-runext.service.
May 09 22:09:17 qemu0 kola-runext-replace-rt-kernel[1945]: + rm -rf '/etc/yum.repos.d/*'
May 09 22:09:17 qemu0 kola-runext-replace-rt-kernel[1947]: + curl -sSLf https://raw.githubusercontent.com/openshift/os/master/c9s.repo -o /etc/yum.repos.d/cs.repo
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1950]: + curl -sSLf https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1953]: + sed -i 's|^gpgkey.*|gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official|' /etc/yum.repos.d/cs.repo
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1939]: Testing overriding with CentOS Stream kernel
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1955]: + sed -i s/enabled=1/enabled=0/g /etc/yum.repos.d/cs.repo
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1957]: + sed -i '/\[baseos\]/,/^ *\[/ s/enabled=0/enabled=1/' /etc/yum.repos.d/cs.repo
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1959]: + sed -i '/\[appstream\]/,/^ *\[/ s/enabled=0/enabled=1/' /etc/yum.repos.d/cs.repo
May 09 22:09:18 qemu0 kola-runext-replace-rt-kernel[1961]: + rpm-ostree override replace --experimental --from repo=baseos kernel kernel-core kernel-modules kernel-modules-extra kernel-modules-core
May 09 22:09:20 qemu0 kola-runext-replace-rt-kernel[1962]: Checking out tree 2b62a0d...done
May 09 22:09:20 qemu0 kola-runext-replace-rt-kernel[1962]: No enabled rpm-md repositories.
May 09 22:09:20 qemu0 kola-runext-replace-rt-kernel[1962]: Importing rpm-md...done
May 09 22:09:20 qemu0 kola-runext-replace-rt-kernel[1962]: error: No matches for 'kernel' in repo 'baseos'
May 09 22:09:20 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE
```

PR #1456, opened 2024-03-07
COS-2613: kola: fix rhaos-pkgs-match-openshift

c4rt0:
See: https://github.com/openshift/os/issues/1427

PR #1516, opened 2024-05-27
COS-2613: [release-4.13]: kola: fix rhaos-pkgs-match-openshift

c4rt0:
Due to errors in CI of automated cherry-picks, this is a manual alternative combining 2 PR's.

See: https://github.com/openshift/os/pull/1456#issuecomment-2122930652


jlebon:
Can you add the original commit that was cherry-picked? When using `git cherry-pick`, you can use `-x` to have it do it for you.

Issue #1427, opened 2024-01-23
rhaos-pkgs-match-openshift test likely not working properly


travier:
See: https://github.com/openshift/os/blob/master/tests/kola/version/rhaos-pkgs-match-openshift

Our RHCOS images have non matching `rhaos` packages and the test don't fail.

See: https://issues.redhat.com/browse/COS-2613

travier:
See https://github.com/openshift/os/pull/1448/commits/d328db94854a26085c30a257039b2a8b8097031d for an example of using grep in an if statement.

jlebon:
Yeah, we should fix this. It will help us catch the case where we have newer versions in rhaos tag X than in the X+1 tag. (We'd still ship the one from X even on the X+1 stream, but it looks confusing.)

(I will note though that with #799, we'll basically eventually want all of the non-OpenShift packages we want in the base compose to be in RHEL proper.)

travier:
Related to https://issues.redhat.com/browse/OCPNODE-1986

c4rt0:
@travier - can we close this issue? #1456 with the fix was successfully merged.

travier:
Let's close it once we have done all the backports.

c4rt0:
> Let's close it once we have done all the backports.

| rhcos | Status |
| -- | -- |
| [4.16](https://github.com/openshift/os/pull/1456) | Merged |
| [4.15](https://github.com/openshift/os/pull/1500) | Merged | 
| [4.14](https://github.com/openshift/os/pull/1512) | Open |

travier:
We'll skip backporting to 4.12. Thanks Adam!

PR #1517, opened 2024-05-27
COS-2613: [release-4.13]: kola: fix rhaos-pkgs-match-openshift

c4rt0:
See: https://github.com/openshift/os/issues/1427

Co-authored-by: Clement Verna <cverna@users.noreply.github.com>
(cherry picked from commit https://github.com/openshift/os/commit/219d60f52eece0b09339d0f8568d06cf2553a575)

ci/prow-entrypoint: tell git repo under /src is safe
That directory is part of the image and so not owned by us. Likely the
base image used for the src container was updated recently to include
newer git.

(cherry picked from commit https://github.com/openshift/os/commit/630d3f3d9c530c238e03c93210598101eafca266)

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -181,13 +181,18 @@ validate() {
     workdir="$(mktemp -d)"
     echo "Using $workdir as working directory"
 
+    # for `git config --global` below
+    export HOME=${workdir}
+
     # Figure out if we are running from the COSA image or directly from the Prow src image
     if [[ -d /src/github.com/openshift/os ]]; then
         cd "$workdir"
+        git config --global --add safe.directory /src/github.com/openshift/os
         git clone /src/github.com/openshift/os os
     elif [[ -d ./.git ]]; then
         srcdir="${PWD}"
         cd "$workdir"
+        git config --global --add safe.directory "${srcdir}/.git"
         git clone "${srcdir}" os
     else
         echo "Could not found source directory"
--- a/tests/kola/version/rhaos-pkgs-match-openshift
+++ b/tests/kola/version/rhaos-pkgs-match-openshift
@@ -8,7 +8,11 @@ set -xeuo pipefail
 
 . $KOLA_EXT_DATA/commonlib.sh
 
-# Check that rhaos packages do not match the OpenShift version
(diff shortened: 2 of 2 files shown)


c4rt0:
https://github.com/openshift/os/pull/1516#issuecomment-2133872485

@jlebon - I messed up commits so I created a new branch, and this new PR. I used `cherry-pick -x` as suggested, but since it's a combination of two commits (one developed only in `4.14` + the one I was working on in `4.16`) I had to add the description here manually. Additionally in @HuijingHei [PR](https://github.com/openshift/os/commit/7aef742c74d158d6af500126a25ef7586c92e808), multiple files were modified. I have included only the one file required for this cherry-pick.

HuijingHei:
Overall looks good to me, just one small question:
kola `description` is based on [PR](https://github.com/coreos/coreos-assembler/pull/3425)(which is 4.14+), and not backport in 4.13(see https://github.com/coreos/coreos-assembler/compare/rhcos-4.13...main), maybe the change for `description` is not needed?

c4rt0:
> Overall looks good to me, just one small question: kola `description` is based on [PR](https://github.com/coreos/coreos-assembler/pull/3425)(which is 4.14+), and not backport in 4.13(see [coreos/coreos-assembler@rhcos-4.13...main](https://github.com/coreos/coreos-assembler/compare/rhcos-4.13...main)), maybe the change for `description` is not needed?

Thank's @HuijingHei - when trying to only cherry-pick the [main commit](https://github.com/openshift/os/pull/1517/commits/afc5dd15e6a2aa7f4a013ac61143853d39e59cb2) I was coming across merge conflicts. This is where I had a flash thought and decided, that seeing the full development cycle of a file also has its benefits. Now I might be very wrong here, but when trying to resolve conflicts through VS Code I was proposed to accept suggested changes, which had nothing to do with the original PRs... I quickly switched to VIM and implemented it this way. Please let me know if you think I should approach this another way.

jbtrystram:
@c4rt0 a good habit is to cherry-pick the "most compatible" commit. So for 4.15 you would cherry-pick from master. Once merged, you'd cherry-pick from 4.15 branch to 4.14.. and so on. 

here you should drop the first commit. I agree the git conflict resolution can be confusing ! :)

HuijingHei (approved):
lgtm

c4rt0:
@jlebon Could you also update the `jira/valid-bug` label here?

Issue #1519, opened 2024-05-29
Remove NVMe-related UUID from images `/usr/etc/nvme/hostid` `/usr/etc/nvme/hostnqn`


travier:
See: https://github.com/openshift/os/pull/1445#issuecomment-2115182903

There are apparently hardcoded NMVE ids in the images right now and we should remove those to make sure those are unique and generated per-node.

travier:
See https://issues.redhat.com/browse/OCPBUGS-34629

dustymabe:
I guess either way we don't want the same ID baked in so what we have right now is clearly wrong.

We can start with deleting them and then if we need to generate them on first boot for EL9 we can create a service to do that.

PR #1520, opened 2024-05-29
manifests: Clean-up NVME-related IDs

travier:
See: https://github.com/openshift/os/issues/1519

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -150,6 +150,12 @@ postprocess:
     # generating it.
     ln -sr /usr/share/zoneinfo/UTC /etc/localtime
 
+  - |
+    #!/usr/bin/env bash
+    set -xeo pipefail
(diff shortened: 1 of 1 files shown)


jlebon:
Didn't dig into this but two things offhand:
1. What e.g. scriptlet exactly creates these? Maybe there's a cleaner way to do this, including working with the package maintainers directly.
2. We need to ensure that the identifiers are correctly generated on first boot (or whenever they're necessary).

travier:
100% agree.
/hold

cgwalters:
- We should have a tracker bug against the package
- Workarounds like this ideally land in the bootc base image and are inherited from there

dustymabe:
obsoleted by https://github.com/openshift/os/pull/1722

PR #1522, opened 2024-06-06
OCPBUGS-33124: coreos-cryptfs: drop growpart call

madhu-pillai:
It was weird how we would growpart *on every boot* in the old-style LUKS
code path. In comparison, on new installs nowadays, we only growpart
once on first boot and never again. Since we don't support new installs
that use the old-style LUKS scheme anymore, let's just drop the growpart
call.

This also fixes some issues we've been seeing on s390x systems using
DASD devices where it would make the device temporarily unaccessible
due to sfdisk not understanding DASD partitioning, and some locking that
growpart does that lingers and breaks the following `blockdev` call.

Fixes: https://issues.redhat.com/browse/OCPBUGS-33124

### DIFF
--- a/overlay.d/05rhcos/usr/libexec/coreos-cryptfs
+++ b/overlay.d/05rhcos/usr/libexec/coreos-cryptfs
@@ -125,18 +125,10 @@ try_open() {
         # this avoids the cryptsetup performance penalty.
         msg "${dev} is not encrypted. Device will be mounted as device-mapper linear target."
 
-        # We inline a growpart here since it's simpler than trying to reconfigure dm-linear
-        local major minor devpath partition parent_path parent_device
-        major=$((0x$(stat -c '%t' "${dev}")))
-        minor=$((0x$(stat -c '%T' "${dev}")))
-        devpath=$(realpath "/sys/dev/block/${major}:${minor}")
-        partition=$(cat "$devpath/partition")
-        parent_path=$(dirname "$devpath")
-        parent_device=/dev/$(basename "${parent_path}")
-
-        # TODO: make this idempotent, and don't error out if
-        # we can't resize.
-        growpart "${parent_device}" "${partition}" || true
+        # Note: we used to growpart here, but don't anymore. That should only
+        # have been done on firstboot, not every boot. Since we don't support
+        # this LUKS setup anymore on new installs, we dropped it. That also
+        # fixes some growpart-related issues (see OCPBUGS-33124).
 
         local dev_size=$(($(blockdev --getsize "${dev}") - "${LUKS_HEADER_OFFSET_SECTORS}"))
         echo "0 ${dev_size} linear ${dev} ${LUKS_HEADER_OFFSET_SECTORS}" \


overlay.d/05rhcos/usr/libexec/coreos-cryptfs
```diff
@@ -136,7 +136,9 @@ try_open() {
 
         # TODO: make this idempotent, and don't error out if
         # we can't resize.
-        growpart "${parent_device}" "${partition}" || true
+        # Investigation OCPBUGS-33124 proved the growpart causes the partition disappears.
+        # Hence removed from coreos-cryptfs.
+        # growpart "${parent_device}" "${partition}" || true
```
jlebon, line 141: No need to comment it out. Let's just drop that line and that's it.

jlebon (commented):
Can you expand on the commit message? I think it's important to capture why it's safe to remove this `growpart` call.

overlay.d/05rhcos/usr/libexec/coreos-cryptfs
madhu-pillai: Added the commit message and dropped the growpart.

jlebon:
Pushed an update here. Removed more dead code and expanded on the commit message.

/lgtm

Fixed up the linked Jira card.

/jira refresh

That said, CI will likely fail here due to https://github.com/openshift/os/issues/1523 but we can force it in since it doesn't cover this path at all. As a final sanity-check, @madhu-pillai can you test this latest version manually?

jlebon:
As expected, CI is failing on https://github.com/openshift/os/issues/1523.

That said, adding a hold for

> As a final sanity-check, @madhu-pillai can you test this latest version manually?

Feel free to drop the hold (with `/hold cancel`) once you did that.

madhu-pillai:
hi @jlebon,
 I have tested it. I can reproduce the issue from 4.5 -> 4.12 -> 4.17
 After applying  the fix , I can upgrade 4.5 -> 4.12 -> 4.17
 
```
 [core@master-1 ~]$ sudo rpm-ostree status
State: idle
Deployments:
* 573f47ef426ff0687c28669ecc720dfc483a8a3f5962eb4de6b62237b2982938
                  Version: 412.90.202301101512-0 (2023-01-10T15:44:38Z)

  658da8df8993a58836ba73eff84360fa6ccd61bb23b4b4f1b89048200992e0df
                  Version: 45.82.202007101157-0 (2020-07-10T12:05:50Z)


[core@master-1 ~]$ sudo rpm-ostree status
State: idle
Deployments:
● ostree-unverified-image:oci-archive:/srv/rhcos-417.94.202406200502-0-ostree.s390x.ociarchive
                   Digest: sha256:4c578e83be70c60e372e8c76108f38e8e30067a0d195298aad2df71d6b28bab3
                  Version: 417.94.202406200502-0 (2024-06-20T05:24:17Z)

  573f47ef426ff0687c28669ecc720dfc483a8a3f5962eb4de6b62237b2982938
                  Version: 412.90.202301101512-0 (2023-01-10T15:44:38Z)

```

Issue #1523, opened 2024-06-06
[4.17-9.4] `crio.base` failing with "openssl: can't retrieve OpenSSL version"


nikita-dubrovskii:
Tests fail with:

```
    core : PWD=/var/home/core ; USER=root ; COMMAND=/bin/crictl exec 5907f463f2f37a60afb9c29a374659aab9749109144175f7f5379e481fbf4076 bash -c 'sleep 25 && echo PASS > /tmp/test/restart-test'
pam_unix(sudo:session): session opened for user root(uid=0) by core(uid=1000)
run-runc-5907f463f2f37a60afb9c29a374659aab9749109144175f7f5379e481fbf4076-runc.tA3M9r.mount: Deactivated successfully.
panic: opensslcrypto: can't initialize OpenSSL : openssl: can't retrieve OpenSSL version

goroutine 1 gp=0xc0000061c0 m=0 mp=0x55780958a1c0 [running]:
panic({0x55780919e680?, 0xc00002ee10?})
        /usr/lib/golang/src/runtime/panic.go:779 +0x158 fp=0xc0000f3ce8 sp=0xc0000f3c38 pc=0x557808cdaa78
crypto/internal/backend.init.0()
        /usr/lib/golang/src/crypto/internal/backend/openssl.go:50 +0x26c fp=0xc0000f3e20 sp=0xc0000f3ce8 pc=0x557808ee7a4c
runtime.doInit1(0x5578095722f0)
        /usr/lib/golang/src/runtime/proc.go:7176 +0xea fp=0xc0000f3f50 sp=0xc0000f3e20 pc=0x557808ceca0a
runtime.doInit(...)
        /usr/lib/golang/src/runtime/proc.go:7143
runtime.main()
        /usr/lib/golang/src/runtime/proc.go:253 +0x357 fp=0xc0000f3fe0 sp=0xc0000f3f50 pc=0x557808cde157
runtime.goexit({})
        /usr/lib/golang/src/runtime/asm_amd64.s:1695 +0x1 fp=0xc0000f3fe8 sp=0xc0000f3fe0 pc=0x557808d11361
...
```

[journal.txt](https://github.com/user-attachments/files/15636972/journal.txt)

nikita-dubrovskii:
As i can see, only those packages between successful and failed builds have different version:
```
openshift-clients-4.17.0-202406040213.p0.gb9859d5.assembly.stream.el9.x86_64
[...]

PR #1531, opened 2024-06-18
tests/kola/version/rhel-major-version: adapt for pure RHEL variants

jlebon:
This probably should've been part of 4e09d27 ("Make c9s and rhel-9.4 variants be pure C9S/RHEL 9.4 content"), which edited that same test for the for pure CentOS Stream variant.


marmijo (approved):
LGTM

jlebon:
This is going to keep failing on https://github.com/openshift/os/issues/1523. But `ext.config.version.rhel-major-version` passed in both, which is the only test changed here. So:

That said, this still needs an actual `/lgtm`.

PR #1533, opened 2024-06-20
 [release-4.16] OCPBUGS-33124: coreos-cryptfs: drop growpart call

madhu-pillai:
It was weird how we would growpart *on every boot* in the old-style LUKS
    code path. In comparison, on new installs nowadays, we only growpart
    once on first boot and never again. Since we don't support new installs
    that use the old-style LUKS scheme anymore, let's just drop the growpart
    call.

    This also fixes some issues we've been seeing on s390x systems using
    DASD devices where it would make the device temporarily unaccessible
    due to sfdisk not understanding DASD partitioning, and some locking that
    growpart does that lingers and breaks the following `blockdev` call.

    Fixes: https://issues.redhat.com/browse/OCPBUGS-33124


jlebon:
This PR should only be one single commit cherry-picked, which is the one from https://github.com/openshift/os/pull/1522.

Also, new Jira tickets need to be created for each backport. You can use the `@rhcos-bug-bot` internally on Slack to help with this.

madhu-pillai:
Refer https://github.com/openshift/os/pull/1534. Closing this as duplicate.

Issue #1540, opened 2024-06-27
fips.enable failing in c9s


jlebon:
```
--- [31mFAIL[0m: fips.enable (47.25s)
        harness.go:1823: mach.Start() failed: machine 8330598d-4529-45ff-9568-efe2b9636806 entered emergency.target in initramfs
        harness.go:1740: Found Go panic (opensslcrypto: can't enable FIPS mode for OpenSSL 3.2.2 4 Jun 2024: OSSL_PROVIDER_try_load
```

```
         Starting [0;1;39mIgnition (fetch-offline)[0m...
[    9.238794] ignition[763]: panic: opensslcrypto: can't enable FIPS mode for OpenSSL 3.2.2 4 Jun 2024: OSSL_PROVIDER_try_load
[    9.246143] ignition[763]: openssl error(s):
[    9.249212] ignition[763]: error:12800067:DSO support routines::could not load the shared library
[    9.255057] ignition[763]: 	crypto/dso/dso_dlfcn.c:118
[    9.258551] ignition[763]: error:12800067:DSO support routines::could not load the shared library
[    9.264293] ignition[763]: 	crypto/dso/dso_lib.c:147
[    9.267750] ignition[763]: error:07880025:common libcrypto routines::reason(37)
[    9.272762] ignition[763]: 	crypto/provider_core.c:949
[    9.276444] ignition[763]: goroutine 1 gp=0xc0000081c0 m=0 mp=0x560b55045420 [running]:
[    9.281902] ignition[763]: panic({0x560b545e7ca0?, 0xc0001be110?})
[    9.286327] ignition[763]: 	/usr/lib/golang/src/runtime/panic.go:779 +0x158 fp=0xc00053fce8 sp=0xc00053fc38 pc=0x560b53572318
[    9.293619] ignition[763]: crypto/internal/backend.init.0()
[    9.298779] ignition[763]: 	/usr/lib/golang/src/crypto/internal/backend/openssl.go:72 +0x21f fp=0xc00053fe20 sp=0xc00053fce8 pc=0x560b536d5c7f
[    9.307026] ignition[763]: runtime.doInit1(0x560b54fdf500)
[    9.310822] ignition[763]: 	/usr/lib/golang/src/runtime/proc.go:7176 +0xea fp=0xc00053ff50 sp=0xc00053fe20 pc=0x560b535842ea
[    9.317976] ignition[763]: runtime.doInit(...)
[...]

PR #1544, opened 2024-07-10
NO-JIRA: kola-denylist: add fips.* and ext.config.shared.networking.* on c9s

jlebon:
We're currently hitting OpenSSL issues there:
https://github.com/openshift/os/issues/1540

---

kola-denylist: add ext.config.shared.networking.* on c9s

A bunch of tests are failing because `systemd-network-generator.service`
is hitting SELinux denials:

https://issues.redhat.com/browse/RHEL-47033

Some of these tests are testing things that don't strictly need that
service, but test kargs that trigger it anyway. Other tests do (e.g.
`ifname=`) and so that functionality is indeed currently broken on c9s.

That said, there's continuing requests to have preliminary c9s
bootimages for testing. So let's not block on this issue for now.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -23,6 +23,11 @@
   osversion:
     - c9s
 
+- pattern: ext.config.shared.networking.*
+  tracker: https://issues.redhat.com/browse/RHEL-47033
+  osversion:
+    - c9s
+
 # we're missing a cri-o rebuild for 4.17, which blocks on buildroot issues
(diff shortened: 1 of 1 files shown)


jlebon:
CI failing on systemd-network-generator, tracked as part of https://github.com/openshift/os/issues/1514 (specifically https://issues.redhat.com/browse/RHEL-47033).

jlebon:
We don't need the second commit here with the SELinux workaround in https://github.com/openshift/os/pull/1556. I've inlined the first commit here in that PR, so we can close this.

Issue #1551, opened 2024-07-16
Prow tests are timing out


jlebon:
We're seeing this on all branches, so this is likely related to something on the infra side. E.g. see https://github.com/openshift/os/pull/1550, which is purely about running *less* tests and yet it hit:

```
 INFO[2024-07-16T05:35:40Z] Executing test rhcos-92-build-test-qemu      
{"component":"entrypoint","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:169","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.ExecuteProcess","level":"error","msg":"Process did not finish before 4h0m0s timeout","severity":"error","time":"2024-07-16T09:35:38Z"}
INFO[2024-07-16T09:35:38Z] Received signal.                              signal=interrupt
INFO[2024-07-16T09:35:38Z] error: Process interrupted with signal interrupt, cancelling execution... 
INFO[2024-07-16T09:35:38Z] cleanup: Deleting test pod rhcos-92-build-test-qemu 
INFO[2024-07-16T09:35:38Z] Ran for 4h0m0s                               
```

The archived logs don't really show at what point of the test we timed out.

jlebon:
A neat thing about Prow is that you can access the temporary namespace that gets created in which the tests are running. Right now, it's not clear on what exactly it's timing out, and the logs appear incomplete. So to debug this, one can start a test (e.g. on an existing PR or a dummy PR), and then log into the cluster it's running on and attach to the test pods getting launched to follow what's happening more closely.

Our jobs run on build02: https://github.com/openshift/release/blob/e4b98d1804ed4cd88854a5977065209f9d1ebc83/ci-operator/config/openshift/os/openshift-os-master.yaml#L92-L119. More info on available clusters: https://docs.ci.openshift.org/docs/getting-started/useful-links/

marmijo:
Here's what the Event log shows in the pod:
```
0/39 nodes are available: 1 node(s) were unschedulable, 14 node(s) had untolerated taint {node-role.kubernetes.io/ci-tests-worker: }, 
2 node(s) had untolerated taint {node-role.kubernetes.io/ci-longtests-worker: ci-longtests-worker}, 
3 node(s) had untolerated taint {node-role.kubernetes.io/ci-prowjobs-worker: ci-prowjobs-worker}, 
3 node(s) had untolerated taint {node-role.kubernetes.io/infra: }, 
[...]

PR #1552, opened 2024-07-16
okd-scos: Remove override for selinux-policy

Prashanth684:
For now we build with the latest selinux-policy as the `selinux-policy-38.1.36-1.el9` package is not available in the mirrors anymore. This needs to be fixed before we create c9s bootimages, for CI to pass.


overrides-c9s.yaml
```diff
@@ -3,10 +3,13 @@
...
+#  - c9s-appstream-mirror
 
-packages:
+#packages:
   # https://github.com/openshift/os/issues/1514
-  - selinux-policy-38.1.36-1.el9
+  # For now we build with the latest selinux as this package is not available in the
+  # mirrors anymore. This needs to be fixed before we create c9s bootimages, for CI
+  # to pass.
+  #  - selinux-policy-38.1.36-1.el9
```
jlebon, line 15: You can just nuke all these lines and only leave `#packages:`.

jlebon:
But fine as is too.

/lgtm
/approve

We'll have to override CI to get this in.

overrides-c9s.yaml
Prashanth684: done.

Prashanth684:
@jlebon mind overriding the ci-jobs? looks like it can't find a node to run the job?

jlebon:
Yup, that's https://github.com/openshift/os/issues/1551.

PR #1556, opened 2024-07-18
OKD-223: Load custom SELinux rules in SCOS and workaround afterburn failures

aleskandro:
Recent changes in the SELinux policy to make more service confined has
broken a lot of our code. The SELinux team is working through relaxing
the policy, but in the meantime, let's revert back the affected types to
permissive mode:

1. afterburn fail when trying to write to `/run`, `/run/metadata` and
   `/home/$user/.ssh`.
   See: https://issues.redhat.com/browse/RHEL-49735
2. coreos-installer installation fails due to various denials.
   See: https://issues.redhat.com/browse/RHEL-38614
3. network functionality that rely on systemd-network-generator is
   broken due to the latter being unable to create temporary files.
   See: https://issues.redhat.com/browse/RHEL-47033

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -18,6 +18,11 @@
   osversion:
     - c9s
 
+- pattern: fips.*
+  tracker: https://github.com/openshift/os/issues/1540
+  osversion:
+    - c9s
+
 # we're missing a cri-o rebuild for 4.17, which blocks on buildroot issues
 - pattern: ext.config.version.rhaos-pkgs-match-openshift
   tracker: https://issues.redhat.com/browse/RHEL-35883
--- a/manifest-c9s.yaml
+++ b/manifest-c9s.yaml
@@ -26,3 +26,23 @@ mutate-os-release: "9"
 
 packages:
  - centos-stream-release
+
+postprocess:
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    cat > /tmp/scos-workarounds.cil << EOF
+    ; https://issues.redhat.com/browse/RHEL-49735
+    (typeattributeset cil_gen_require afterburn_t)
+    (typepermissive afterburn_t)
+
+    ; https://issues.redhat.com/browse/RHEL-38614
+    (typeattributeset cil_gen_require coreos_installer_t)
+    (typepermissive coreos_installer_t)
+
+    ; https://issues.redhat.com/browse/RHEL-47033
+    (typeattributeset cil_gen_require systemd_network_generator_t)
+    (typepermissive systemd_network_generator_t)
+    EOF
+    /usr/sbin/semodule -i /tmp/scos-workarounds.cil
+    rm /tmp/scos-workarounds.cil


overlay.d/50scos/usr/lib/okd/selinux/afterburn.cil
```diff
@@ -0,0 +1,7 @@
+(typeattributeset cil_gen_require var_run_t)
```
Prashanth684, line 1: will these rules also persist after the bootimage pivots to the scos image?

aleskandro: yes

aleskandro: I'd also say that we could delete this file once a proper fix lands in CS9? I didn't reproduce in a CS9 fresh env with no other modifications as the ones here, but it seems a bug to file there for afterburn or the selinux-policy package you mentioned in the issue.

Prashanth684: once we have scos publicly available, yes we would need this. But by that time I hope that we can get the actual issue with selinux-policy sorted out.

aleskandro: https://github.com/fedora-selinux/selinux-policy/pull/1362/files might be the fix to downstream?

aleskandro: And https://github.com/fedora-selinux/selinux-policy/pull/2000/files

aleskandro:
> Hum, why is this not failing in FCOS?

I think it is because of previous patches to SELinux-policy that are already available in FCOS.

overlay.d/50scos/usr/lib/okd/selinux/afterburn.cil
aleskandro: @travier see the two PRs above for https://github.com/openshift/os/pull/1556#issuecomment-2239291679

aleskandro:
Oh here it can be: 

https://github.com/zpytela/selinux-policy/commit/18e59aad0eddd221973f77001fb4e7df6b34e288 removed the permissive domain to c9s, but non-rawhide fedora has it yet: https://github.com/zpytela/selinux-policy/blob/f40/policy/modules/contrib/afterburn.te#L15

travier:
How will all of this work in the initrd? We really need to fix this in Rawhide first.

I'm not convinced we should carry SELinux policy modules here. Those units were in permissive mode before, let's make them permissive again instead.

aleskandro:
> How will all of this work in the initrd? We really need to fix this in Rawhide first.
> 
> I'm not convinced we should carry SELinux policy modules here. Those domains were in permissive mode before, let's make them permissive again instead.

I've changed the module to set the afterburn_t domain in permissive mode. I could do the same for the coreos installer one so that we have a workaround for the other issue too, until they are are solved upstream and we can revert the two policies' commits. I would still continue to propose having that unit that loops through a folder in /usr for applying custom SELinux rules that can be needed for time to time. 

We have something similar in okd-machine-os for the FCOS variant.

Prashanth684:
> > How will all of this work in the initrd? We really need to fix this in Rawhide first.
> > I'm not convinced we should carry SELinux policy modules here. Those domains were in permissive mode before, let's make them permissive again instead.
> 
> I've changed the module to set the afterburn_t domain in permissive mode. I could do the same for the coreos installer one so that we have a workaround for the other issue too, until they are are solved upstream and we can revert the two policies' commits. I would still continue to propose having that unit that loops through a folder in /usr for applying custom SELinux rules that can be needed from time to time.
> 
> We have something similar in okd-machine-os for the FCOS variant.

@travier could we get an approval for this?

jlebon:
Note that in FCOS rawhide, we're currently pinning selinux-policy due to other issues: https://github.com/coreos/fedora-coreos-config/blob/441ccdd5cc3f2e3ac0cb5e8e0bcf71b301a51c7e/manifest-lock.overrides.yaml#L17-L26

But even then, the policy between c9s and rawhide are different and not mean to be kept in sync AIUI. See the conversation in https://github.com/openshift/os/issues/1514.\

I sympathize with this. I'm not opposed in theory, but it's also a much more invasive workaround that might have subtle repercussions. Note also that not many people on the team are familiar with maintaining SELinux policies.

What we really need of course is better CI integration with c9s. This should be a big part of the bootable containers effort.

Short-term, I guess we can just host the last working version RPM somewhere else and revert https://github.com/openshift/os/pull/1552 ? (Or, if you have the right channels, try to get a new tag created in the CentOS Stream koji instance with tag2distrepo set up so we can just tag older packages we need in there.)

Prashanth684:
> I sympathize with this. I'm not opposed in theory, but it's also a much more invasive workaround that might have subtle repercussions. Note also that not many people on the team are familiar with maintaining SELinux policies.

What repercussions are we worried about?

> Short-term, I guess we can just host the last working version RPM somewhere else and revert #1552 ? (Or, if you have the right channels, try to get a new tag created in the CentOS Stream koji instance with tag2distrepo set up so we can just tag older packages we need in there.)

Isn't this equivalent to having a workaround here?

travier:
As I asked before:

> How will all of this work in the initrd?

I don't think this will work as this unit runs after the initrd, only in the real root.

travier:
/hold
As I don't think this is the right fix.

aleskandro:
> As I asked before:
> 
> > How will all of this work in the initrd?
> 
> I don't think this will work as this unit runs after the initrd, only in the real root.

@travier this branch is already being used in the releases that turned green from Sunday thanks to this fix (for AWS). 

- Do we really need this after rebasing from the FCOS boot image (and initrd) to SCOS? 
- Which cases in particular? 
- How can installing the previous RPM in the base OS tree resolve this problem for initrd too?
- Is there a temporary solution that we can achieve to ensure the releases get green again before the actual and permanent fix downstreams to CS9?

> Note that in FCOS rawhide, we're currently pinning selinux-policy due to other issues: https://github.com/coreos/fedora-coreos-config/blob/441ccdd5cc3f2e3ac0cb5e8e0bcf71b301a51c7e/manifest-lock.overrides.yaml#L17-L26

> But even then, the policy between c9s and rawhide are different and not mean to be kept in sync AIUI. See the conversation in https://github.com/openshift/os/issues/1514.%5C

> I sympathize with this. I'm not opposed in theory, but it's also a much more invasive workaround that might have subtle repercussions. Note also that not many people on the team are familiar with maintaining SELinux policies.

> What we really need of course is better CI integration with c9s. This should be a big part of the bootable containers effort.

> Short-term, I guess we can just host the last working version RPM somewhere else and revert https://github.com/openshift/os/pull/1552 ? (Or, if you have the right channels, try to get a new tag created in the CentOS Stream koji instance with tag2distrepo set up so we can just tag older packages we need in there.)


@jlebon the current version of this PR 'reverts' the SELinux policies to the previous ones for the `afterburn_t` domain (permissive). If it's invasive because we allow a lot more than what's needed, I can go back to the fine-grained policies we proposed earlier.

I also understand that maintaining SELinux policies is not the most exciting thing here. Still, it allows the SCOS release to unblock and OKD/SCOS to have dedicated SELinux policies when needed, especially since, based on what you said, the SELinux policies in Fedora C9S might not always be in sync (and well-tested as they are for RHCOS). 

To have working and quickly released OKD/SCOS, we will occasionally need more workarounds dedicated to C9S than RHCOS.

This PR adds that systemd unit only to C9S builds, with the hope that whoever maintains and reviews a change in that folder will always bind a Jira or GH issue to revert the changes eventually when they are fixed and downstreamed to be consumed as a permanent solution.

We had similar cases for OKD/FCOS: https://github.com/openshift/okd-machine-os/blob/master/overlay.d/99okd/usr/lib/okd/selinux-fixes.cil. 

I would avoid further delays for the release of OKD/SCOS, 4.16 and beyond when similar cases arise.

Prashanth684:
@travier @jlebon any suggestions on how this can be unblocked in the short term? I think @aleskandro has given sufficient reasoning above for the approach.

ausil:
From my point of view this approach is better than setting up a repo and locking to an old version, we will be missing all of the bug fixes in the newer selinux-policy builds.  This is not an issue in the FCOS portion because the bug is fixed upstream. We are missing all the other fixes in selinux-policy if we lock to an old version. The selinux team needs to prioritise backporting the upstream fix. As I understand the issue, we only hit it on reboot when moving to CentOS Stream, which is where the fix is needed.

ausil:
@dustymabe @cverna This is blocking OKD releases from going stable cc @preethit

jlebon:
How about having just the `overlay.d/50scos/usr/lib/okd/selinux/afterburn-permissive.cil` file, and a postprocess step to `semodule -i` it? I think it'd be a lot cleaner to have this be part of the baked policy instead of forcing a local policy rebuild.

aleskandro:
> How about having just the `overlay.d/50scos/usr/lib/okd/selinux/afterburn-permissive.cil` file, and a postprocess step to `semodule -i` it? I think it'd be a lot cleaner to have this be part of the baked policy instead of forcing a local policy rebuild.

Hi @jlebon, that is better. Done

Thanks for following up.

jlebon:
Thanks! I've reworked this as follow:
1. There's actually no need to ship this in `/usr`. Doing `semodule -i` will automatically compile and copy the results into `/etc/selinux/targeted/active/` along with the other packages, so it'd just be duplicating it. Instead, we now just create a temporary file to be able to install it, and then delete the file.
2. I've added additional workarounds for other outstanding SELinux issues.
3. I've moved the postprocess script to `manifest-c9s.yaml` since this affects both the `c9s` and `okd-c9s` variants.

jlebon:
I've also inlined the first commit of https://github.com/openshift/os/pull/1544 here. This PR _should_ now be all we need to unblock building c9s-based bootimages in the CoreOS pipeline.

aleskandro:
Looks good to me!

Thanks @jlebon

PR #1562, opened 2024-08-08
COS-2822: port gcp-routes to nftables

danwinship:
iptables is going away in RHEL10; this PR ports gcp-routes.sh from iptables to nftables.

This is just a port of https://github.com/openshift/machine-config-operator/pull/4494 which updated MCO's copy of gcp-routes... the version in os is slightly simpler (because it's older and it only needs to support bootstrap and so hasn't needed the fixes that are in the MCO version). (Or, honestly, probably "it has the same problems the MCO version used to have, but our CI doesn't flag the apiserver connection failures-and-retries during bootstrap switchover like it does during upgrade, so nobody ever bothered to fix the bugs in the RHCOS version.) Anyway, see discussion in that PR.

I tried to make a jira issue for this but jira insists (to me anyway) that there is no such project as "COS" and will not let me create a new issue in it. (But if someone else wants to create one and link it to OCPSTRAT-940, go ahead.)

### DIFF
--- a/overlay.d/06gcp-routes/usr/sbin/gcp-routes.sh
+++ b/overlay.d/06gcp-routes/usr/sbin/gcp-routes.sh
@@ -55,10 +55,6 @@ initialize() {
     ensure_chain nat "${CHAIN_NAME}"
     ensure_rule nat PREROUTING -m comment --comment 'gcp LB vip DNAT' -j ${CHAIN_NAME}
 
-    # Need this so that existing flows (with an entry in conntrack) continue to be
-    # balanced, even if the DNAT entry is removed
-    ensure_rule filter INPUT -m comment --comment 'gcp LB vip existing' -m addrtype ! --dst-type LOCAL -m state --state ESTABLISHED,RELATED -j ACCEPT
-
     mkdir -p "${RUN_DIR}"
 }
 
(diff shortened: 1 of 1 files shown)


danwinship:
/assign @jcaamano 
(There is a separate copy of MCO's gcp-routes in RHCOS itself, which is only used on the bootstrap host.)

jlebon (commented):
I've moved https://issues.redhat.com/browse/MCO-1266 to https://issues.redhat.com/browse/COS-2882. Hopefully that's what you meant. If so, you can use that for the Jira link. Also, can you add a commit message to the second commit and link to the OCPSTRAT?

Trivial LGTM for the change itself. The script lives here but we're not really the SMEs for it.

danwinship:
> I've moved https://issues.redhat.com/browse/MCO-1266 to https://issues.redhat.com/browse/COS-2882. Hopefully that's what you meant.

That was my first attempt, yes, though it seems like "Clone" in jira is generally terrible anyway, because that card ended up with a bunch of MCO-tagged qe-related sub-tasks whereas presumably it should have ended up with COS ones? :man_shrugging:

danwinship:
@jlebon repushed with updated commit message

jlebon (approved):
/approve
/lgtm

danwinship:
@jlebon try again? bot didn't seem to notice...

jlebon:
Quay.io infra flake
/retest

PR #1563, opened 2024-08-08
OCPBUGS-35247: common.yaml: add subscription-manager

jlebon:
We currently ship dnf in RHCOS but it's cumbersome to use because the
librhsm magic we had for rpm-ostree to make content available during
builds on a subscribed node doesn't work with dnf. For that, it seems
like we need full-blown subscription-manager in the container image.

Add it.

This matches centos-bootc/rhel-bootc shipping subman as well, though I'd
like in the future to investigate lowering the dependency requirements
to just a dnf plugin + librhsm.

Fixes: https://issues.redhat.com/browse/OCPBUGS-35247

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -189,8 +189,6 @@ check-groups:
   filename: "group"
 
 exclude-packages:
-  # https://bugzilla.redhat.com/show_bug.cgi?id=1798278
-  - subscription-manager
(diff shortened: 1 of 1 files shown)


jlebon:
Huh OK, I didn't realize subscription-manager was a consolehelper program. So this is triggering the setuid tests on the classic `userhelper`.

OK opened https://github.com/coreos/fedora-coreos-config/pull/3147.

aaradhak (approved):

marmijo (approved):
/lgtm

Issue #1564, opened 2024-08-08
SCOS builds in CI failing with "Error: There are no enabled repositories"


jlebon:
E.g. see in https://github.com/openshift/os/pull/1539.

```
 [1/3] STEP 8/15: RUN if rpm -q centos-stream-release; then dnf install -y centos-release-{cloud,nfv,virt}-common; fi
centos-stream-release-9.0-26.el9.noarch
Error: There are no enabled repositories in "/etc/yum.repos.d", "/etc/yum/repos.d", "/etc/distro.repos.d".
Error: building at STEP "RUN if rpm -q centos-stream-release; then dnf install -y centos-release-{cloud,nfv,virt}-common; fi": while running runtime: exit status 1
error: calling build-extensions-container.sh: exit status 1 
```

jlebon:
This is a regression from https://github.com/openshift/os/pull/1561.

PR #1565, opened 2024-08-08
NO-JIRA: extensions/Dockerfile: fix for okd-c9s variant

jlebon:
In fc9e703 ("extensions/Dockerfile: install SIG GPG keys on CentOS"), we
fixed the extensions container build for the `c9s` variant, but that
broke it for the `okd-c9s` variant, which CI currently still tests.

We should switch CI over to target the `c9s` variant and get rid of
`okd-c9s`, but for now at least let's keep it happy.

The problem is that in the `okd-c9s` variant, we disable all the repos
as part of `packages-openshift.yaml`. But anyway, we also already
include the various release packages containing the GPG keys there, so
we can skip this step entirely.

Fixes fc9e703 ("extensions/Dockerfile: install SIG GPG keys on CentOS").
Fixes: #1564

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -10,7 +10,7 @@ ARG COSA
 ARG VARIANT
 RUN if [[ -z "$COSA" ]] ; then ci/get-ocp-repo.sh ; fi
 # on SCOS, we need to add the GPG keys of the various SIGs we need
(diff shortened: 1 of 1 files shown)


marmijo (approved):
/lgtm

gursewak1997 (approved):

PR #1568, opened 2024-08-14
NO-JIRA: Revert "c9s: work around various SELinux issues"

jlebon:
This reverts commit 9ba8b7129f086c7618fcd3136d738cdae1df57b7.

All the issues here should be fixed by selinux-policy-38.1.44-1.el9 according to the linked tickets. So we shouldn't need this anymore.

### DIFF
--- a/manifest-c9s.yaml
+++ b/manifest-c9s.yaml
@@ -26,23 +26,3 @@ mutate-os-release: "9"
 
 packages:
  - centos-stream-release
-
-postprocess:
-  - |
-    #!/usr/bin/env bash
-    set -xeuo pipefail
-    cat > /tmp/scos-workarounds.cil << EOF
-    ; https://issues.redhat.com/browse/RHEL-49735
-    (typeattributeset cil_gen_require afterburn_t)
-    (typepermissive afterburn_t)
-
-    ; https://issues.redhat.com/browse/RHEL-38614
-    (typeattributeset cil_gen_require coreos_installer_t)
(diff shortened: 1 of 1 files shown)


jlebon:
Quay.io flakes

/retest

aleskandro:
Quay is having issues in the past few hours

jbtrystram (approved):
/lgtm 
/approve

travier:
From https://github.com/openshift/os/pull/1556, https://issues.redhat.com/browse/RHEL-38614 is still in progress.

jlebon:
> From #1556, [issues.redhat.com/browse/RHEL-38614](https://issues.redhat.com/browse/RHEL-38614) is still in progress.

I think the status there is misleading. I think it's Release Pending because it's attached to a RHEL 9.5 errata, but it was indeed fixed in c9s (since the testiso tests pass here). So I think we can get this in?

/retest

dustymabe:
@jbtrystram - can you try to replicate the CI failures by doing a local build and running tests against SCOS? There are a lot of failures so I'd imagine we should be able to reproduce locally.

jbtrystram:
> @jbtrystram - can you try to replicate the CI failures by doing a local build and running tests against SCOS? There are a lot of failures so I'd imagine we should be able to reproduce locally.

This is hitting https://issues.redhat.com/browse/RHEL-61612
Maybe fixed by https://github.com/openshift/os/pull/1631

Issue #1546, opened 2024-07-15
Stop using EOL Fedora container images in older releases


jlebon:
Some of our tests shared with FCOS build containers locally. Once we branch f-c-c for RHCOS, those container references are never updated. Once those Fedora releases go EOL, their content move. And while it should keep working, hiccups do happen in that transition (see e.g. https://pagure.io/fedora-infrastructure/issue/11221 and https://pagure.io/releng/issue/12204).

But anyway, it's also not great to be relying on EOL content in our tests.

My proposal is to modify those tests to use UBI images on RHCOS, c9s images on SCOS, and Fedora images on FCOS.

dustymabe:
Does this problem change much in the new world where we are building a RHEL only base layer?

PR #1583, opened 2024-08-22
NO-JIRA: extensions/Dockerfile: use latest Fedora to generate repo

jlebon:
In this stage of the build, we're just using Fedora to generate repodata. Since this is a pretty basic operation, let's just always use the latest Fedora instead of hardcoding a specific version. This ensures that e.g. we're never going to use an EOL Fedora release.

This became an issue after 8301c67 ("extensions/Dockerfile: use the FCOS defined fedora.repo to set up container"), which was backported to older RHCOS branches that used EOL Fedora releases. Before, we were using the mirrors, which papered over the fact that the content for EOL releases actually moved to another location. But now that we're using the canonical Fedora server, we're directly exposed to that.

Really, we should just stop hardcoding Fedora versions in this repo. We shouldn't be using EOL releases at all as part of our builds or tests. (Obviously a massive offender here is cosa... though that's a discussion of its own.)

See also: https://github.com/openshift/os/issues/1546

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -22,7 +22,7 @@ RUN if [[ -n "${VARIANT}" ]]; then MANIFEST="manifest-${VARIANT}.yaml"; EXTENSIO
 ## Creates the repo metadata for the extensions.
 ## This uses Fedora as a lowest-common-denominator because it will work on
 ## current p8/s390x. See https://github.com/openshift/os/issues/1000
-FROM quay.io/fedora/fedora:40 as builder
+FROM quay.io/fedora/fedora:latest as builder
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

marmijo:
/retitle NO-JIRA: extensions/Dockerfile: use latest Fedora to generate repo

marmijo:
backport PRs: 
- https://github.com/openshift/os/pull/1592
- https://github.com/openshift/os/pull/1591
- https://github.com/openshift/os/pull/1590
- https://github.com/openshift/os/pull/1589
- https://github.com/openshift/os/pull/1588
- https://github.com/openshift/os/pull/1587

PR #1594, opened 2024-08-22
extensions/Dockerfile - Use fedora:latest

aaradhak:
The fedora modular repo is not used anymore and the repo download link seems broken after switching to fedora:40.

Hence, we are changing fedora version to be latest.


marmijo:
I don't think this one is needed either. Looks like `release-4.19` and `master` are in sync with each other.

PR #1600, opened 2024-09-04
OCPBUGS-39536: Only run rhcos-fips services if fips=1

derekhiggins:
The existence of the fips kernel param doesn't imply fips is enabled. It needs to be set to a '1' for the kenel to enable fips
https://github.com/torvalds/linux/blob/88fac17500f4ea49c7bac136cf1b27e7b9980075/crypto/fips.c#L23


derekhiggins:
I think the kernel may also enable fips if somebody sets fips=\<any_non_zero_int>
So If we don't want trigger the systems unit based "fips=1"  
we could instead exit overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fips/rhcos-fips.sh early if "fips=0" is in the kernel params

jlebon:
Thanks! This makes sense to me.

> I think the kernel may also enable fips if somebody sets fips=<any_non_zero_int>

I'm not too worried about this. Apart from the fact that anything other than 0/1 is not canonical, the `fips` karg in the common flow is actually set by ourselves when we detect a MachineConfig that flips that bool.

/jira refresh
/approve
/lgtm

PR #1603, opened 2024-09-04
NO-JIRA: extensions/Dockerfile: use fedora-archive.repo and revert "use fedora:latest"

marmijo:
A `fedora-archive.repo` file will now be used by older branches using EOL fedora container versions[1]. This now acts as our mechanism for handling the situation where the EOL Fedora content actually moves to a different location by pointing to the archived content directly. The `fedora-archive.repo` file in the testing-devel branch will be curled during the container setup instead of the `fedora.repo` file when the version goes EOL.

This does mean we'll still have to maintain this container setup by updating the curl command as the versions go EOL around each new release.

This reverts commit 94dfed82c7882276e4e7d07d94aa519a9a037988.

EDIT: 

We can actually now use the `fedora-archive.repo` for both EOL and non-EOL fedora releases. We decided to use it for the extensions container so we dont have to swap out `fedora.repo` when versions go EOL. Make that change here too.


marmijo:
/retitle NO-JIRA: extensions/Dockerfile: use fedora-archive.repo and revert "use fedora:latest"

PR #1604, opened 2024-09-04
[release-4.17] NO-JIRA: bump fedora-coreos-config and update extensions container

marmijo:

EDIT:

We can actually now use the fedora-archive.repo for both EOL and non-EOL fedora releases. We decided to use it for the extensions container so we dont have to swap out fedora.repo when versions go EOL. Make that change here too.

This reverts half of commit 2477d030b80690ed1299a50cc6b82469cb3cc3ea.
and backports: https://github.com/openshift/os/pull/1603

---

This also includes a submodule bump for `[release-4.17]`

```
Michael Armijo (1):
      tests/kola: use fedora-archive.repo when setting up fedora container
```


marmijo:
/hold
wait to include the submodule bump for https://github.com/coreos/fedora-coreos-config/pull/3148 here too

marmijo:
/retitle [release-4.17] NO-JIRA: bump fedora-coreos-config and update extensions container

PR #1608, opened 2024-09-05
[release-4.13] NO-JIRA: bump fedora-coreos-config and update extensions container

marmijo:
#### bump fedora-coreos-config
The submodule bump is required for the RHCOS pipeline migration to the ITUP cluster.
```
Michael Armijo (1):
      tests/kola: use fedora-archive.repo for EOL Fedora containers
```

xref: https://github.com/coreos/fedora-coreos-config/pull/3137

---
#### extensions/Dockerfile: use fedora:37 and fedora-archive.repo

This essentially reverts these two commits:
- https://github.com/openshift/os/commit/b56bcd8e00e737213a965d3300ce18d7c068977d
- https://github.com/openshift/os/commit/7487bccb911f19971a7fb8fdaacb82ea36a7e430


aaradhak (approved):

c4rt0 (approved):
/lgtm

PR #1609, opened 2024-09-05
[release-4.12] NO-JIRA: bump fedora-coreos-config and update extensions container

marmijo:
#### bump fedora-coreos-config
The submodule bump is required for the RHCOS pipeline migration to the ITUP cluster.
```
Michael Armijo (1):
      tests/kola: use fedora-archive.repo for EOL Fedora containers
```
xref: https://github.com/coreos/fedora-coreos-config/pull/3138

---
#### extensions/Dockerfile: use fedora:37 and fedora-archive.repo

This essentially reverts these two commits:
- https://github.com/openshift/os/commit/acc278a0b4e855b43779f3748ab2d25e83418680
- https://github.com/openshift/os/commit/3e11f2866a383daef4f84d3621315576d486db04


aaradhak (approved):

c4rt0 (approved):
/lgtm

PR #1550, opened 2024-07-15
[release-4.14] NO-JIRA: denylist: add dhcp-propagation and podman.rootless-systemd

marmijo:
In 4.14, these tests are using the F38 container. The F38 content was moved to the archive on the mirrors but the metalinks still reference the non-archive URLs causing these tests to fail.

Let's denylist them for now until https://pagure.io/releng/issue/12204 is resolved.

PR #1614, opened 2024-09-09
[release-4.14] NO-JIRA: Revert "denylist: add dhcp-propagation and podman.rootless-systemd"

marmijo:
The Fedora 38 content is available using the metalinks again and these two kola tests are passing again in 4.14.

ref: https://pagure.io/releng/issue/12204

This reverts commit 65d0b71dfc75dc37eef1cfac504b6c52cdefc63e.

see: https://github.com/openshift/os/pull/1550

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -28,9 +28,3 @@
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
   - c9s
-
-- pattern: ext.config.shared.podman.rootless-systemd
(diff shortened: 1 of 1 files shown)


marmijo:
Actually, we've specified the download location for EOL content in `fedora-archive.repo` which we're using in these tests, so the content is definitely available. https://github.com/openshift/os/pull/1607

dustymabe (approved):
/lgtm

PR #1615, opened 2024-09-09
rhel-9.4/c9s: make versioning scheme semver compatible

jlebon:
There's a bunch of things out there that currently parse the RHCOS version string. The semantics of this string will change as part of this change, but we should at least try to retain the same semver-compatible format.

E.g. currently, an RHCOS version string of the `ocp-rhel-9.4` variant is 418.94.202409090849-0. With this PR, an RHCOS version string of the `rhel-9.4` variant is 9.4.202409090849-0 (specifically, this PR changes the `.0` for `-0`).


dustymabe:
> specifically, this PR changes the `.0` for `-0`

given the preceding sentence I'm not sure I understand this statement fully

jbtrystram (approved):
/lgtm
/approve

c4rt0 (approved):
I had to do a little digging to find [automatic-version-suffix](https://coreos.github.io/rpm-ostree/treefile/).
/lgtm

jlebon:
> > specifically, this PR changes the `.0` for `-0`
> 
> given the preceding sentence I'm not sure I understand this statement fully

Updated the git message to make this clearer!

PR #1623, opened 2024-09-26
NO-JIRA: overlay.d/05rhcos: remove rhcos-usrlocal-selinux-fixup.service

dustymabe:
The policy seems to be correct now for /usr/local/sbin.

```
[core@localhost ~]$ rpm -q selinux-policy
selinux-policy-38.1.35-2.el9_4.2.noarch
[core@localhost ~]$ matchpathcon /usr/local/sbin /usr/sbin
/usr/local/sbin system_u:object_r:bin_t:s0
/usr/sbin       system_u:object_r:bin_t:s0
```

Context in https://bugzilla.redhat.com/show_bug.cgi?id=1956836

-----

Also, a second commit to drop a preset from a service we no longer ship.


gursewak1997 (approved):
lgtm

PR #1629, opened 2024-10-02
COS-2959: Add rhel-9.6 and ocp-rhel-9.6 variants

jbtrystram:
Right now these are pure centOS Stream 9 builds with a spoofed `os-release` file.

For the ocp variant, pull the packages from the 9.4-4.18 repos.

https://issues.redhat.com/browse/COS-2959

### DIFF
--- a/ci/prow-entrypoint.sh
+++ b/ci/prow-entrypoint.sh
@@ -73,6 +73,12 @@ prepare_repos() {
     # Figure out which version we're building
     rhelver=$(rpm-ostree compose tree --print-only "${manifest}" | jq -r '.["automatic-version-prefix"]' | cut -f2 -d.)
 
+    # XXX change to rhel 9.6 when beta is GA
+    # use 9.4 repos for 9.6
+    if [[ "${rhelver}" == "96" ]]; then
+        rhelver="94"
+    fi
+
     # Temporary workaround until we publish builds in the default path
     if [[ "${rhelver}" == "94" ]]; then
         prev_build_url="${REDIRECTOR_URL}/${ocpver}-9.4/builds/"
--- a/extensions-ocp-rhel-9.6.yaml
+++ b/extensions-ocp-rhel-9.6.yaml
@@ -1,0 +1,1 @@
+extensions-rhel-9.6.yaml
--- a/extensions-rhel-9.6.yaml
+++ b/extensions-rhel-9.6.yaml
@@ -1,0 +1,94 @@
+# RPMs as operating system extensions, distinct from the base ostree commit/image
+# https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md
+# and https://github.com/coreos/fedora-coreos-tracker/issues/401
+
+extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  wasm:
+    architectures:
+      - x86_64
+      - aarch64
+    repos:
+      # XXX todo: swap to rhel 9.6 repos when beta is GA
+      # Ideally we would use content from c9s as it's the same as rhel 9.6 for now
+      # but this particular package does not exist there
+      - rhel-9.4-server-ose-4.18
+    packages:
+      - crun-wasm
+  # https://github.com/coreos/fedora-coreos-tracker/issues/1504
+  ipsec:
+    packages:
+      - libreswan
+      - NetworkManager-libreswan
+  # https://github.com/coreos/fedora-coreos-tracker/issues/326
+  usbguard:
+    packages:
+      - usbguard
+  kerberos:
+    packages:
+      - krb5-workstation
+      - libkadm5
+  # https://github.com/kmods-via-containers/kmods-via-containers/issues/3
+  # https://gitlab.cee.redhat.com/coreos/redhat-coreos/merge_requests/866
+  # These are currently overlaid onto the host so that they can be bind-mounted
+  # into build containers... in the future they should be a `development`
+  # extension: https://github.com/openshift/machine-config-operator/pull/2143.
+  kernel-devel:
+    packages:
+      - kernel-devel
+      - kernel-headers
+    match-base-evr: kernel
+  # These are already in the base, so they're not OS extensions, but they're
+  # useful to have in RPM form to install in kmod build containers.
+  kernel:
+    kind: development
+    packages:
+      - kernel
+      - kernel-core
+      - kernel-modules
+      - kernel-modules-extra
+    match-base-evr: kernel
+  # GRPA-2822
+  # https://github.com/openshift/machine-config-operator/pull/1330
+  # https://github.com/openshift/enhancements/blob/master/enhancements/support-for-realtime-kernel.md
+  kernel-rt:
+    architectures:
+      - x86_64
+    repos:
+      # XXX todo: swap to rhel 9.6 repos when beta is GA
+      - c9s-nfv
+    packages:
+      - kernel-rt-core
+      - kernel-rt-kvm
+      - kernel-rt-modules
+      - kernel-rt-modules-extra
+      - kernel-rt-devel
+    match-base-evr: kernel
+  # https://github.com/openshift/machine-config-operator/pull/2456
+  # https://github.com/openshift/enhancements/blob/master/enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
+  # GRPA-3123
+  # - kata-containers
+  sandboxed-containers:
+    architectures:
+      - x86_64
+      - s390x
+    repos:
+      # XXX ideally we would pull that from c9s as it's identical to rhel 9.6 for now
+      # but c9s does not build this for s390x so we have to pull it from rhel 9.4
+      # todo: swap to rhel 9.6 repos when beta is GA
+      - rhel-9.4-server-ose-4.18
+    packages:
+      - kata-containers
+  # https://issues.redhat.com/browse/COS-2402
+  kernel-64k:
+    architectures:
+      - aarch64
+    packages:
+      - kernel-64k-core
+      - kernel-64k-modules
+      - kernel-64k-modules-core
+      - kernel-64k-modules-extra
+  # https://issues.redhat.com/browse/COS-2940
+  sysstat:
+    packages:
+      - sysstat
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -7,21 +7,25 @@
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
     - c9s
+    - rhel-9.6
 
 - pattern: iso-live-login.uefi-secure
   tracker: https://github.com/openshift/os/issues/1237
   osversion:
     - c9s
+    - rhel-9.6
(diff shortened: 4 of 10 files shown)


ci/prow-entrypoint.sh
```diff
@@ -316,13 +321,13 @@ main() {
             ;;
         "rhcos-9-build-test-qemu")
             setup_user
-            cosa_init "ocp-rhel-9.4"
+            cosa_init "ocp-rhel-9.6"
```
jlebon, line 324: We can't change our primary CI over to 9.6 yet. The master branch currently tracks 4.18, which will ship on 9.4. I guess we could add another CI run for it for now, though it's going to be mostly covered by SCOS anyway, at least until branching on the RHEL side.

manifest-rhel-9.6.yaml
```diff
@@ -0,0 +1,88 @@
...
+      ; https://issues.redhat.com/browse/RHEL-38614
+      (typeattributeset cil_gen_require coreos_installer_t)
+      (typepermissive coreos_installer_t)
+
+      ; https://issues.redhat.com/browse/RHEL-47033
+      (typeattributeset cil_gen_require systemd_network_generator_t)
+      (typepermissive systemd_network_generator_t)
+      EOF
+      /usr/sbin/semodule -i /tmp/scos-workarounds.cil
+      rm /tmp/scos-workarounds.cil
```
jlebon, line 88: Let's see if we can merge https://github.com/openshift/os/pull/1568 so we can drop this.

manifest-rhel-9.6.yaml
```diff
@@ -0,0 +1,88 @@
...
+     CPE_NAME="cpe:/o:redhat:enterprise_linux:9::coreos"
+     HOME_URL="https://www.redhat.com/"
+     DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
+     BUG_REPORT_URL="https://issues.redhat.com/"
+
+     REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
+     REDHAT_BUGZILLA_PRODUCT_VERSION=9.6
+     REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
+     REDHAT_SUPPORT_PRODUCT_VERSION="9.6"
+     EOF
```
jlebon, line 63: Can you show the resulting `/etc/os-release` file in the built container image/VM? Compare to https://gist.github.com/jlebon/908e45c42ce9e365fb83d7470cbbb40c for how it currently looks for `rhel-9.4`.

jbtrystram: I just update to have something that matches : 
```
NAME="Red Hat Enterprise Linux CoreOS"
VERSION="9.6.202410031540-0 (Plow)"
ID="rhel"
ID_LIKE="fedora"
VERSION="9.6.202410031540-0"
VARIANT="CoreOS"
VARIANT_ID=coreos
VERSION_ID="9.6"
PLATFORM_ID="platform:el9"
PRETTY_NAME="Red Hat Enterprise Linux CoreOS 9.6.202410031540-0 (Plow)"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:redhat:enterprise_linux:9::baseos"
HOME_URL="https://www.redhat.com/"
DOCUMENTATION_URL="https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9"
BUG_REPORT_URL="https://issues.redhat.com/"

REDHAT_BUGZILLA_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_BUGZILLA_PRODUCT_VERSION=9.6
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux"
REDHAT_SUPPORT_PRODUCT_VERSION="9.6"
OSTREE_VERSION="9.6.202410031540-0"
```

dustymabe:
is the intention to squash these commits when ready to merge?

kola-denylist.yaml
```diff
@@ -31,3 +35,13 @@
...
+
+# as it's a fake rhel build (from c9s) versions won't match
+- pattern: ext.config.version.rhel-major-version
+  tracker: https://github.com/openshift/os/issues/1635
+  osversion:
+    - rhel-9.6
+- pattern: ext.config.shared.content-origins
+  tracker: https://github.com/openshift/os/issues/1635
+  osversion:
+    - rhel-9.6
```
dustymabe, line 47: can we add a snooze to these so that we don't forget to start running the test?

I wish we could do it for all of the UEFI ones above but those have `c9s` on them too.

extensions-ocp-rhel-9.6.yaml
```diff
@@ -0,0 +1 @@
+extensions-ocp-rhel-9.4.yaml
```
dustymabe, line 1: if this is going to be a symlink to `extensions-ocp-rhel-9.4.yaml` (which is itself a symlink to `extensions-rhel-9.4.yaml`) then why create `extensions-rhel-9.6.yaml` at all?

extensions-rhel-9.6.yaml
```diff
@@ -0,0 +1,90 @@
+# RPMs as operating system extensions, distinct from the base ostree commit/image
+# https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/extensions.md
+# and https://github.com/coreos/fedora-coreos-tracker/issues/401
+
+extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  # Not available in c9s
```
dustymabe, line 7: I'm not sure why this `Not available in c9s` comment is here. Can you add more context?

extensions-rhel-9.6.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - kernel-modules-extra
+    match-base-evr: kernel
+  # GRPA-2822
+  # https://github.com/openshift/machine-config-operator/pull/1330
+  # https://github.com/openshift/enhancements/blob/master/enhancements/support-for-realtime-kernel.md
+  kernel-rt:
+    architectures:
+      - x86_64
+    repos:
+      - c9s-nfv
```
dustymabe, line 57: I imagine this is something we'll want to change back later? Maybe we should add an `# XXX` comment above everything that should change once we are no longer "fake" 9.6.

extensions-rhel-9.6.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - kernel-rt-kvm
+      - kernel-rt-modules
+      - kernel-rt-modules-extra
+      - kernel-rt-devel
+    match-base-evr: kernel
+  # https://github.com/openshift/machine-config-operator/pull/2456
+  # https://github.com/openshift/enhancements/blob/master/enhancements/sandboxed-containers/sandboxed-containers-tech-preview.md
+  # GRPA-3123
+  # - kata-containers
+  # s390x builds not available in c9s
```
dustymabe, line 69:
```
  # s390x builds not available in c9s
```

but what change did you make because of that fact? the following section matches exactly `extensions-rhel-9.4.yaml`

extensions-rhel-9.6.yaml
```diff
@@ -0,0 +1,90 @@
...
+extensions:
+  # https://issues.redhat.com/browse/RFE-4177
+  # Not available in c9s
+  wasm:
+    architectures:
+      - x86_64
+      - aarch64
+    repos:
+      - rhel-9.4-server-ose-4.18
+      - rhel-9.4-appstream
```
dustymabe, line 14: Why did we add `rhel-9.4-appstream` here when compared to the same section in `extensions-rhel-9.4.yaml` ?

manifest-rhel-9.6.yaml
```diff
@@ -0,0 +1,74 @@
...
+# This ensures we're semver-compatible which OpenShift wants
+automatic-version-suffix: "-"
+
+mutate-os-release: "9.6"
+
+repo-packages:
+  - repo: c9s-baseos
+    packages:
+ # We include the generic centos release package and fake the red hat os-release
+ # info in a post-process script
```
dustymabe, line 33: add indentation to this comment?

manifest-rhel-9.6.yaml
```diff
@@ -0,0 +1,74 @@
...
+     - centos-stream-release
+
+# Fake out RHEL version in the os-release while waiting for RHEL-9.6 release.
```
dustymabe, line 36: should we just be find/replace `9.4` with `9.6` in the `os-release` here? i.e. in `manifest-rhel-9.4.yaml` there is no postprocess of `os-release` that is done now (since 4e09d27) so we shouldn't need to replace the whole thing?

extensions-rhel-9.6.yaml
jbtrystram: The `crun-wasm` package is not available in c9s repos. It's needed for OCP extensions so we pull it from RHEL 9.4 repos

extensions-rhel-9.6.yaml
jbtrystram: Good idea, will update

manifest-rhel-9.6.yaml
jbtrystram: > should we just be find/replace 9.4 with 9.6 in the os-release here?

Doing that would require installing `redhat-release` from the rhel 9.4 repos. That release file comes from `centos-strea-release`. I tried to do that, as that would make things simpler here but then conflicts with `centos-stream-release` which is required to get the RPM repository files and GPG keys and causes issues when trying to build the extensions container (which needs to downloads RPMs)

extensions-rhel-9.6.yaml
jbtrystram: I just removed the comment and added another one below with more clarification

jbtrystram:
> is the intention to squash these commits when ready to merge?

I was thinking of making several commits so the denylists changes would be easier to revert but since i'll have to do changes in the manifest anyway i'll do it in one go I suppose

dustymabe (approved):
/lgtm

ci/prow-entrypoint.sh
```diff
@@ -305,7 +311,7 @@ main() {
             prepare_repos
             ;;
         "build" | "init-and-build-default")  # TODO: change prow job to use init-and-build-default
-            cosa_init "ocp-rhel-9.4"
+            cosa_init "ocp-rhel-9.6"
```
travier, line 314: Not yet

manifest-ocp-rhel-9.6.yaml
```diff
@@ -0,0 +1,111 @@
+# Manifest for OCP node based on RHEL 9.4
```
travier, line 1:
```suggestion
# Manifest for OCP node based on RHEL 9.6
```

travier (changes requested):
One nit and one fix needed

ci/prow-entrypoint.sh
jbtrystram: good catch ! Thanks

Issue #1640, opened 2024-10-21
[rhel-9.4 variant] ext.config.rpm-ostree.replace-rt-kernel fails


c4rt0:
This failed in [4.16](https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com/job/build/10188/), [4.17](https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com/job/build/10186/) and [4.18](https://jenkins-rhcos.apps.ocp-virt.prod.psi.redhat.com/job/build/10187/) pipeline builds. 

```
[2024-10-21T15:10:48.484Z] Oct 21 15:10:44 qemu0 systemd[1]: Started kola-runext.service.
[2024-10-21T15:10:48.484Z] Oct 21 15:10:44 qemu0 kola-runext-test.sh[2118]: + rm -rf '/etc/yum.repos.d/*'
[2024-10-21T15:10:48.484Z] Oct 21 15:10:44 qemu0 kola-runext-test.sh[2120]: + cp /var/opt/kola/extdata/c9s.repo /etc/yum.repos.d/cs.repo
[2024-10-21T15:10:48.484Z] Oct 21 15:10:44 qemu0 kola-runext-test.sh[2122]: + curl -sSLf https://centos.org/keys/RPM-GPG-KEY-CentOS-Official -o /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-Official
[2024-10-21T15:10:48.484Z] Oct 21 15:10:45 qemu0 kola-runext-test.sh[2125]: package centos-stream-release is not installed
[2024-10-21T15:10:48.484Z] Oct 21 15:10:45 qemu0 kola-runext-test.sh[2112]: Testing overriding with CentOS Stream kernel
[2024-10-21T15:10:48.484Z] Oct 21 15:10:45 qemu0 kola-runext-test.sh[2126]: + sed -i s/enabled=1/enabled=0/g /etc/yum.repos.d/cs.repo
[2024-10-21T15:10:48.484Z] Oct 21 15:10:45 qemu0 kola-runext-test.sh[2128]: + sed -i '/\[baseos\]/,/^ *\[/ s/enabled=0/enabled=1/' /etc/yum.repos.d/cs.repo
[...]

PR #1641, opened 2024-10-21
denylist: add ext.config.rpm-ostree.replace-rt-kernel

c4rt0:
In 4.16, 4.17 and 4.18 [rhel-9.4] this test is failing after kernel being upgraded in the Build OSTree stage.

See: https://github.com/openshift/os/issues/1640

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -48,3 +48,8 @@
   snooze: 2025-01-01
   osversion:
     - rhel-9.6
+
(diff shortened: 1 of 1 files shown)


dustymabe:
/hold

I don't think this is needed any longer.

c4rt0:
This was happening due to an issue with the yum repos, closing.
https://github.com/openshift/os/issues/1640#issuecomment-2428589180

PR #1648, opened 2024-11-13
NO-JIRA: [ocp-]rhel-9.6: consume from RHEL 9.6 repos

jlebon:
RHEL 9.6 nightly content is already available to use so let's start tracking that instead for the 9.6 variants.

Note that OCP is approved to use that content.

While we're here, enact https://issues.redhat.com/browse/RHELPLAN-170883 where there was agreement that all container-related RPMs starting from 4.19+ will come from RHEL. This is prep for
https://github.com/openshift/enhancements/pull/1637.

### DIFF
--- a/manifest-ocp-rhel-9.6.yaml
+++ b/manifest-ocp-rhel-9.6.yaml
@@ -16,20 +16,9 @@ include:
   - packages-openshift.yaml
 
 # Additional repos we need for OCP components
-# right now we use rhel-9.4 OCP repos as RHEL 9.6 is not available yet
 repos:
-  - rhel-9.4-fast-datapath
-  - rhel-9.4-server-ose-4.18
-  - rhel-9.4-appstream
-  - c9s-extras-common
-  - c9s-sig-nfv
-  - c9s-nfv
-
-packages:
- # RPM GPG keys for CentOS SIG repos
-  - centos-release-cloud-common
-  - centos-release-nfv-common
-  - centos-release-virt-common
+  - rhel-9.6-fast-datapath
+  - rhel-9.6-server-ose-4.18
 
 # We include hours/minutes to avoid version number reuse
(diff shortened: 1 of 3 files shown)


manifest-ocp-rhel-9.6.yaml
```diff
@@ -1,4 +1,4 @@
-# Manifest for OCP node based on RHEL 9.6
+# Manifest for OCP node based on RHEL 9.4
```
travier, line 1:
```suggestion
# Manifest for OCP node based on RHEL 9.6
```

manifest-rhel-9.6.yaml
```diff
@@ -1,4 +1,4 @@
-# Manifest for RHCOS based on RHEL 9.6
+# Manifest for RHCOS based on RHEL 9.4
```
travier, line 1:
```suggestion
# Manifest for RHCOS based on RHEL 9.6
```

jbtrystram (approved):
/lgtm
/approve

jlebon:
We're still waiting on final confirmation on this.

/hold

jlebon:
We got confirmation on this now. For posterity, look at internal comments in https://url.corp.redhat.com/31bad51 and links there.

/hold cancel

Note that there is no 9.6 content yet on the mirrors. So for now we'll only be able to test this in earnest in the pipeline. I'll request mirrors to be added but since the content isn't actually on the CDN I'm not sure if it's possible.

jlebon:
This one just needs a `/lgtm` now I think.

mike-nguyen:
/lgtm

I'll have a follow-on version bump PR that I'll need to rebase on top of this. We also need to get the RHEL 9.6 content mirrored in prow so we can update our CI.

jlebon:
> We also need to get the RHEL 9.6 content mirrored in prow so we can update our CI.

FTR, that request is tracked here: https://issues.redhat.com/browse/DPTP-4257

PR #1649, opened 2024-11-15
NO-JIRA: Containerfile: use correct tag in FROM

jlebon:
I think at the time I wrote this, I hadn't set up the pipeline yet for it, so I wasn't sure what its tag name was going to me. We know it now, so just fix this.


jlebon:
/hold

I have another PR internally to tweak the tag names to conform to what this is using.

dustymabe (approved):
/lgtm

PR #1650, opened 2024-11-19
NO-JIRA: ci/get-ocp-repo.sh: generalize and reuse

jlebon:
I'm looking at building the node image in OpenShift CI and of course we
also need to fetch the OCP repos in that flow too.

Let's make `get-ocp-repo.sh` support all the use cases it needs, which
allows us to dedupe with `ci/prow-entrypoint.sh`.

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -1,3 +1,9 @@
+metadata:
+  # This should match the /etc/os-release manipulation we do below when
+  # injecting `OPENSHIFT_VERSION`. It's used by CI to determine the repos to
+  # inject when building the layered image.
+  ocp_version: "4.18"
+
 packages:
   # The packages below are required by OpenShift/OKD
   # but are not present in CentOS Stream and RHEL.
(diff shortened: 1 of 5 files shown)


jlebon:
Let's wait for branching to get this one in.
/hold

jlebon:
> CI output looks weird:

Yeah, it requires https://github.com/openshift/os/pull/1651.

packages-openshift.yaml
```diff
@@ -1,3 +1,9 @@
+metadata:
+  # This should match the /etc/os-release manipulation we do below when
+  # injecting `OPENSHIFT_VERSION`. It's used by CI to determine the repos to
+  # inject when building the layered image.
+  ocp_version: "4.18"
```
travier, line 5: Aren't we 4.19 now?

jlebon: Yeah, this will be bumped along with the other 4.18 references in the repo.

mike-nguyen:
I'm working on the version bump. We have a openshift/release PR that hasn't merged yet.

PR #1654, opened 2024-11-26
NO-JIRA: bump version to 4.19

mike-nguyen:
4.18 branched so bump the versioni in main development to 4.19.

Note: CI hasn't been updated to use 9.6 repos yet.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -138,6 +138,7 @@ RHCOS/OCP version | RHEL version
 4.16 | 9.4 EUS
 4.17 | 9.4 EUS
 4.18 | 9.4 EUS
+4.19 | 9.6 EUS
 
 ## Q: How do I determine what version of an RPM is included in an RHCOS release?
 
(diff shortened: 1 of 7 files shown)


mike-nguyen:
I think we have some fallout from https://github.com/openshift/release/pull/59000/ for the `ci/prow/validate`.  Moving from `fcos-buildroot` to the `cosa` container, we lost `shellcheck` and unable to install it on the running container due to privileges.

jlebon:
Minor typo in commit message: s/versioni/version/

It'll take some time until cosa is rebuilt and propagates to the build cluster. I think the auto-retest here won't work because it needs to actually rebuild the image too. So we'll need `/test images`. But no point in doing that until we at least see the latest tag in https://quay.io/repository/coreos-assembler/coreos-assembler?tab=tags&tag=latest get bumped.

mike-nguyen:
Fixed the commit message but it removed the `lgtm` label.

Issue #1656, opened 2024-11-28
[rhel-9.6] ext.config.shared.var-mount.luks failing


c4rt0:
When building a rhel-9.4 & 9.6 variants on x86_64, the ext.config.shared.var-mount.luks fails.
```
[2024-11-28T12:53:38.065Z] --- FAIL: ext.config.shared.var-mount.luks (27.24s)
[2024-11-28T12:53:38.065Z]         harness.go:1823: mach.Start() failed: machine 6fb46fd3-e03d-4132-a88b-c5209d3629ee entered emergency.target in initramfs
```

In the attached [console.txt](https://github.com/user-attachments/files/17948753/console.txt) an error is found:

```
[   21.257136] ignition[835]: Ignition failed: failed to create luks: binding clevis device: exit status 1: Cmd: "clevis" "luks" "bind" "-f" "-k" "/tmp/ignition-luks-1296438722" "-d" "/run/ignition/dev_aliases/dev/disk/by-partlabel/varlog" "sss" "{\"pins\":{\"tpm2\":{}},\"t\":1}" Stdout: "" Stderr: "/usr/bin/clevis-encrypt-tpm2: line 137: tpm2_getcap: command not found\nUnable to find non-empty PCR algorithm bank, please check output of tpm2_getcap pcrs\nUnable to perform encryption with PIN 'sss' and config '{\"pins\":{\"tpm2\":{}},\"t\":1}'\nError adding new binding to /run/ignition/dev_aliases/dev/disk/by-partlabel/varlog\n"
0;1;31m*[0m[0;31m*    [0m] A start job is running for Ignition (disks) (19s / no limit)
[   21.273424] systemd[1]: ignition-disks.service: Main process exited, code=exited, status=1/FAILURE
[   21.275361] systemd[1]: ignition-disks.service: Failed with result 'exit-code'.
[   21.277132] systemd[1]: Failed to start Ignition (disks).
[0;1;31mFAILED[0m] Failed to start [0;1;39mIgnition (disks)[0m].
See 'systemctl status ignition-disks.service' for details.
[   21.280439] systemd[1]: Dependency failed for Ignition Complete.
```

Attaching remaining artifacts for completeness:
[ignition.json](https://github.com/user-attachments/files/17948818/ignition.json)
[ignition-virtio-dump.txt](https://github.com/user-attachments/files/17948813/ignition-virtio-dump.txt)

jlebon:
[...]

PR #1660, opened 2024-11-29
NO-JIRA: Add tpm2_getcap to the initrd for Clevis TPM2 functionality

c4rt0:
The tpm2_getcap binary is required for Clevis TPM2 operations but was missing from the initrd.

See: https://github.com/openshift/os/issues/1656

### DIFF
--- a/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
+++ b/overlay.d/05rhcos/usr/lib/dracut/modules.d/40rhcos-fde/module-setup.sh
@@ -62,6 +62,7 @@ install() {
         tpm2_create
         tpm2_createpolicy
         tpm2_createprimary
+        tpm2_getcap
         tpm2_load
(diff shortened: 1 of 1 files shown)


jlebon:
What I meant by https://github.com/openshift/os/issues/1656#issuecomment-2506543720 is that once https://github.com/coreos/ignition/pull/1978 gets to RHEL, it should fix it here too. So I think instead we want to get that backported (actually to both Fedora and RHEL). And until that happens, we can denylist those tests.

c4rt0:
> What I meant by [#1656 (comment)](https://github.com/openshift/os/issues/1656#issuecomment-2506543720) is that once [coreos/ignition#1978](https://github.com/coreos/ignition/pull/1978) gets to RHEL, it should fix it here too. So I think instead we want to get that backported (actually to both Fedora and RHEL). And until that happens, we can denylist those tests.

Ouch, cool - thanks for the clarification.
I will add any related tests to denylist instead.

PR #1661, opened 2024-11-29
NO-JIRA: Miscellanous node image layering-related patches

jlebon:
A lot of changes here, but the main goal is to reduce the diff between the traditional variants like `ocp-rhel-9.6` and a node image built on to of the newer el-only variants like `rhel-9.6`. The smaller that diff is, the lower risk it becomes to switch over to it.

See individual commit messages.


jlebon:
OK cool, I've also now just validated using a layered node image built on top of this PR can successfully be deployed in a cluster without anything seemingly blowing up:

```
$ cat layered-node-image.yaml
apiVersion: machineconfiguration.openshift.io/v1
kind: MachineConfig
metadata:
  labels:
    machineconfiguration.openshift.io/role: "worker"
  name: layered-node-image
spec:
  osImageURL: quay.io/jlebon/rhel-coreos@sha256:9961b57a71e8030b647d03c815ae2900722f46077bafe42fc79c88e6e688fb59

$ oc create -f layered-node-image.yaml

$ oc debug node/ip-10-0-59-92.us-west-2.compute.internal -- chroot /host rpm-ostree status
Starting pod/ip-10-0-59-92us-west-2computeinternal-debug-vfxsg ...
To use host binaries, run `chroot /host`
State: idle
Deployments:
* ostree-unverified-registry:quay.io/jlebon/rhel-coreos@sha256:9961b57a71e8030b647d03c815ae2900722f46077bafe42fc79c88e6e688fb59
                   Digest: sha256:9961b57a71e8030b647d03c815ae2900722f46077bafe42fc79c88e6e688fb59
                  Version: 9.4.202411280418-0 (2024-11-29T16:24:09Z)

Removing debug pod ...
```

So e.g. the kubelet running on that host is from a `dnf install -y openshift-kubelet` in a layered build. :tada:

jlebon:
The versioning there is TBD. I initially was going to make it 418.94... as it is today, but now I'm thinking we don't do that and just inherit the same RHEL-based version string from the base RHCOS image. That might break things, we'll see...

packages-openshift.yaml
```diff
@@ -142,4 +142,16 @@ postprocess:
...
+      cat > /usr/lib/tmpfiles.d/openshift.conf << EOF
+    L /opt/cni - - - - ../../usr/lib/opt/cni
+    d /var/lib/cni 0755 root root - -
+    d /var/lib/cni/bin 0755 root root - -
+    d /var/lib/containers 0755 root root - -
+    d /var/lib/openvswitch 0755 root root - -
+    d /var/lib/openvswitch/pki 0755 root root - -
+    d /var/log/openvswitch 0750 openvswitch hugetlbfs - -
+    d /var/lib/unbound 0755 unbound unbound - -
+    EOF
```
jbtrystram, line 161: The indentation looks off here

jbtrystram (approved):
Nice series ! Thanks for the detailed commit messages.
/lgtm

packages-openshift.yaml
jlebon: It's actually correct. :) It's a heredoc, and so would include the whitespace in the file we're writing if we indented it.

Issue #1662, opened 2024-12-02
 [rhel-9.6] ext.config.shared.root-reprovision.luks.autosave-xfs failing


c4rt0:
When building rhel-9.6 on x86_64, the ext.config.shared.root-reprovision.luks.autosave-xfs fails.
```
[2024-11-28T12:53:38.065Z] --- FAIL: ext.config.shared.var-mount.luks (27.24s)
[2024-11-28T12:53:38.065Z]         harness.go:1823: mach.Start() failed: machine 6fb46fd3-e03d-4132-a88b-c5209d3629ee entered emergency.target in initramfs
```

In the attached [console.txt](https://github.com/user-attachments/files/17979482/console.txt) an error is found:

```
[   34.306369] ignition[968]: Ignition failed: failed to create luks: binding clevis device: exit status 1: Cmd: "clevis" "luks" "bind" "-f" "-k" "/tmp/ignition-luks-704535105" "-d" "/run/ignition/dev_aliases/dev/disk/by-partlabel/root" "sss" "{\"pins\":{\"tpm2\":{}},\"t\":1}" Stdout: "Warning: keyslot operation could fail as it requires more than available memory.\n" Stderr: "/usr/bin/clevis-encrypt-tpm2: line 137: tpm2_getcap: command not found\nUnable to find non-empty PCR algorithm bank, please check output of tpm2_getcap pcrs\nUnable to perform encryption with PIN 'sss' and config '{\"pins\":{\"tpm2\":{}},\"t\":1}'\nError adding new binding to /run/ignition/dev_aliases/dev/disk/by-partlabel/root\n"
[   34.321149] systemd[1]: ignition-disks.service: Main process exited, code=exited, status=1/FAILURE
[   34.322970] systemd[1]: ignition-disks.service: Failed with result 'exit-code'.
```

This is very likely related to other issues where we wait for the latest release of clevis, ie.:

https://github.com/coreos/fedora-coreos-tracker/issues/1836
https://github.com/openshift/os/issues/1656

and should be fixed once clevis-21-207.el9 lands.

jlebon:
This is fixed now.

PR #1663, opened 2024-12-02
NO-JIRA: denylist: add ext.config.shared.root-reprovision.luks.autosave-xfs

c4rt0:
rhel-9.6 builds are now failing with the above. Let's denylist it until a new version of clevis lands.

See: https://github.com/openshift/os/issues/1662


jbtrystram (approved):
/lgtm

waiting on clevis-21-207.el9
https://github.com/openshift/os/issues/1656#issuecomment-2507952617

PR #1668, opened 2024-12-04
NO-JIRA: extension-rhel-9.6.yaml: use RHEL 9.6 content

mike-nguyen:
c9s-nfv was incorrectly used instead of rhel-9.6-nfv causing errors building the extensions container.

CI would have caught this but we don't have the RHEL 9.6 repos mirrored in Prow yet.


dustymabe (approved):
/lgtm

Issue #1630, opened 2024-10-07
Clevis update on centos stream 9 results in boot failure


Prashanth684:
A recent upgrade to the Clevis package from [clevis-20-200.el9](https://kojihub.stream.centos.org/koji/buildinfo?buildID=61659) to [clevis-21-201.el9](https://kojihub.stream.centos.org/koji/buildinfo?buildID=68429) pulls in the opensc and pcsc-lite packages, causing this error:

```
[    1.931395] dracut-initqueue[621]: pcscd: unrecognized option '--disable-polkit'
[    1.742582] dracut-initqueue[621]: pcscd: unrecognized option '--disable-polkit'
[    1.933525] dracut-initqueue[621]: Usage: pcscd options
[    1.933837] dracut-initqueue[621]: Options:
[[0;32m  OK  [[    1.934124] dracut-initqueue[621]:   -a, --apdu		log APDU commands and results
0m] Found device [0;1;39mAmazon Elastic Block Store root[0m.[    1.934449] dracut-initqueue[621]:   -c, --config		path to reader.conf
[    1.934737] dracut-initqueue[621]:   -f, --foreground	run in foreground (no daemon),

[    1.935028] dracut-initqueue[621]: 			send logs to stdout instead of syslog
[    1.935341] dracut-initqueue[621]:   -T, --color		force use of colored logs
[    1.935628] dracut-initqueue[621]:   -h, --help		display usage information
[    1.935912] dracut-initqueue[621]:   -H, --hotplug		ask the daemon to rescan the available readers
[[0;32m  OK  [0m] Reached target [0;1;39mInit[    1.936468] dracut-initqueue[621]:   -v, --version		display the program version number
rd Root Device[[    1.936748] dracut-initqueue[621]:   -d, --debug		display lower level debug messages
0m.
[    1.937028] dracut-initqueue[621]:   -i, --info		display info level debug messages
[    1.937316] dracut-initqueue[621]:   -e  --error		display error level debug messages (default level)
[    1.937834] dracut-initqueue[621]:   -C  --critical	display critical only level debug messages
[    1.938109] dracut-initqueue[621]:   --force-reader-polling ignore the IFD_GENERATE_HOTPLUG reader capability
[    1.938665] dracut-initqueue[621]:   -t, --max-thread	maximum number of threads (default 200)
[    1.938948] dracut-initqueue[621]:   -s, --max-card-handle-per-thread	maximum number of card handle per thread (default: 200)
[...]

PR #1675, opened 2024-12-09
NO-JIRA: Remove clevis overrides for OKD

Prashanth684:
Due to https://github.com/openshift/os/issues/1630, we had to pin clevis to an older version. This was fixed in clevis-21-203.el9, so removing it.


dustymabe (approved):
/lgtm

PR #1676, opened 2024-12-09
NO-JIRA: packages-openshift: fix bash conditional syntax

jlebon:
packages-openshift: fix bash conditional syntax

When using the globby checks, we need to use the bash-specific double
brackets, not single. Also fix the appending of "CoreOS" to the `NAME`
variable.

Fixes 621ac80 ("packages-openshift: fix duplicate "CoreOS" in motd").

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -122,8 +122,8 @@ postprocess:
     fi
     # in the el-only variants, we already have CoreOS in the NAME, so don't
     # re-add it when building the node image
-    if [ $NAME != *CoreOS* ]; then
-      NAME=$NAME CoreOS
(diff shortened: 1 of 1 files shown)


dustymabe:
> When using the globby checks

Interesting.. I don't think I've ever done this before.. I've always used matching with `=~`

dustymabe (approved):
/lgtm

jlebon:
Apparently, I didn't test this when I originally wrote it. Fixed another bug in that area.

aaradhak (approved):
/lgtm

dustymabe (approved):
/lgtm

PR #1683, opened 2024-12-09
NO-JIRA: common.yaml: disable composefs on ppc64le for now

jlebon:
There are issues with composefs on ppc64le right now: https://issues.redhat.com/browse/RHEL-70199

Work around this for now by disabling composefs on ppc64le only.

We don't want to ship to production with this kind of delta across arches, but at least for now this allows us to keep running tests on arches where it does work.

Note this will also run on 9.4 variants, but it's harmless there since `prepare-root.conf` doesn't exist.

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -57,6 +57,15 @@ documentation: false
 recommends: true
 
 postprocess:
+  - |
+    #!/usr/bin/bash
+    set -xeuo pipefail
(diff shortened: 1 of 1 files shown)


jlebon:
I think we should add a new non-exclusive test to f-c-c to verify that composefs is on, and then here we would snooze that test for e.g. 1 month or so.

dustymabe (approved):
/lgtm

PR #1685, opened 2024-12-10
NO-JIRA: tests/replace-rt-kernel: ensure we always pick a different EVR

jlebon:
Right now, only on c9s-based systems, we have logic to ensure we pick a different EVR. But even in the RHEL case, we need this logic because we might be in the period where c9s and RHEL are in sync and so the kernel builds have the same EVRs in both RHEL and c9s.

Work around this by just always doing the repoquery and filtering.

This is technically an rpm-ostree bug I think, but highly unlikely to happen in the wild because you normally replace with a kernel from the same vendor.

### DIFF
--- a/tests/kola/rpm-ostree/replace-rt-kernel/test.sh
+++ b/tests/kola/rpm-ostree/replace-rt-kernel/test.sh
@@ -42,12 +42,7 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
     runv sed -i '/\[baseos\]/,/^ *\[/ s/enabled=0/enabled=1/' /etc/yum.repos.d/cs.repo
     runv sed -i '/\[appstream\]/,/^ *\[/ s/enabled=0/enabled=1/' /etc/yum.repos.d/cs.repo
 
-    evr=
-    if rpm -q centos-stream-release; then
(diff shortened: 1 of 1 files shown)


jlebon:
Didn't test this yet.

dustymabe (approved):
/lgtm

mike-nguyen:
/retitle NO-JIRA: tests/replace-rt-kernel: ensure we always pick a different EVR

jlebon:
Without patch:

```
$ kola run ext.config.rpm-ostree.replace-rt-kernel; bell
...
=== RUN   ext.config.rpm-ostree.replace-rt-kernel
...
Dec 10 15:13:52 qemu0 kola-runext-test.sh[2237]: Resolving dependencies...done
Dec 10 15:13:52 qemu0 kola-runext-test.sh[2237]: error: Request to reinstall exact base package versions: kernel kernel-core kernel-modules kernel-modules-core kernel-modules-extra
```

With patch:

```
$ kola run ext.config.rpm-ostree.replace-rt-kernel; bell
...
=== RUN   ext.config.rpm-ostree.replace-rt-kernel
--- PASS: ext.config.rpm-ostree.replace-rt-kernel (326.57s)
```

Let's just get this in and have another go in the pipeline.

PR #1688, opened 2024-12-10
NO-JIRA: kola-denylist: add ext.config.shared.boot.bootupd-validate

jlebon:
This test is failing on el9.6/c9s. It needs investigation but let's denylist it for now to unblock things.


mike-nguyen:
/retitle NO-JIRA: kola-denylist: add ext.config.shared.boot.bootupd-validate

Issue #1617, opened 2024-09-10
simplify date component of versioning scheme


dustymabe:
Up until this point our versioning has looked something like:

- `418.94.202409090849-0`

We are moving to a RHEL based base layer and the versioning [will look](https://github.com/openshift/os/pull/1615) something like:

- `9.4.202409090849-0`

As part of this I propose that we simplify the date component of the version to just daily precision. I argue that the hour/minute in the component make that part of the version too hard to read for humans such that they often ignore it (I certainly do).

The proposal here is to move to something like this for multiple successive builds on a single day:

- `9.4.20240909-0`
- `9.4.20240909-1`
- `9.4.20240909-2`

This is much easier to parse for humans and makes it clear when it was produced. I think of it like your computer having hardware assisted features for decoding things (like compression or multimedia codecs). It's much easier for the computer to do those things if they have the hardware to help. We as humans easily parse `20240909`, but not so much `202409090849`.

jlebon:
From what I remember, the date used to be shorter. It was made longer to make it more unique when the different arches were built by separate Jenkins instances (don't quite recall why, I think it was related to clobbering? git likely has the answer).

This would be trivial to change, but I'm quite sure there are things out there parsing the current format. They might not care about the date component of it, but the parser/regex might currently be accounting for them.

Definitely need to raise awareness on this before changing it.

dustymabe:
> Definitely need to raise awareness on this before changing it.

Agree. Where's the best place to go? Mailing list? `openshift/enhancements`?

jlebon:
Probably not worth a full-blown enhancement. aos-devel to start sounds good. And maybe linking to the thread in the ART and DPTP channels.

PR #1689, opened 2024-12-10
NO-JIRA: el-only variants: drop hours and minutes from version timestamp

jlebon:
As part of the layered node image work, I was initially thinking we'd still keep the e.g. `419.96...` format there, but I'm now thinking we _don't_ override the versioning info from the base RHCOS layer anymore. So e.g. `bootc status` on an OCP node would still show `9.6...`. This is potentially breaking; I'm ready to help fix any issues that occur from this change in any OCP code that parses the version string today.

So then, it becomes much more enticing to also bundle with this change the suggestion in #1617 since any consumers  will have to adapt to the new version string anyway.

So let's do that.

Closes: #1617

### DIFF
--- a/manifest-c9s.yaml
+++ b/manifest-c9s.yaml
@@ -21,7 +21,7 @@ repos:
 # Eventually we should try to build these images as part of the c9s composes.
 # In that case, the versioning should instead be exactly the same as the pungi
 # compose ID.
-automatic-version-prefix: "9.<date:%Y%m%d>"
+automatic-version-prefix: "9.0.<date:%Y%m%d>"
 # This ensures we're semver-compatible which OpenShift wants
 automatic-version-suffix: "-"
 
--- a/manifest-c9s.yaml
+++ b/manifest-c9s.yaml
@@ -18,9 +18,6 @@ repos:
   - c9s-baseos
   - c9s-appstream
 
-# Eventually we should try to build these images as part of the c9s composes.
-# In that case, the versioning should instead be exactly the same as the pungi
-# compose ID.
 automatic-version-prefix: "9.0.<date:%Y%m%d>"
 # This ensures we're semver-compatible which OpenShift wants
 automatic-version-suffix: "-"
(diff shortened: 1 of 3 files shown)


jlebon:
> As part of the layered node image work, I was initially thinking we'd still keep the e.g. `419.96...` format there, but I'm now thinking we _don't_ override the versioning info from the base RHCOS layer anymore.

See https://github.com/openshift/enhancements/pull/1727 for more information on this.

manifest-c9s.yaml
```diff
@@ -21,7 +21,7 @@ repos:
 # Eventually we should try to build these images as part of the c9s composes.
 # In that case, the versioning should instead be exactly the same as the pungi
 # compose ID.
-automatic-version-prefix: "9.<date:%Y%m%d%H%M>"
```
dustymabe, line 24: I know CentOS stream doesn't have point releases but it would be really nice to have the RHEL and C9S eqivalent builds have the same number of components in their version. I wonder if make the c9s one have a `X` in the component would be unpalatable.

dustymabe: `"9.X.<date:%Y%m%d>"`

jlebon: Actually, the format I suggested here isn't semver-compatible. It needs to be three components. Letters are not allowed either, so let's just go with `0`? It's awkward but, any randomly large number feels too arbitrary.

dustymabe: ok sounds good

dustymabe: another option (not sure how much I prefer one or the other) is:

- centos stream: `9.<date:%Y%m%d>.spin`
- rhel 9.6: `96.<date:%Y%m%d>.spin`

where `spin` is the number of times we've had a build that day instead of `-spin` like we have now.

jlebon: I'd like to get away from the "squashed" version numbering that we've been doing and have version strings that look more natural. Not just for optics, but also e.g. it makes parsing it easier. So I think I prefer the previous option.

jlebon:
Updated and also added another commit to drop some comments!

dustymabe (approved):
LGTM

PR #1695, opened 2024-12-11
[release-4.18] OCPBUGS-46012: pin libreswan version to 4.6 in rhcos extension

sdodson:
Pin to full NEVR because otherwise we'd pick up the older 9.1 package


zshi-redhat:
you beat me to it :-)
/lgtm

PR #1699, opened 2024-12-14

coreosbot-releng:

```
Aashish Radhakrishnan (1):
      denylist: remove multipath.partition denial

Dusty Mabe (1):
      live: update readme with more accurate comment

HuijingHei (1):
      bootupd-validate: remove condition as we have newer bootupd
```


HuijingHei:
`bootupd-validate` failed in `ci/prow/rhcos-9-build-test-qemu`:
```
OSTREE_VERSION=419.94.202412200020-0
--- FAIL: ext.config.shared.boot.bootupd-validate (52.72s)
```
Need to skip on rhel9.4, see https://github.com/openshift/os/pull/1705

PR #1702, opened 2024-12-19
NO-JIRA: ci/get-ocp-repo.sh: add `--output-dir`

jlebon:
In the node layering case, we're directly bind-mounting the build context dir. It's mounted read-only, so we can't write `ocp.repo` there. And even if we could, we shouldn't modify the build context if we don't need to.

Add a new `--output-dir` option to allow changing the directory to which to write `ocp.repo`. Use this in the node layering flow to output to `/run/yum.repos.d`, which is an already established API for transient repo definitions.

### DIFF
--- a/ci/get-ocp-repo.sh
+++ b/ci/get-ocp-repo.sh
@@ -51,6 +51,8 @@ if [ -n "$ocp_manifest" ]; then
     info "Got OpenShift version $ocp_version from $ocp_manifest"
     # osname is used lower down, so set it
     osname=$(source /usr/lib/os-release; if [ $ID == centos ]; then echo scos; fi)
+
+    output_dir=$(dirname "$ocp_manifest")
 else
     [ -n "$cosa_workdir" ]
     # --cosa-workdir path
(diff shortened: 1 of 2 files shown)


jlebon:
This was successfully tested in https://github.com/openshift/release/pull/59490 (well, the node layering flow).

ci/get-ocp-repo.sh
```diff
@@ -25,20 +29,24 @@ info() {
...
+    case "$1" in
+        -h | --help) print_usage_and_exit;;
+        --cosa-workdir) cosa_workdir=$2; shift;;
+        --ocp-layer) ocp_manifest=$2; shift;;
+        --output-dir) output_dir=$2; shift;;
+        --) break;;
+        *) echo "$0: invalid argument: $1" >&2; exit 1;;
+    esac
+    shift
+done
```
dustymabe, line 49: previously `--cosa-workdir` and `--ocp-layer` were exclusive, or at least only the first one would be considered. Do we want to do anything specific if a user passed both here?

dustymabe (approved):
/approve

ci/get-ocp-repo.sh
jlebon: Yeah... when switching to getopt, keeping that check didn't feel worth the lines. The current behaviour of just letting `--ocp-layer` win seemed fine to me. This script doesn't need a lot of polish. All users of it are in this repo.

jlebon:
This is only used in CI, which passed.

PR #1706, opened 2024-12-20
I added detailed comments at each step explaining the purpose of the commands.

infyomer:
Only the necessary operations for the final build are retained, and the steps are streamlined.


infyomer:
Modifyy

travier:
Sorry but this is slightly wrong and feels AI generated.
/close

PR #1703, opened 2024-12-19
NO-JIRA: ci/prow: test on RHEL 9.6

jlebon:
We now have RHEL 9.6 content in the CI mirror so we can run tests.

PR #1714, opened 2025-01-21
NO-JIRA: manifest-ocp-rhel-9.6: get toolbox from ART plashet

mike-nguyen:
RHCOS is trying to move to a point where all container tool packages come from RHEL instead of the ART plashet[0]. This is part of the effort to have a RHEL only bootstrap image and a layered container image for the node. Unfortunately there is already a package in RHEL called toolbox which conflicts with the RHCOS toolbox.

Ensure that RHCOS gets the RHCOS toolbox by only getting the package from the ART plashet.

[0] https://issues.redhat.com/browse/RHELPLAN-170883

### DIFF
--- a/packages-rhcos.yaml
+++ b/packages-rhcos.yaml
@@ -15,7 +15,6 @@ packages:
   - nss-altfiles
   - ostree
   - rpm-ostree
-  - toolbox
   # for kdump:
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
   - kexec-tools
--- a/manifest-ocp-rhel-9.4.yaml
+++ b/manifest-ocp-rhel-9.4.yaml
@@ -98,3 +98,4 @@ repo-packages:
       - cri-tools
       - openshift-clients
       - openshift-kubelet
+      - toolbox
--- a/manifest-ocp-rhel-9.6.yaml
+++ b/manifest-ocp-rhel-9.6.yaml
@@ -97,7 +97,9 @@ repo-packages:
       - podman
       - runc
       - skopeo
-      - toolbox
   - repo: rhel-9.6-appstream
     packages:
       - ignition
+  - repo: rhel-9.6-server-ose-4.19
+    packages:
+      - toolbox


jbtrystram (approved):
lgtm

packages-rhcos.yaml
```diff
@@ -15,7 +15,11 @@ packages:
   - nss-altfiles
   - ostree
   - rpm-ostree
-  - toolbox
   # for kdump:
   # https://github.com/coreos/fedora-coreos-tracker/issues/622
   - kexec-tools
+
+repo-packages:
```
travier, line 22: I don't think that it's a key supported by the apply-manifest script yet.

mike-nguyen:
> ```
> error: Installing packages: No matches for 'toolbox' in repo 'rhel-9.6-server-ose-4.19'
> ```
> 
> We also have to build it there

It should be there already as it is pointing to the same location as rhel-9.4-server-ose-4.19

packages-rhcos.yaml
mike-nguyen: @travier I built it locally and it seemed to work.  I'll have to move it up into `manifest-ocp-rhel-9.6.yaml` because our CI is testing 9.4 (9.6 in the works https://github.com/openshift/os/pull/1703) and adding `repo-packages` in the `package-rhcos.yaml` is making the 9.4 CI look in the `rhel-9.6-server-ose-4.19` repo which isn't defined for 9.4 CI.

dustymabe:
I think (for now) we are stilling building 4.19-9.4. I think we are planning to drop this very soon, but this change will (I think) cause toolbox to drop out of that stream.. Am I right? Do we care?

dustymabe (approved):

mike-nguyen:
> I think (for now) we are stilling building 4.19-9.4. I think we are planning to drop this very soon, but this change will (I think) cause toolbox to drop out of that stream.. Am I right? Do we care?



> I think (for now) we are stilling building 4.19-9.4. I think we are planning to drop this very soon, but this change will (I think) cause toolbox to drop out of that stream.. Am I right? Do we care?

@dustymabe  Nice catch! We probably want to keep it in just in case we need to switch back. I'll add toolbox to the ocp-rhel-9.4 manifest. We were lucky before because we never pinned toolbox to a repo and the rhcos toolbox just happened to be newer than the upstream toolbox.  When we started using 9.6 content, the upstream toolbox version was newer and we incorrectly picked it up.

dustymabe (approved):
/lgtm

PR #1716, opened 2025-01-21
NO-JIRA: change default variant to ocp-rhel-9.6

mike-nguyen:


manifest-okd-c9s.yaml
```diff
@@ -22,8 +22,7 @@ repos:
   # CentOS Cloud SIG repo for cri-o, cri-tools and conmon-rs
   - c9s-sig-cloud-okd
   # Include RHCOS 9 repo for oc, hyperkube
-  # rhel-9.6-server-ose-4.19 is not mirrored in CI yet
-  - rhel-9.4-server-ose-4.19
+  - rhel-9.6-server-ose-4.19 is not mirrored in CI yet
```
jlebon, line 25: This looks off.

jlebon (commented):
Thanks!

manifest-okd-c9s.yaml
mike-nguyen: Fixed

mike-nguyen:

This should only change the default variant built and should not change anything related to builds that flow into OCP.

PR #1719, opened 2025-01-27
COS-2873: extensions/Dockerfile: Get extensions rpm list

ravanelli:
- Generate JSON file listing installed extension packages;
 - List `/usr/share/rpm-ostree/extensions/` for `.rpm` files, extracts package names and versions from the filenames, and generates a JSON file at `/extensions/extensions.json`. Where the volume is mounted;
 - It is needed to generated the extensions package list in meta.json.

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -29,6 +29,16 @@ RUN rm -f /etc/yum.repos.d/*.repo \
 RUN dnf install -y createrepo_c
 RUN createrepo_c /usr/share/rpm-ostree/extensions/
 
+# Generate extensions.json for meta.json, written to a bind-mounted path during the build.
+# Use dnf repoquery to print 'name: version,' for each RPM
+# sed to remove the comma from the last RPM
+RUN sh -c 'echo "{" > /tmp/extensions.json && \
+dnf repoquery --repofrompath=extensions,/usr/share/rpm-ostree/extensions/ \
+  --quiet --disablerepo=* --enablerepo=extensions \
+  --queryformat "\"%{name}\": \"%{version}\"," | \
+sed "$ s/,$//" >> /tmp/extensions.json && \
+echo "}" >> /tmp/extensions.json'
+
 ## Final container that has the extensions repo dir
 FROM registry.access.redhat.com/ubi9/ubi:latest
 COPY --from=builder /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/


extensions/Dockerfile
```diff
@@ -32,3 +32,14 @@ RUN createrepo_c /usr/share/rpm-ostree/extensions/
...
+RUN mkdir -p /extensions
+RUN echo "{" > /extensions/extensions.json && \
+    files=( /usr/share/rpm-ostree/extensions/*.rpm ); \
+    for entry in "${files[@]}"; do \
+        package_name=$(basename "$entry" .rpm | sed -E 's/-[0-9]+.*$//'); \
+        package_version=$(basename "$entry" .rpm | sed -E 's/^[^-]+-(.*)$/\1/'); \
+        echo "  \"$package_name\": \"$package_version\"," >> /extensions/extensions.json; \
+    done && \
+    sed -i '$ s/,$//' /extensions/extensions.json && \
+    echo "}" >> /extensions/extensions.json
```
dustymabe, line 45: I think relying on the filenames of the RPMs and parsing those will cause us headaches over time. Let's actually query the yum repo and use get the information that way?

```
(
# Opening brace for JSON document
echo '{'
# Use dnf repoquery to print 'name: version,' for each RPM
# sed to remove the comma from the last RPM
dnf repoquery --repofrompath=extensions,/usr/share/rpm-ostree/extensions/ \
        --quiet --disablerepo=* --enablerepo=extensions                   \
        --queryformat '"%{name}": "%{version}",' | sed '$ s/,$//g'
# Closing brace for JSON document
echo '}'
) > extensions.json
```

extensions/Dockerfile
```diff
@@ -32,3 +32,14 @@ RUN createrepo_c /usr/share/rpm-ostree/extensions/
 ## Final container that has the extensions repo dir
 FROM registry.access.redhat.com/ubi9/ubi:latest
 COPY --from=builder /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
+
```
dustymabe, line 35: probably need to add a comment above this block that mentions where we are writing `extensions.json` is bind mounted in during the build and what we are doing here will be used to generated the meta.json entry for extensions. 

you can link to the COSA code that does the `--volume` part for added clarity

dustymabe (approved):
LGTM

dustymabe:
I think we still need to update https://github.com/coreos/coreos-assembler/pull/4009/files ?

extensions/Dockerfile
```diff
@@ -32,3 +32,13 @@ RUN createrepo_c /usr/share/rpm-ostree/extensions/
...
+# Generate extensions.json for meta.json, written to a bind-mounted path during the build.
+# Use dnf repoquery to print 'name: version,' for each RPM
+# sed to remove the comma from the last RPM
+RUN sh -c 'echo "{" > /tmp/extensions.json && \
```
jlebon, line 39: Hmm, pretty sure even though this doesn't create anything in the rootfs, this is still going to add a layer to the final image. Instead, we can run this from the `builder` stage right after creating the repo.

jlebon (commented):
One comment, but otherwise looks sane to me!
/hold

extensions/Dockerfile
dustymabe: SGTM

dustymabe (approved):

Issue #1673, opened 2024-12-06
c9s: ext.config.shared.rpm-ostree.kernel-replace fails


c4rt0:
Needs investigation.

```
[2024-12-05T13:50:11.182Z] --- FAIL: ext.config.shared.rpm-ostree.kernel-replace (1206.05s)
[2024-12-05T13:50:11.182Z]         harness.go:106: TIMEOUT[20m0s]: ssh: sudo /usr/local/bin/kolet run-test-unit kola-runext.service
[2024-12-05T13:50:11.182Z]         harness.go:106: TIMEOUT[20m0s]: ssh: journalctl -t kola-runext-kernel-replace
```

I failed to find useful info in [console.txt](https://github.com/user-attachments/files/18025175/console.txt).

[journal.txt](https://github.com/user-attachments/files/18025296/journal.txt) presents:
```
Dec  5 13:01:40.836001 rpm-ostreed.service[2992]: dracut:     microcode_ctl: kernel version "5.14.0-533.el9.x86_64" failed early load check for "intel-06-4e-03", skipping
```

Attaching also [journal-raw.txt.gz](https://github.com/user-attachments/files/18025392/journal-raw.txt.gz)

jbtrystram:
@c4rt0 maybe we can try to update the test to replace to a kernel more recent than  "5.14.0-533.el9.x86_64" ?

jbtrystram:
> @c4rt0 maybe we can try to update the test to replace to a kernel more recent than "5.14.0-533.el9.x86_64" ?

we investigated this today and the kernel version the test tries to switch to is automatically pulled from what's available in the repos, so that is not a concern. let's investigate it more

c4rt0:
Upon further investigation, latest scos build passes the test.

from journal.txt:

```bash
Jan 20 15:31:41.210716 kola-runext.service[1380]: ++ . /etc/os-release
Jan 20 15:31:41.210916 kola-runext.service[1380]: +++ NAME='CentOS Stream CoreOS'
Jan 20 15:31:41.210971 kola-runext.service[1380]: +++ VERSION=9.0.20250120-0
Jan 20 15:31:41.211021 kola-runext.service[1380]: +++ ID=centos
Jan 20 15:31:41.211056 kola-runext.service[1380]: +++ ID_LIKE='rhel fedora'
Jan 20 15:31:41.211094 kola-runext.service[1380]: +++ VERSION_ID=9
[...]

PR #1727, opened 2025-01-30
NO-JIRA: kola-denylist: drop ext.config.shared.rpm-ostree.kernel-replace

c4rt0:
See: https://github.com/openshift/os/issues/1673


dustymabe (approved):
LGTM

Issue #1731, opened 2025-01-31
[rcos-9.6.s390x]: secex tests failed


nikita-dubrovskii:
Looks like new version of `genprotimg` has some issues:
```
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]: WARNING: Use of deprecated flag '-V'. Use '-v' or '--verbose' instead.
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]: Host-key document verification is disabled. The Secure Execution image may not be protected.
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]: Use host-key document at '/etc/se-hostkeys/ibm-z-hostkey-1'
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]: error: Cannot rename '/tmp/coreos-installer-bjZsqa/sdboot.part' to '/tmp/coreos-installer-bjZsqa/sdboot'
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]: Caused by:
Jan 31 15:38:40.433016 coreos-ignition-firstboot-complete.service[2730]:     File exists (os error 17)
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]: Error: Command {
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:     program: "genprotimg",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:     args: [
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "genprotimg",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "-V",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "-i",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "/boot/boot/ostree/rhcos-58ae35a9a20dafeb45886331103c4292f666660e1a8ff9257a58574e60b695de/vmlinuz-5.14.0-558.el9.s390x",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "-r",
Jan 31 15:38:40.462686 coreos-ignition-firstboot-complete.service[2730]:         "/tmp/initrdvs7XDj.img",
[...]

PR #1733, opened 2025-02-03
NO-JIRA: common.yaml: pin s390utils-base-2.33.1-2.el9

mike-nguyen:
genprotimg in newer version of s390utils-base is causing secex tests to fail.  

See https://github.com/openshift/os/issues/1731


marmijo (approved):

nikita-dubrovskii:
@mike-nguyen - probably here is a [Jira](https://issues.redhat.com/browse/RHEL-77592)

mike-nguyen:
/override scos-9-build-test-qemu

Failing on access to repo mirror.  This is not related to this change and CI doesn't cover s390x.

PR #1737, opened 2025-02-04
NO-JIRA: extensions: include release and arch in extensions.json

jlebon:
`%{version}` just prints the version field, but we want the epoch, release, and arch too. This matches what we previously would output before moving extensions building to a container.

There is code that will parse this metadata assuming that it's an EVRA:

https://github.com/openshift-eng/art-tools/blob/6a29949b2b2819afe00829646e6c7db9b784ff8a/doozer/doozerlib/rhcos.py#L248


ravanelli (approved):
/lgtm

PR #1739, opened 2025-02-05
[release-4.18] OCPBUGS-49833: extensions: include release and arch in extensions.json

ravanelli:
`%{version}` just prints the version field, but we want the epoch, release, and arch too. This matches what we previously would output before moving extensions building to a container.

There is code that will parse this metadata assuming that it's an EVRA:

https://github.com/openshift-eng/art-tools/blob/6a29949b2b2819afe00829646e6c7db9b784ff8a/doozer/doozerlib/rhcos.py#L248 (cherry picked from commit 51e149dfeafb2b0e6aa7a4a92a85eb508665818e)


mike-nguyen:
@ravanelli I used the bot and it merged https://github.com/openshift/os/pull/1738 sorry for the duplicated efforts

ravanelli:
Not at all, thanks @mike-nguyen!

PR #1735, opened 2025-02-03
NO-JIRA: common.yaml: re-enable composefs on el9

jlebon:
The composefs issue (https://issues.redhat.com/browse/RHEL-63985)
should now be fixed in the latest el9 kernel. So let's optimistically
re-enable it. Now that we have tests for now, CI will validate this for
us as well.

PR #1740, opened 2025-02-05
NO-JIRA: Containerfile: use secrets API for yum repo injection

jlebon:
Instead of the `/run/yum.repos.d` mechanism, use the secrets API which is already well-known and does what we want here, which is to avoid having to copy repo files into the image.

The `RUN --mount` option gracefully ignores the case where no secret was provided, which is also the semantic we want for this.

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -31,7 +31,10 @@ postprocess:
     # repos will be attempted to be fetched by rpm-ostree when doing node-local
     # kernel overrides today for e.g. kernel-rt.
     for x in $(find /etc/yum.repos.d/ -name '*.repo'); do
-      sed -i -e s,enabled=1,enabled=0, $x
+      # ignore repo files that are mountpoints since they're likely secrets
+      if ! findmnt "$x" &>/dev/null; then
+        sed -i -e s,enabled=1,enabled=0, $x
+      fi
     done
 
   # These enable librhsm which enables host subscriptions to work in containers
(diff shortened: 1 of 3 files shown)


cgwalters (approved):

packages-openshift.yaml
```diff
@@ -31,7 +31,10 @@ postprocess:
     # repos will be attempted to be fetched by rpm-ostree when doing node-local
     # kernel overrides today for e.g. kernel-rt.
     for x in $(find /etc/yum.repos.d/ -name '*.repo'); do
-      sed -i -e s,enabled=1,enabled=0, $x
+      # ignore repo files that are mountpoints since they're likely secrets
+      if ! findmnt "$x" &>/dev/null; then
+        sed -i -e s,enabled=1,enabled=0, $x
+      fi
```
dustymabe, line 37: findmt seems a bit ugly here, but I guess that's the best option?

only other thing I can think of is checking to see if a file is read-only before attempting to disable the repo

dustymabe (approved):
/approve

packages-openshift.yaml
cgwalters: `/usr/bin/mountpoint` is also a bit more direct of an API

jlebon: Ahh yes, forgot about `mountpoint`. Will switch to that in a future patch.

jlebon:
Just going to override CI. We're working on fixing the extensions container build.

/override ci/prow/images

packages-openshift.yaml
jlebon: > Ahh yes, forgot about `mountpoint`. Will switch to that in a future patch.

Part of https://github.com/openshift/os/pull/1735.

PR #1741, opened 2025-02-06
NO-JIRA: manifest-rhel-9.6: add early kernel repo

jlebon:
We need to enable the `rhel-9.6-early-kernel` repo to ensure that we still get access to the early kernels. Non-el-only variants (like `ocp-rhel-9.6`) get this already through their plashets.

(In fact, the `rhel-9.6-early-kernel` repo is just an OCP plashet in disguise; there are excludes added so that only the kernel is permitted. This doesn't break the "el-only" concept because those kernels _are_ RHEL-versioned. It's just that historically we've consumed them through OCP plashets, but ideally in the future we'd have a separate, clearly RHEL-versioned, repo for them... But ideally ideally, when we rebase to rhel-bootc, we have a rhel-bootc stream that uses those kernels that we would derive from.)

### DIFF
--- a/manifest-rhel-9.6.yaml
+++ b/manifest-rhel-9.6.yaml
@@ -16,6 +16,7 @@ include:
 repos:
   - rhel-9.6-baseos
   - rhel-9.6-appstream
+  - rhel-9.6-early-kernel
 
 automatic-version-prefix: "9.6.<date:%Y%m%d>"
 # This ensures we're semver-compatible which OpenShift wants


jlebon:
CI will need https://github.com/openshift/release/pull/61361.

This would've been cleaner probably as an rpm-ostree `repo-packages`-like feature. Though OTOH, the low-tech approach of using excludes is appealing and more universal (e.g. dnf can also use this).

aaradhak (approved):
LGTM

jlebon:
Just going to force this in. The extensions container is hitting the same issue as https://github.com/openshift/release/pull/61040#issuecomment-2634490968. But we're doing this to _fix_ the RHCOS pipeline in the first place.

/override ci/prow/e2e-aws
/override ci/prow/images

PR #1742, opened 2025-02-10
NO-JIRA: rework manifests and drop mentions of 9.4

dustymabe:
We've moved on to 9.6 for OCP 4.19+ so there is no need to keep the definitions around in our repo here.


ci/prow-entrypoint.sh
```diff
@@ -271,28 +271,16 @@ main() {
...
-            kola_test_qemu
-            ;;
-        "rhcos-9-build-test-qemu")
-            setup_user
-            cosa_init "ocp-rhel-9.4"
+            cosa_init "ocp-rhel-9.6"
             cosa_build
             kola_test_qemu
             ;;
-        "rhcos-9-build-test-metal")
```
jbtrystram, line 290: Let's keep the "9" for 9.6 and update openshift/release as well?

jbtrystram (changes requested):

ci/prow-entrypoint.sh
dustymabe: ehh. I kind of prefer the namespaced to subversion model.. here's an example of why:

today in https://github.com/openshift/os/pull/1735#issuecomment-2648845754 I was looking to see why the CI on this PR was failing and had to actually dig into the tests to see that `ci/prow/rhcos-9-build-test-qemu` is RHEL 9.4 based.

I'm OK with changing it back to the way it was if there are good reasons, but do prefer for RHEL to have major/minor in there.

jlebon: That sounds good to me. Though we do still need to update openshift/release to drop those tests.

jlebon (commented):
Can you also update the README?

ci/prow-entrypoint.sh
jbtrystram: Some of those may be generated through `make update`

jlebon: Yeah, it looks a lot worse than it is. The only file you need to manually change is https://github.com/openshift/release/blob/efd41d3fee52d517cb521a59343871cd440b01dc/ci-operator/config/openshift/os/openshift-os-master.yaml.

The awkward thing is that wiring in a new minor is a three step process:
- add the entrypoints here; they just do `exit 0`
- add the tests in openshift/release
- change `exit 0` to actually test and iterate here

which was the rationale for not including a minor version here.

It's true that it gets confusing when we set up a new minor though. One thing we could do is to have `rhcos-9-build-test-qemu` and `rhcos-9-next-build-test-qemu` instead. So in this case, the first would've been 9.4 and the second 9.6. But then when we purge 9.4, we make `rhcos-9-build-test-qemu` become 9.6, and neuter `-next` to just `exit 0` again.

dustymabe: ok I did a new upload here and also opened https://github.com/openshift/release/pull/61527

I went with `rhcos-9next` as the prefix. I wanted to combine the `9` with `next` to force my brain not to separate them.

dustymabe:
> Can you also update the README?

done

ci/prow-entrypoint.sh
```diff
@@ -271,39 +271,33 @@ main() {
...
             cosa_build
             kola_test_qemu
             ;;
         "rhcos-9-build-test-metal")
             setup_user
-            cosa_init "ocp-rhel-9.4"
+            cosa_init "ocp-rhel-9.6"
             cosa_build
             kola_test_metal
             ;;
```
jlebon, line 295: We can drop these too now. We should get https://github.com/openshift/release/pull/61527 in anyway first so CI won't block on those anymore.

ci/prow-entrypoint.sh
```diff
@@ -271,39 +271,33 @@ main() {
...
-        "rhcos-96-build-test-qemu")
-            setup_user
-            cosa_init "ocp-rhel-9.6"
-            cosa_build
-            kola_test_qemu
+        "rhcos-9next-build-test-qemu")
```
jlebon, line 296: The addition of these entrypoints could be a separate commit with its own rationale.

jlebon (approved):
Some comments but looks good overall. Thanks for cleaning this up!

ci/prow-entrypoint.sh
jlebon: Sorry, I crossed wires there.

jlebon:
The extensions container CI failure is being worked on.

/override ci/prow/images
/override ci/prow/e2e-aws

Issue #1745, opened 2025-02-13
[4.19-9.6] `ext.config.shared.containers.quadlet` failing


jbtrystram:
```
[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2506]: ++ jq -r '.[0].ImageName'

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: + [[ quay.io/fedora/fedora-minimal:latest != \q\u\a\y\.\i\o\/\f\e\d\o\r\a\/\f\e\d\o\r\a\-\m\i\n\i\m\a\l\:\l\a\t\e\s\t ]]

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2508]: ++ jq -r '.[0].NetworkSettings.Networks[].NetworkID'

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: + [[ fdafc6d8899df672e0892555b4540e39842fa788cea600d462b35694b266b00f != \s\y\s\t\e\m\d\-\t\e\s\t ]]

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: + fatal 'Container not using the correct network'

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: + echo 'Container not using the correct network'

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: Container not using the correct network

[2025-02-13T11:48:22.444Z] Feb 13 11:48:21 qemu0 kola-runext-test.sh[2274]: + exit 1
```

Podman version : [2025-02-13T11:34:27.962Z]   podman-5:5.4.0-1.el9.x86_64 (rhel-9.6-appstream)

travier:
Likely a bug fix / Docker compatibility fix that landed in the 5.4.0 release, should reach FCOS soon as well.

PR #1730, opened 2025-01-31
NO-JIRA: Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Michael Armijo (1):
      tests/quadlet: Fix network name retrieval in podman inspect
```

PR #1746, opened 2025-02-13
NO-JIRA: denylist: add ext.config.shared.containers.quadlet

jbtrystram:
See https://github.com/openshift/os/issues/1745

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -60,3 +60,8 @@
   warn: true
   arches:
     - s390x
+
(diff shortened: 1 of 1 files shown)


c4rt0 (approved):
/lgtm

dustymabe:
> I think we just need #1730 instead of this?

yes. it includes https://github.com/coreos/fedora-coreos-config/pull/3343

jbtrystram:
ah, thanks @jlebon  ! Sorry I missed that.
I'll close this then

PR #1750, opened 2025-02-17
NO-JIRA: overrides-c9s: add s390utils-base override for CentOS too

dustymabe:
~~As mentioned in [1] the mirrors have older versions and we're currently pinning an older version of s390utils-base [2] so we need to use these for now. I'm not sure of the reason to use the compose repos but if we change it back to that we should include the rationale in the commit and maybe a comment in the code.~~

~~[1] https://github.com/openshift/os/commit/88e41a0fef0dbe8e0b2d103cadd37217c8cd0933~~
~~[2] https://github.com/openshift/os/commit/71313d82c43d27fa26201f9ecd987b1716f31749~~

EDIT: See https://github.com/openshift/os/pull/1750#issuecomment-2669436000

Let's move the override added in [1] to a new file dedicated to
overrides. This will make it easier for us to identify and focus
on overcoming the issues related to the override.

[1] https://github.com/openshift/os/commit/71313d82c43d27fa26201f9ecd987b1716f31749

----

This adds the override from [1] and also enables the *mirror repos
since those are the repos that have older versiosn of packages in
CentOS Stream.

[1] https://github.com/openshift/os/commit/71313d82c43d27fa26201f9ecd987b1716f31749

### DIFF
--- a/c9s-mirror.repo
+++ b/c9s-mirror.repo
@@ -1,6 +1,7 @@
 # These are the official c9s repos. They are slower to update, but contain older
 # versions of packages, which is useful when pinning for lack of a "coreos-pool"
-# equivalent. They are unused when no pinning is needed.
+# equivalent. When no pinning is needed you may find the compose repo URLs
+# defined in c9s.repo are quicker to get new content.
 
 [c9s-baseos-mirror]
 name=CentOS Stream 9 - BaseOS
--- a/c9s.repo
+++ b/c9s.repo
@@ -1,3 +1,9 @@
+# These are compose repo URLs that represent the latest composes in
+# CentOS Stream 9. Sometimes these repos get content a little faster
+# than the mirror repos defined in c9s-mirror.repo, but they won't
+# have multiple versions of packages, which make them not ideal when
+# needing to pin on older package versions.
+
 [c9s-baseos]
 name=CentOS Stream 9 - BaseOS
 baseurl=https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/BaseOS/$basearch/os
--- a/common.yaml
+++ b/common.yaml
@@ -298,9 +298,8 @@ packages-aarch64:
   - irqbalance
 
 packages-s390x:
-  # Required for IBM Secure Execution
-  # Pinned due to https://github.com/openshift/os/issues/1731
-  - s390utils-base-2.33.1-2.el9
+  # Required genprotimg for IBM Secure Execution
+  - s390utils-base
 
 remove-from-packages:
   - - filesystem
--- a/manifest-rhel-9.6.yaml
+++ b/manifest-rhel-9.6.yaml
@@ -12,6 +12,7 @@ variables:
 # Include manifests common to all RHEL and CentOS Stream versions
 include:
   - common.yaml
+  - overrides-rhel-9.6.yaml
 
 repos:
   - rhel-9.6-baseos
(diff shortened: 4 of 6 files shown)


marmijo (approved):
LGTM

jlebon:
Hmm, we already have the c9s mirror repos:
https://github.com/openshift/os/blob/master/c9s-mirror.repo

Seems cleaner to just enable those repos instead?
/hold

dustymabe:
I guess what isn't clear to me is why we would ever use non mirror repos if the mirror ones have a clear advantage..

The repo file states `They are slower to update`, but how much slower are we talking?

jlebon:
> I guess what isn't clear to me is why we would ever use non mirror repos if the mirror ones have a clear advantage..
> 
> The repo file states `They are slower to update`, but how much slower are we talking?

I'm not sure TBH, but enough that we've had at least one instance in the past where we wanted new content faster (https://github.com/openshift/os/commit/bfe3581b91def4df8b1b838415435333272d12bb).

dustymabe:
> I'm not sure TBH, but enough that we've had at least one instance in the past where we wanted new content faster ([bfe3581](https://github.com/openshift/os/commit/bfe3581b91def4df8b1b838415435333272d12bb)).

OK. I'll just switch them back to the mirrors now. I think we should default to this unless we find real need in the future. We don't really do this in FCOS, but I guess we also can use overrides in `coreos-pool` if we had the specific case there too.

dustymabe:
Honestly the more I think about it the more I really don't like having two versions of the repo in here. I'd rather have a single repo file, but just with a commented out baseurl with an explanation of one versus the other.

WDYT?

jlebon:
I think my ideal is that we only consume from the composes by default because it frames SCOS as another artifact of that compose. E.g. an early iteration of the el-only work actually had code that would make the version ID match the compose ID it's building from. I dropped that, but I find it a useful way to think about it. In practice, yes the package set should be identical once the mirrors are in sync.

This relates to https://github.com/openshift/os/blob/master/overrides-c9s.yaml which references the mirror repos and is normally what we use for c9s overrides. It's just that in this case we also needed it for RHEL, which is why it didn't go in there. I guess that'll happen whenever we're tracking a RHEL minor which hasn't branched off of c9s yet.

It still seems cleaner though to have them be separate repo files and add a rationale in the manifest where we enable one vs the other? Not strongly against your suggestion though.

dustymabe:
updated

marmijo (approved):
/lgtm

Thanks for adding to the repo descriptions!

jlebon:
Discussed this with @dustymabe. The mirrors are currently much older (e.g. Feb 10th vs Feb 19th), so it'd be better to stick with the canonical one. We'll create a RHEL version of the `c9s-overrides.yaml` manifest and hoist the s390utils pin up into those. That way the c9s one can specify its own repo names, and it vaguely resembles the setup in FCOS, where we have separate files for overrides and periodically check them.

marmijo (approved):
I think the commit referenced in the message of https://github.com/openshift/os/pull/1750/commits/ee2dc661f7c9ef84fdbbb692effd6ef3314d7439 might need to be updated, but otherwise LGTM.

dustymabe:
> I think the commit referenced in the message of [ee2dc66](https://github.com/openshift/os/commit/ee2dc661f7c9ef84fdbbb692effd6ef3314d7439) might need to be updated, but otherwise LGTM.

you are right. I'll update the commits in the git messages and push here. Since there is no content change I'll just go ahead and merge since CI passed already.

PR #1753, opened 2025-02-20
NO-JIRA: denylist: enable `ostree.sync` and disable on c10s

HuijingHei:
As https://github.com/coreos/coreos-assembler/pull/4026 fixes the issue on c9s & rhcos9.6


c4rt0 (approved):
/lgtm

PR #1748, opened 2025-02-14
DNM: Remove c9s tests for testing purpose

aaradhak:

PR #1755, opened 2025-02-24
NO-JIRA: denylist: remove c9s fips.* entry

marmijo:
The fips.* tests are passing now in c9s. Tested in: https://github.com/openshift/os/pull/1748.


c4rt0 (approved):
/lgtm

PR #1761, opened 2025-03-03
NO-JIRA: Containerfile: add metadata in last layer of node image

jlebon:
Prior to the layered RHCOS enhancement[[1]], there was tooling in ART and in the RHCOS release browser (notably linked from the release controller) that relied on the commit metadata JSON file published separately to be able to inspect the RPM contents of a build.

With the move to a layered node image, we lose this. The more container-native analogue would be SBOMs, which we should get for free once we move to Konflux. However, until then, let's provide an easy-to- generate alternative that still allows avoiding to download the whole image.

We add a new semantic where the last layer of the node image is actually a metadata-only layer. Then, tooling like `oc image extract` capable of extracting only specific layers from an image can be used to pull it down. E.g.

```
$ oc image extract quay.io/jlebon/rhcos:latest[-1] --file /usr/share/openshift/base/meta.json
$ jq . meta.json
```

Add a label to signal that this image implements this semantic.

Compared to SBOMs, this has the _major_ advantage that it's in-band and implicitly gets preserved as the image is slung around (whereas an attached SBOM may not until the tooling is careful to do so).

But like SBOMs, it has the pitfall that the metadata is only valid for the original node image, and obviously skews once more layering is done. The source of truth for the package list is always the rpmdb.

That said, a client implicitly knows if the metadata no longer matches the node content simply by the fact that it wouldn't be in the last layer.

[1]: https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/split-rhcos-into-layers.md

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -27,7 +27,7 @@
 #   --security-opt label=disable -t localhost/openshift-node-c9s \
 #   src/config
 
-FROM quay.io/openshift-release-dev/ocp-v4.0-art-dev:c9s-coreos
+FROM quay.io/openshift-release-dev/ocp-v4.0-art-dev:c9s-coreos as build
 ARG OPENSHIFT_CI=0
 # Avoid shipping modified .pyc files. Due to https://github.com/ostreedev/ostree/issues/1469,
 # any Python apps that run (e.g. dnf) will cause pyc creation.
@@ -37,3 +37,10 @@ RUN --mount=type=bind,target=/run/src --mount=type=secret,id=yumrepos,target=/et
   /run/src/scripts/apply-manifest /run/src/packages-openshift.yaml && \
   find /usr -name '*.pyc.bak' -exec sh -c 'mv $1 ${1%.bak}' _ {} \; && \
   ostree container commit
+
+FROM build as metadata
+RUN --mount=type=bind,target=/run/src /run/src/scripts/generate-metadata
+
+FROM build
+COPY --from=metadata /usr/share/openshift /usr/share/openshift
+LABEL io.openshift.metalayer=true
--- a/scripts/generate-metadata
+++ b/scripts/generate-metadata
@@ -1,0 +1,56 @@
+#!/usr/bin/python3 -u
+
+"""
+This script generates metadata intended to be stored as a separate layer.
+For now, only the RPM listing is generated. It follows the same schema as the
+`rpmostree.rpmdb.pkglist` commit metadata key in a base rpm-ostree compose.
+This is because it's intended to be consumed by code previously adapted for
+that purpose.
+
+Also, for now only a single `meta.json` file is generated. Obviously more
+metadata of interest can be generated in the future.
+"""
+
+import json
+import os
+import subprocess
+import sys
+
+# The "base" here denotes the fact that this describes the base node image and
+# will inherently not be valid if another layering operation is done on top of
+# this image to install packages. The source of truth for package lists will
+# always be the rpmdb.
+METADATA_FILE = "/usr/share/openshift/base/meta.json"
+
+
+def main():
+    metadata = {
+        # this follows the same schema as rpmostree.rpmdb.pkglist but let's
+        # not actually name it the same because it's not actually rpm-ostree
+        # generating this
+        'rpmdb.pkglist': get_rpmdb_pkglist()
+    }
+    os.makedirs(os.path.dirname(METADATA_FILE), exist_ok=True)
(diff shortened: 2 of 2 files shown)


HuijingHei (approved):
LGTM

HuijingHei:
CI error:
```
Running: dnf install --setopt=tsflags=nodocs --noplugins -y cri-o cri-tools conmon-rs openshift-clients openshift-kubelet openvswitch3.4 NetworkManager-ovs ose-aws-ecr-image-credential-provider ose-azure-acr-image-credential-provider ose-gcp-gcr-image-credential-provider
CentOS Stream 9 - BaseOS                         50 MB/s | 8.7 MB     00:00    
CentOS Stream 9 - AppStream                      95 MB/s |  23 MB     00:00    
CentOS Stream 9 - Extras packages               123 kB/s |  20 kB     00:00    
rhel-9-server-ose                                36 MB/s | 9.1 MB     00:00    
rhel-9-server-ose-ppc64le                       4.5 MB/s | 3.0 MB     00:00    
rhel-9-server-ose-s390x                         5.2 MB/s | 2.4 MB     00:00    
rhel-9-server-ose-aarch64                       9.5 MB/s | 7.2 MB     00:00    
No match for argument: openvswitch3.4
Error: Unable to find a match: openvswitch3.4
```

scripts/generate-metadata
```diff
@@ -0,0 +1,56 @@
...
+                                  encoding='utf-8')
+    rpmdb = []
+    for line in out.splitlines():
+        n, e, v, r, a = line.split()
+        # canonicalize none to 0 to match rpm-ostree semantics
+        if e == '(none)':
+            e = 0
+        rpmdb.append([n, e, v, r, a])
+
+    # sort it by package name to be nice to humans who like JSON
```
dustymabe, line 50: humans who like JSON :)

Containerfile
```diff
@@ -37,3 +37,10 @@ RUN --mount=type=bind,target=/run/src --mount=type=secret,id=yumrepos,target=/et
   /run/src/scripts/apply-manifest /run/src/packages-openshift.yaml && \
   find /usr -name '*.pyc.bak' -exec sh -c 'mv $1 ${1%.bak}' _ {} \; && \
   ostree container commit
+
+FROM build as metadata
+RUN --mount=type=bind,target=/run/src /run/src/scripts/generate-metadata
+
+FROM build
+COPY --from=metadata /usr/share/openshift /usr/share/openshift
+LABEL io.openshift.metalayer=true
```
dustymabe, line 46: I'm slightly confused.. I thought each operation in a Containerfile made a new layer anyway.. so how is what you have here different than not doing a multi-stage build and just doing:

```suggestion
RUN --mount=type=bind,target=/run/src /run/src/scripts/generate-metadata
LABEL io.openshift.metalayer=true
```

jlebon: Running the script itself adds more than just the metadata. It regenerates Python cache, it touches the rpmdb (just querying will change the timestamp), etc... Doing it this way ensures there's only _just_ the metadata in there.

jlebon:
OK, this now requires https://github.com/openshift/release/pull/62376.

jlebon:
e2e-aws failure looks unrelated. Cluster came up fine.
/override ci/prow/e2e-aws

nikita-dubrovskii:
@jlebon 

Yep, already disabled in https://github.com/coreos/coreos-assembler/pull/4035

jlebon:
> @jlebon
> 
> Yep, already disabled in [coreos/coreos-assembler#4035](https://github.com/coreos/coreos-assembler/pull/4035)

These tests are failing on x86_64, not s390x.

scripts/generate-metadata
```diff
@@ -0,0 +1,56 @@
...
+def get_rpmdb_pkglist():
+    QUERYFORMAT = '%{NAME}\t%{EPOCH}\t%{VERSION}\t%{RELEASE}\t%{ARCH}\n'
+    out = subprocess.check_output(['rpm', '-qa', '--queryformat', QUERYFORMAT],
+                                  encoding='utf-8')
+    rpmdb = []
+    for line in out.splitlines():
+        n, e, v, r, a = line.split()
+        # canonicalize none to 0 to match rpm-ostree semantics
+        if e == '(none)':
+            e = 0
```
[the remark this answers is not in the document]
jlebon: Good catch, thanks!
https://github.com/openshift/os/pull/1764

PR #1765, opened 2025-03-11
OCPBUGS-48614: Move ose-aws-ecr-image-credential-provider to CoreOS base

jlebon:
The ose-aws-ecr-image-credential-provider package is used in AWS by the kubelet to get ECR credentials, as described by:

https://kubernetes.io/docs/tasks/administer-cluster/kubelet-credential-provider/

However, it's also currently used at least by ROSA to provide a pull secret to podman during provisioning. This use case then broke when we switched to el-only bootimages and stopped including it.

The long-term goal is to have that functionality in RHEL proper, but at least for 4.19 let's work around this by lowering the package down to the CoreOS layer.

See also: https://issues.redhat.com/browse/RHEL-82921

### DIFF
--- a/common.yaml
+++ b/common.yaml
@@ -276,6 +276,10 @@ packages:
  - dnf
  # https://issues.redhat.com/browse/OCPBUGS-35247
  - subscription-manager
+ # XXX: This should be in packages-openshift.yaml instead. For now,
+ # it's in the base until the equivalent functionality lands in RHEL:
+ # https://issues.redhat.com/browse/RHEL-82921
+ - ose-aws-ecr-image-credential-provider
 
 packages-x86_64:
   # Temporary add of open-vm-tools. Should be removed when containerized
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -13,7 +13,8 @@ packages:
   # The packages below are present in CentOS Stream/RHEL,
   # and depend on one or more of the above.
   - NetworkManager-ovs
-  - ose-aws-ecr-image-credential-provider
+  # XXX: There should also be ose-aws-ecr-image-credential-provider here. See
+  # common.yaml.
   - ose-azure-acr-image-credential-provider
   - ose-gcp-gcr-image-credential-provider
 


jlebon:
CI will require https://github.com/openshift/release/pull/62674.

jlebon:
Yeah, we have known testiso CI issues currently (opened https://github.com/openshift/os/pull/1769).

The OKD jobs are new. Looks like the extensions build is failing on getting kernel-rt from the NFV repo. For now I'll override but cc @Prashanth684 can you take a look?

/override ci/prow/okd-scos-e2e-aws-ovn

The e2e-aws test is failing on... something. Looks like bootstrapping failed but I don't know where to look in the artifacts for the right logs.

I don't think it's the case here, but if e2e-aws requires ose-aws-ecr-image-credential-provider, it's expeted that it'd fail because in CI we only build the node layer, we don't rebuild the CoreOS base. So that layered node image built in CI is missing the package entirely.

Anyway, for now

/override ci/prow/e2e-aws

Prashanth684:
> The OKD jobs are new. Looks like the extensions build is failing on getting kernel-rt from the NFV repo. For now I'll override but cc @Prashanth684 can you take a look?

I just looked at these and this is because the baseos image has a kernel version 570 and the composes have already moved on to 571. I think i need to pull in the mirror repos as well when building the node images. I will open a PR shortly

Prashanth684:
> The e2e-aws test is failing on... something

The e2e test for OKD is failing because the `ose-aws-ecr-image-credential-provider` has been moved out of packages-openshift.yaml. Thus, the image is built without it and kubelet fails with : 

```
Mar 12 21:12:21 ip-10-0-6-206 kubenswrapper[2469]: E0312 21:12:21.734374    2469 kuberuntime_manager.go:276] "Failed to register CRI auth plugins" err="plugin binary executable  did not exist"
```
shouldn't OCP be hitting this as well?

Prashanth684:
> > The e2e-aws test is failing on... something
> 
> The e2e test for OKD is failing because the `ose-aws-ecr-image-credential-provider` has been moved out of packages-openshift.yaml. Thus, the image is built without it and kubelet fails with :
> 
> ```
> Mar 12 21:12:21 ip-10-0-6-206 kubenswrapper[2469]: E0312 21:12:21.734374    2469 kuberuntime_manager.go:276] "Failed to register CRI auth plugins" err="plugin binary executable  did not exist"
> ```
> 
> shouldn't OCP be hitting this as well?

actually the aws job is failing for the same reason in this PR: https://github.com/openshift/os/pull/1770

jlebon:
> actually the aws job is failing for the same reason in this PR: #1770 . So once the base image is built with the package, it shouldn't fail anymore?

Right, I mentioned this in https://github.com/openshift/os/pull/1765#issuecomment-2718606091.

Issue #1768, opened 2025-03-12
Test `pxe-offline-install.rootfs-appended.bios` perma-failing in Prow


jlebon:
For some reason, this fails in Prow but not in the pipeline.
Looks like ignition-fetch times out unable to actually get to the pxe-live.ign hosted in the TFTP qemu server:

```
[  132.272322] ignition[689]: GET http://192.168.76.2:43841/pxe-live.ign: attempt #28
[  132.278949] ignition[689]: GET error: Get "http://192.168.76.2:43841/pxe-live.ign": dial tcp 192.168.76.2:43841: connect: network is unreachable
```

PR #1769, opened 2025-03-12
NO-JIRA: ci/prow-entrypoint: skip pxe-offline-install.rootfs-appended.bios for…

jlebon:
… now

This is perma-failing in Prow right now, so just skip over it for now.

See: https://github.com/openshift/os/issues/1768


marmijo (approved):
/lgtm

PR #1770, opened 2025-03-12
NO-JIRA: ci/get-ocp-repo.sh: pull in mirror repos when building scos

Prashanth684:
Recently, building scos extensions hit this error:
```
error: Packages not found: kernel-rt-core-5.14.0-570.el9, kernel-rt-devel-5.14.0-570.el9, kernel-rt-kvm-5.14.0-570.el9, kernel-rt-modules-5.14.0-570.el9, kernel-rt-modules-extra-5.14.0-570.el9
```

This is because the baseos image was built with the 570 version of the kernel, but the composes had already moved on to the 571 version. While pulling the mirror repos fixes this for now, it would be ideal to always build from mirror repos as composes seem to have a faster cadence. We should probably also have the node image building be triggered at the same time the base images are being built to avoid these races.

### DIFF
--- a/ci/get-ocp-repo.sh
+++ b/ci/get-ocp-repo.sh
@@ -194,7 +194,11 @@ if [ "$osname" = scos ]; then
     if [ -n "$ocp_manifest" ]; then
         workdir=$(dirname "$ocp_manifest")
     fi
+    # pull in the mirror repo as well in case there are newer versions in the composes
+    # and we require older versions - this happens because we build the node images async
+    # and the composes move fast.
     cat "$workdir/c9s.repo" >> "$repo_path.tmp"
+    cat "$workdir/c9s-mirror.repo" >> "$repo_path.tmp"
     mv "$repo_path.tmp" "$repo_path"
     create_gpg_keys
 fi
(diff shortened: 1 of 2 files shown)


overrides-c9s.yaml
```diff
@@ -6,6 +6,10 @@
 repos:
  - c9s-baseos-mirror
  - c9s-appstream-mirror
+# rt-mirror is needed while building scos node images through CI in case the composes have already moved kernel version
+# this is true for any othe package as well - so we will need to find a better solution for this [TODO]
+ - c9s-rt-mirror
+ - c9s-nfv-mirror
```
Prashanth684, line 12: @jlebon is there a better alternative to keep the building of the node image in sync with the bootimage? how  is it done for RHCOS?

Prashanth684: should SCOS just build from the mirror rather than the composes?

jlebon: It's awkward but currently, the base image is built with the pungi compose _and_ mirror (see convo in https://github.com/openshift/os/pull/1765), so yeah we should just match this for the extensions too.

But these should be in `extensions-okd-c9s.yaml`, not here. (And only in the `kernel-rt` extension since the repo doesn't exist on every arch.)

Prashanth684: makes sense. moved it to the extensions file.

PR #1776, opened 2025-03-19
COS-3126: overlay.d: replace 25rhcos-azure-udev

marmijo:
The current `25rhcos-azure-udev` overlay has been obsolete since we moved to EL9. Let's drop it and replace it with another overlay with the same name.

The new `25rhcos-azure-udev` overlay includes udev rules that support Azure Managed NVMe disks. These udev rules are shipped as part of the `azure-vm-utils` package[1], but that package wont be included until EL10[2]. We can drop this overlay when we move to EL10, provided the package is included at that time.

[1] https://github.com/Azure/azure-vm-utils/blob/9c596916b6774f24420dac0ee7a72a6c9ddb5060/udev/80-azure-disk.rules
[2] https://issues.redhat.com/browse/RHEL-73904


manifest-el9-shared.yaml
```diff
@@ -8,6 +8,9 @@ remove-from-packages:
   - - zram-generator
     - "/usr/lib/systemd/zram-generator.conf"
 
+ostree-layers:
+  - overlay/25rhcos-azure-udev
```
dustymabe, line 12: Having the same comment from below here would be nice:

```suggestion
  # The azure-vm-utils package provides these udev rules[1], but it wont be added
  # until EL10 [1]. This can be dropped when moving to EL10, provided the
  # package is included at that time.
  - overlay/25rhcos-azure-udev
```

overlay.d/25rhcos-azure-udev/usr/lib/udev/rules.d/80-azure-disk.rules
```diff
@@ -0,0 +1,50 @@
...
+LABEL="azure_disk_nvme_direct_v2"
+ATTRS{nsid}=="?*", ENV{AZURE_DISK_TYPE}="local", ENV{AZURE_DISK_SERIAL}="$env{ID_SERIAL_SHORT}"
+GOTO="azure_disk_nvme_id"
+
+LABEL="azure_disk_nvme_remote_v1"
+ATTRS{nsid}=="1", ENV{AZURE_DISK_TYPE}="os", GOTO="azure_disk_nvme_id"
+ATTRS{nsid}=="?*", ENV{AZURE_DISK_TYPE}="data", PROGRAM="/bin/sh -ec 'echo $$((%s{nsid}-2))'", ENV{AZURE_DISK_LUN}="$result"
+
+LABEL="azure_disk_nvme_id"
+ATTRS{nsid}=="?*", IMPORT{program}="/usr/sbin/azure-nvme-id --udev"
```
dustymabe, line 35: There is a diff from upstream (fedora RPM) here:

```patch 
--- /usr/lib/udev/rules.d/80-azure-disk.rules   1970-01-01 00:00:00.000000000 +0000
+++ /var/mnt/workdir/80-azure-disk.rules        2025-03-20 16:12:18.020106392 +0000
@@ -32,7 +32,7 @@
 ATTRS{nsid}=="?*", ENV{AZURE_DISK_TYPE}="data", PROGRAM="/bin/sh -ec 'echo $$((%s{nsid}-2))'", ENV{AZURE_DISK_LUN}="$result"
 
 LABEL="azure_disk_nvme_id"
-ATTRS{nsid}=="?*", IMPORT{program}="/usr/bin/azure-nvme-id --udev"
+ATTRS{nsid}=="?*", IMPORT{program}="/usr/sbin/azure-nvme-id --udev"
 
 LABEL="azure_disk_symlink"
 # systemd v254 ships an updated 60-persistent-storage.rules that would allow
```

Though.. this really highlights a problem here that I didn't understand before. Previously the udev rules we provided were just rules and didn't require external software.

Here these rules are running `/usr/sbin/azure-nvme-id`, but we aren't shipping that here (because we aren't shipping an RPM).

I think this may be a blocker for us for el9.

marmijo: This is interesting because these rules are directly copied from the `azure-vm-utils` github repo, so there's a diff between that file and the udev rules shipped in the RPM.

as for the **azure-nvme-id** binary: The tool's [description](https://github.com/Azure/azure-vm-utils/blob/9c596916b6774f24420dac0ee7a72a6c9ddb5060/doc/azure-nvme-id.md#description) says:
```
azure-nvme-id provides identification metadata in the response to Identify Namespace command
for some models of NVMe devices. This is found in vendor-specific (vs) field which contains various
identification details with a comma-separated, key=value format.
```

I'm not sure what, within Azure, would rely on this vendor specific data, but there could be unexpected behavior if the binary is missing. However, the original bug was confirmed to be resolved by adding these udev rules without installing the whole RPM, and using them in an RHCOS Azure instance seems to work as expected.

dustymabe: > This is interesting because these rules are directly copied from the `azure-vm-utils` github repo, so there's a diff between that file and the udev rules shipped in the RPM.

yeah. In rawhide there was a usr/bin usr/sbin migration thing and that probably made its way into f41/f42 RPMs too so that probably is what this is about. It probably shouldn't matter too much I don't guess.





> I'm not sure what, within Azure, would rely on this vendor specific data, but there could be unexpected behavior if the binary is missing. However, the original bug was confirmed to be resolved by adding these udev rules without installing the whole RPM, and using them in an RHCOS Azure instance seems to work as expected.

So I think you're telling me the rules we need aren't this one that runs this binary?

I guess it's OK then, but I'd almost rather remove the lines that rely on the binary (or comment them out and add a comment why).

marmijo: Yes, the other rules still work and set everything up correctly even if the binary is missing. 

```suggestion
#ATTRS{nsid}=="?*", IMPORT{program}="/usr/sbin/azure-nvme-id --udev"
```
I commented out the line, but left `LABEL="azure_disk_nvme_id"` intact so the other rules with `GOTO="azure_disk_nvme_id"` continue to function properly. The behavior was the same with or without that line commented out:

- `AZURE_DISK_LUN` is set to a value on NVMe data disks
- data and os NVMe disks show `AZURE_DISK_TYPE=data` and `AZURE_DISK_TYPE=os` respectively 
- NVMe disks and their partitions show up in `/dev/disk/azure` and `/dev/disk/azure/data/by-lun` as expected

This means the udev rule that calls the binary must be failing silently and isn't necessary for this specific bug fix.

dustymabe: I'm happy to have it commented out here in the code with a comment that says why we commented it out and then we should be able to move forward I think.

marmijo: done!

travier:
That looks correct, but that seems risky, as we don't know what would need those attributes and how important they are.

dustymabe:
> That looks correct, but that seems risky, as we don't know what would need those attributes and how important they are.

Yeah. I agree it seems risky. But really this is our only option if we want this now.

We should probably name it as a limitation and then move forward based on the feedback.

PR #1778, opened 2025-03-24
packages-openshift.yaml: Force overwrite if the symlink already exists

ravanelli:
- Now with the extensions container build, we are already using the node image as base, in this case the link already exists and fails, force the overwrite instead of failing.

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -42,8 +42,8 @@ postprocess:
   - |
     #!/usr/bin/bash
     set -euo pipefail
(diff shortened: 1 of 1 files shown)


ravanelli:
Ideally I think we could remove the duplication that is done in the image node and here.  Not sure if it is an option

jlebon:
Hmm, not sure why this is needed. The scripts here shouldn't run in the extensions build. Are you sure you're pointing `podman build` at `extensions/Dockerfile`?

ravanelli:
You are right! I probably mistaken it due the layer packages-openshift.yaml

PR #1780, opened 2025-03-28
NO-JIRA: Switch layered build to `treefile-apply`, drain `get-ocp-repo.sh`

jlebon:
See individual commit messages.

PR #1788, opened 2025-04-03
NO-JIRA: Prep patches for #1780

jlebon:
These are split out of #1780 because they need to land first and be in the base images.


jlebon:
Let's just skip CI for this? I'll trigger the pipeline right after so we get fresh base images and it'll test it there. (And then again in earnest in #1780.)

PR #1789, opened 2025-04-03
c9s.repo: hardcode x86_64 in baseurl

jlebon:
As mentioned in the comment, that repo is just noarch packages containing GPG keys and repo definition files but for some reason those noarch packages are only available in the x86_64 one.

So just hardcode x86_64 to make this work on all arches.

Fixes 4fd59d0 ("manifest-c9s: add SIGs GPG keys").


dustymabe (approved):
/lgtm

PR #1791, opened 2025-04-05
NO-JIRA: extensions-ocp-rhel-9.6.yaml: add back in rhel-9.6-server-ose-4.19 repo

dustymabe:
This was inadvertently dropped (I think) in d01f69a. Without it crun-wasm and kata-containers packages can't be found so let's add back rhel-9.6-server-ose-4.19 to the repo list.


dustymabe:
Also a second commit in there to fixup/add some comments to the files.

dustymabe:
I actually added another commit here to simplify the global repo list rule in the extensions file and move more repos up to the global list.

PR #1796, opened 2025-04-09
OCPEDGE-1704: [TNF] Added symlink to allow for custom pacemaker resource agent in two-node OCP variant

jaypoulz:
Two-Node OpenShift with fencing is a variant of OpenShift that will be DevPreview in 4.19.
It adds a special executable to /usr/lib/ocf/resource.d/ocp-tnf/ that is not yet in the CoreOS extension for `two-node-ha`.

We are working to address this missing component ASAP, but in the meantime, this symlink allows us to enable this component via a MachineConfig.

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -189,6 +189,15 @@ postprocess:
     ---
     EOF
 
+  # Inject symlink for Two Node OpenShift with Fencing (TNF)
+  # This is a workaround to allow pacemaker to use a pre-release resource agent for etcd
+  # This should be removed after https://issues.redhat.com/browse/OCPEDGE-1706 is completed
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    mkdir -p /usr/lib/ocf/resource.d/ocp-tnf
+    ln -s /usr/local/lib/ocf/resource.d/ocp-tnf/podman-etcd /usr/lib/ocf/resource.d/ocp-tnf/podman-etcd
+
   - |
     #!/usr/bin/env bash
     set -xeo pipefail


jaypoulz:
Kola tests were failing because there was nothing at the destination of the symlink. I've added an empty file to hopefully please the tests.

packages-openshift.yaml
```diff
@@ -189,6 +189,18 @@ postprocess:
...
+  # This is a workaround to allow pacemaker to use a pre-release resource agent for etcd
+  # This should be removed after https://issues.redhat.com/browse/OCPEDGE-1706 is completed
+  # We touch an empty file in the target location to pass kola tests
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    mkdir -p /usr/lib/ocf/resource.d/ocp-tnf
+    ln -s /usr/local/lib/ocf/resource.d/ocp-tnf/podman-etcd /usr/lib/ocf/resource.d/ocp-tnf/podman-etcd
+    mkdir -p /var/usrlocal/lib/ocf/resource.d/ocp-tnf
+    touch /usr/local/lib/ocf/resource.d/ocp-tnf/podman-etcd
```
travier, line 202: I don't think you need those. They will be created on demand when you install the files there.

To workaround the broken symlink kola check, you can skip by adding an entry in https://github.com/openshift/os/blob/master/kola-denylist.yaml instead.

This should check if the package is fixed and fail if it is.

jaypoulz: With regards to ensuring the symlink is not present once the package is fixed, I spoke to the team about this and our concern is that this would break our installations (and upgrades) until we patch CEO. We intentionally installed this symlink at a non-conflicting location from the one that will be used when the package is fixed. Once the underlying fix is in the RPM, we need to update the our code in CEO to ensure that upgrades and installs all reference the new version. Otherwise, we risk having the code point to where the symlink no longer is.

The advantage of doing it this way is that we can ensure that upgrades always pair a compatible CEO and RHCOS.

jaypoulz: If that's really unacceptable to the team, can land this alternative for now and work on a bug fix for CEO.
https://github.com/openshift/os/pull/1796

kola-denylist.yaml
```diff
@@ -117,3 +117,11 @@
...
     - centos-10
     - rhel-10.1
+
+# We are currently adding a dead symlink to allow MCO to install a pacemaker
+# resource agent in Two Node OpenShift with Fencing (TNF). This will be removed
+# once the underlying agent is shipped as part of the `two-node-ha` extension
+# and the TNF controller in etcd is updated to initialize pacemaker clusters
+# via the version shipped in the extension.
+- pattern: ext.config.shared.files.validate-symlinks
+  tracker: https://issues.redhat.com/browse/OCPEDGE-1706
```
jlebon, line 127: https://github.com/openshift/os/pull/1801#discussion_r2047383088 still applies here.

jaypoulz: Filed https://issues.redhat.com/browse/OCPEDGE-1752 to track that follow-up

PR #1797, opened 2025-04-10
NO-JIRA: c10s.repo: use x86_64 extras-common URL for all arch

jcapiitao:
Currently, CentOS Stream does not ship extras-common repo for s390x [1]. 
This prevents us to build c10s against this arch. I basically replicate the same hack config as we do for c9s. 
Note that this extras-common repo only provides noarch RPMs, at least the one we pull (i.e centos-release-cloud-common, centos-release-nfv-common and centos-release-virt-common) are noarch.

[1] https://pagure.io/centos-infra/issue/1635


dustymabe (approved):
/lgtm

PR #1799, opened 2025-04-11
NO-JIRA: denylist: ext.config.rpm-ostree.replace-rt-kernel for c10s on ppc64le

jcapiitao:
The `ext.config.rpm-ostree.replace-rt-kernel` test is failing [1] when downgrading to previous kernel. 
In order to have the first `c10s` build produced by the pipeline, we can ignore this failure for now.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1923

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -109,3 +109,11 @@
   tracker: https://issues.redhat.com/browse/RHEL-86436
   osversion:
     - rhel-10.1
+
+- pattern: ext.config.rpm-ostree.replace-rt-kernel
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1923
+  arches:
+    - ppc64le
+  osversion:
+    - centos-10
+    - rhel-10.1


kola-denylist.yaml
```diff
@@ -109,3 +109,10 @@
   tracker: https://issues.redhat.com/browse/RHEL-86436
   osversion:
     - rhel-10.1
+
+- pattern: ext.config.rpm-ostree.replace-rt-kernel
+  tracker: https://github.com/coreos/fedora-coreos-tracker/issues/1923
+  arches:
+    - ppc64le
+  osversion:
+    - centos-10
```
dustymabe, line 118: should we consider adding `rhel-10.1` here too even though we don't know for sure?

dustymabe (approved):
/lgtm

kola-denylist.yaml
jcapiitao: right, let's add it just in case

dustymabe (approved):
/lgtm

PR #1801, opened 2025-04-16
OCPEDGE-1704: [TNF] Added symlink to allow for custom pacemaker resource agent in two-node OCP variant

jaypoulz:
Two-Node OpenShift with fencing is a variant of OpenShift that will be DevPreview in 4.19.
It adds a special executable to /usr/lib/ocf/resource.d/ocp-tnf/ that is not yet in the CoreOS extension for two-node-ha.

We are working to address this missing component ASAP, but in the meantime, this symlink allows us to enable this component via a MachineConfig.

--
This PR is an alternative to https://github.com/openshift/os/pull/1796 that ensures the symlink is not present once desired file is in resource-agents.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -117,3 +117,11 @@
   osversion:
     - centos-10
     - rhel-10.1
+
+# We are currently adding a dead symlink to allow MCO to install a pacemaker
+# resource agent in Two Node OpenShift with Fencing (TNF). This will be removed
+# once the underlying agent is shipped as part of the `two-node-ha` extension
+# and the TNF controller in etcd is updated to initialize pacemaker clusters
+# via the version shipped in the extension.
+- pattern: ext.config.shared.files.validate-symlinks
+  tracker: https://issues.redhat.com/browse/OCPEDGE-1706
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -189,6 +189,32 @@ postprocess:
     ---
     EOF
 
+  # Inject symlink for Two Node OpenShift with Fencing (TNF)
+  # This is a workaround to allow pacemaker to use a pre-release resource agent for etcd
+  # This should be removed after https://issues.redhat.com/browse/OCPEDGE-1706 is completed
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    rpm-ostree install --apply-live dnf-utils -y
+    set +e
+    python -m dnf.cli.main > /dev/null
+    if [ $? -ne 0 ]; then
+      exit 1
+    fi
+    python -m dnf.cli.main repoquery -l resource-agents | grep /usr/lib/ocf/resource.d/heartbeat/podman-etcd
+    if [ $? -eq 0 ]; then
+      rpm-ostree apply-live remove dnf-utils || dnf remove dnf-utils -y
+      dnf clean all
+      rpm-ostree cleanup -bpm
+      exit 0
+    fi
+    rpm-ostree apply-live remove dnf-utils || dnf remove dnf-utils -y
+    dnf clean all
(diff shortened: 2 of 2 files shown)


packages-openshift.yaml
```diff
@@ -189,6 +189,32 @@ postprocess:
...
+  - |
+    #!/usr/bin/env bash
+    set -xeuo pipefail
+    rpm-ostree install --apply-live dnf-utils -y
+    set +e
+    python -m dnf.cli.main > /dev/null
+    if [ $? -ne 0 ]; then
+      exit 1
+    fi
+    python -m dnf.cli.main repoquery -l resource-agents | grep /usr/lib/ocf/resource.d/heartbeat/podman-etcd
```
travier, line 204: You should be able to do that with `rpm` directly?

jaypoulz: Maybe it's a knowledge gap thing, but I don't know where the RPM lives.
I know `rpm -q` can help for local RPMs, but how do I get the correct RPM in the first place?

The easiest way I could think to do that is to ask the tool for installing RPMs to look it up.

kola-denylist.yaml
```diff
@@ -117,3 +117,11 @@
...
     - centos-10
     - rhel-10.1
+
+# We are currently adding a dead symlink to allow MCO to install a pacemaker
+# resource agent in Two Node OpenShift with Fencing (TNF). This will be removed
+# once the underlying agent is shipped as part of the `two-node-ha` extension
+# and the TNF controller in etcd is updated to initialize pacemaker clusters
+# via the version shipped in the extension.
+- pattern: ext.config.shared.files.validate-symlinks
+  tracker: https://issues.redhat.com/browse/OCPEDGE-1706
```
jlebon, line 127: Hmm, denylisting that test for an unknown amount of time doesn't feel great.

I think... we can get this in as is for now, but can you also open a PR to add it to https://github.com/coreos/fedora-coreos-config/blob/350e4d59c2ca8dd84419982eb51910961b095f86/tests/kola/files/validate-symlinks#L39 and then we can drop this denylist entry in the next fedora-coreos-config submodule bump.

packages-openshift.yaml
jlebon: Definitely don't want to install packages here.

Hmm, isn't this whole block basically

```bash
if ! test -f /usr/lib/ocf/resource.d/heartbeat/podman-etcd; then
    mkdir -p /usr/lib/ocf/resource.d/ocp-tnf
    ln -s /usr/local/lib/ocf/resource.d/ocp-tnf/podman-etcd /usr/lib/ocf/resource.d/ocp-tnf/podman-etcd
fi
```

?

But again, I would still prefer we add an e.g. `else` block there and `exit 1` so we know when to pull this out.

jlebon:
> But again, I would still prefer we add an e.g. else block there and exit 1 so we know when to pull this out.

Sorry, I wrote this before seeing https://github.com/openshift/os/pull/1796#discussion_r2045673921 which touches on this.

Hmm OK, I think https://github.com/openshift/os/pull/1796 is fine then. We just need to make sure we clean this up eventually.

kola-denylist.yaml
jaypoulz: Filed https://issues.redhat.com/browse/OCPEDGE-1752 to track updating the kola test. I can follow up with a fix for that and this repo once that's checked-in.

packages-openshift.yaml
jaypoulz: That file won't be present until MCO enables the `two-node-ha` extension, so it would never be there by default in the rhcos base image.

jaypoulz:
Closing this PR as obsolete in favor of https://github.com/openshift/os/pull/1796

PR #1759, opened 2025-03-01
NO-JIRA: Bump fedora-coreos-config

coreosbot-releng:
Created by [GitHub workflow](https://github.com/coreos/fedora-coreos-config/actions/workflows/openshift-os.yml) ([source](https://github.com/coreos/fedora-coreos-config/blob/testing-devel/.github/workflows/openshift-os.yml)).

```
Aashish Radhakrishnan (3):
      workflows/remove-graduated-overrides: Install python3-libdnf5
      denylist: add snooze for ext.config.kdump.crash & kdump.crash.ssh
      denylist: add snooze for kdump.crash.nfs

Adam Piasecki (3):
      tests/aws: Modify xen-assert to find xen entry in hypervisor string
      denylist: snooze root-reprovision on rawhide for ppc64le
      denylist: snooze coreos.boot-mirror for x86_64

Dusty Mabe (25):
      tests/manual: trim saved state in coreos-builds-bisect.py
      denylist: skip console warnings on ppc
      overlay.d: drop 40grub overlay
      overrides: fast-track ignition-2.21.0-2.fc41
      buildroot: workaround rpmbuild bug
      buildroot: better workaround for rpm-build bug
      denylist: add denial for toolbox on rawhide
      tests/upgrade: always wait for fix-selinux-labels
      tests/upgrade: rework booted deployment processing
      tests/upgrade: adjust for OCI based updates
      tests/upgrade: use a variable for architecture
      tests/upgrade: fix rebase to container image
      workflows/add-overrides: install python3-libdnf5
      tests: drop some cruft from the net behavior change test
      denylist: extend snooze for toolbox
      Move to Fedora Linux 42
      image.yaml: switch to EROFS; deploying via container
      drop as much dead code/manifests as possible
      overlay.d: Drop 07fix-selinux-labels overlay
      overlay/15fcos: drop cgroups and wifi checks
      overrides: fast-track moby-engine, containerd
      inline the 16disable-zincati overlay
      manifests/user-experience.yaml: include makedumpfile kdumpctl
      some riscv enablement
      denylist: deny ext.config.multipath.resilient on s390x

Etienne Champetier (1):
      overlay.d/40grub: remove 70_coreos-user.cfg

Fifo Phonics (1):
      Add docker swarm overlay network auto-bridge test

Huijing Hei (1):
      tests/networking: update `nm-ifcfg-rh-plugin` and `team-dhcp-via-ignition` working on el9

Jeremy Poulin (1):
[...]

PR #1571, opened 2024-08-15
COS-2758: ITUP Cluster Migration - Use HTTPS where possible

marmijo:
These commits are needed in order to migrate the RHCOS build pipeline to the new ITUP cluster. They deal with modifying our configuration to only use HTTPS where possible since the ITUP cluster restricts the use of HTTP on port 80.

@aaradhak and I have been using these commits in a fork of this repo for testing on the cluster. We've worked out most of the issues, so let's merge them now.

```
commit 8301c67176e71ff2c6c5236109c89a149422a592
Author: Michael Armijo <marmijo@redhat.com>
Date:   Thu Jul 18 17:59:22 2024 -0600

    extensions/Dockerfile: use the FCOS defined fedora.repo to set up container
    
    Use the fedora.repo file defined in fedora-coreos-config to set up the
    container. This will force packages to be downloaded from
    dl.fedoraproject.org, as specified in the FCOS file. The ITUP cluster,
    being used by the RHCOS pipeline, requires all outbound connections
    to be specified in a Firewall Egress file, and this will ensure the
    same connection will always be used.

commit 761db82968b0caf82ce5cf7761c9bef970097237
Author: Michael Armijo <marmijo@redhat.com>
Date:   Mon Jun 24 14:19:27 2024 -0600

    c9s-mirror.repo: use https for repo baseurl
    
    We use https in the main c9s.repo file, so let's use it here
    too to be consistent. Also, the RHCOS pipeline is migrating
    to a new ITUP cluster which is restricting port 80, so
    lets just use HTTPS on port 443 which is already open.
    
    Co-authored-by: Aashish Radhakrishnan <aaradhak@redhat.com>

commit 9d4fa02d9664386f9fa137221886b8321fc1c2d9
Author: Michael Armijo <marmijo@redhat.com>
Date:   Thu Jun 20 13:28:59 2024 -0600

    tests/replace-rt-kernel: use https for baseurl for repos
    
    We use https in the main c9s.repo file, so let's use it here
    too to be consistent. Also, the RHCOS pipeline is migrating
    to a new ITUP cluster which is restricting port 80, so
    lets just use HTTPS on port 443 which is already open.
    
    Co-authored-by: Aashish Radhakrishnan <aaradhak@redhat.com>
```

PR #1808, opened 2025-04-28
extensions/Dockerfile: Update to Fedora 42

aaradhak:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1851

### DIFF
--- a/extensions/Dockerfile
+++ b/extensions/Dockerfile
@@ -12,7 +12,7 @@ RUN --mount=type=secret,id=yumrepos,target=/os/secret.repo extensions/build.sh
 ## Creates the repo metadata for the extensions.
 ## This uses Fedora as a lowest-common-denominator because it will work on
 ## current p8/s390x. See https://github.com/openshift/os/issues/1000
-FROM quay.io/fedora/fedora:41 as builder
+FROM quay.io/fedora/fedora:42 as builder
 COPY --from=os /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
 RUN rm -f /etc/yum.repos.d/*.repo \
 && curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora-archive.repo -o /etc/yum.repos.d/fedora-archive.repo


marmijo (approved):
LGTM

extensions/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN --mount=type=secret,id=yumrepos,target=/os/secret.repo extensions/build.sh
 ## Creates the repo metadata for the extensions.
 ## This uses Fedora as a lowest-common-denominator because it will work on
 ## current p8/s390x. See https://github.com/openshift/os/issues/1000
-FROM quay.io/fedora/fedora:41 as builder
+FROM quay.io/fedora/fedora:42 as builder
```
dustymabe, line 15: @jlebon @travier - should we consider switching back to centos stream9 here?

extensions/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN --mount=type=secret,id=yumrepos,target=/os/secret.repo extensions/build.sh
...
 COPY --from=os /usr/share/rpm-ostree/extensions/ /usr/share/rpm-ostree/extensions/
 RUN rm -f /etc/yum.repos.d/*.repo \
 && curl -L https://raw.githubusercontent.com/coreos/fedora-coreos-config/testing-devel/fedora-archive.repo -o /etc/yum.repos.d/fedora-archive.repo
```
dustymabe, line 18: I don't understand why we are pulling in the archive repo here. This should probably be dropped.

marmijo: We'll still need to keep this while we are building on the ITUP cluster: https://github.com/openshift/os/pull/1571/commits/8301c67176e71ff2c6c5236109c89a149422a592

dustymabe: https://github.com/openshift/os/pull/1603#issuecomment-2840071802

extensions/Dockerfile
travier: Should be OK if we only build RHEL9 now. I think that's the case but I've not 100% verified.

dustymabe:
added a commit to https://github.com/openshift/os/pull/1759 that switches to stream9 here.

PR #1810, opened 2025-05-01

coreosbot-releng:

```
Dusty Mabe (2):
      denylist: drop toolbox test denial on rawhide
      denylist: deny ext.config.multipath.resilient on all arches

Michael Armijo (1):
      denylist: extend snooze for kdump.crash*
```


dustymabe:
These PRs in the future will be opened against https://github.com/coreos/rhel-coreos-config

Issue #1775, opened 2025-03-19
Move layered node image definition to new openshift/node repo


jlebon:
Currently, this repo contains both the definition of the base RHCOS bootable container (which includes `image.yaml`, used for disk image creation) and the layered node image (`Containerfile`).

Those two are attached to different versioning. E.g. the layered node image content should keep following OCP branching, but that doesn't make sense for the RHEL-versioned base image. Additionally, having both defined in the same repo is confusing because it's not clear which file pertains to which and what exactly CI is testing.

See also: https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/split-rhcos-into-layers.md#re-organize-repos

Here's a starting proposal:
- We create a new openshift/node repo and move everything related to building the node image there. We can then also start adding more OCP-specific things, like some of the MCO static templates. That repo follows OCP branching.
- We stop branching this repo for OCP versions. We start branching this repo for RHEL EUS versions. Once dual stream happens, one branch would capture definitions for both RHEL 9.x and RHEL 10.y versions that were released together on the OCP side.
- We rename this repo to openshift/coreos? Optional, but I think it'd be clearer than the generic "os" name.

One thing I had considered was to have the RHCOS/SCOS base definition be defined in Gitlab instead, closer to rhel-bootc/centos-bootc. But I think the pull from OpenShift CI and its integration with GitHub is too strong to ignore. A lot of CI happens at the node layer, but I want us to also be able to easily run cluster tests when the base layer changes.

dustymabe:
I'm generally in favor, but do have some comments.



> * We create a new openshift/node repo and move everything related to building the node image there. We can then also start adding more OCP-specific things, like some of the MCO static templates. That repo follows OCP branching.

SGTM

> * We stop branching this repo for OCP versions. We start branching this repo for RHEL EUS versions. Once dual stream happens, one branch would capture definitions for both RHEL 9.x and RHEL 10.y versions that were released together on the OCP side.

This one is less clear. We branch for RHEL EUS versions but also define el9 and el10 (and eventually el11) together in the latest development branch?

> 
> * We rename this repo to openshift/coreos? Optional, but I think it'd be clearer than the generic "os" name.

meh.. maybe, but I would put this low on the priority list especially if there is a lot of work that has to be done in various places to support it.

dustymabe:
[...]

PR #1811, opened 2025-05-01
tree: delete base compose files, rework README

jlebon:
This is part of https://github.com/openshift/os/issues/1775. See individual commit messages.


jlebon:
OK, this should be ready to go now!

jlebon:
OK right, this requires https://github.com/openshift/release/pull/64590.

Note I do delete the `ci/prow/validate` entrypoint here (since I nuke the entire `ci/prow-entrypoint.sh` script. I think it totally makes sense to have a validation check here, but we should think about how to share it with rhel-coreos-config instead of duplicating it here. Probably cleanest is in rhel-coreos-config, we split it out of `ci/prow-entrypoint.sh` into its own `ci/validate.sh` and here, we just `curl` that down and run it?

Anyway, for now I've dropped it but we can re-add it soon.

jlebon:
OKD AWS e2e test failure seems unrelated. Node did come up.
So this just needs a `/lgtm`!

dustymabe (approved):
/lgtm

PR #1813, opened 2025-05-02
NO-JIRA: tests: fix RHEL 10.1 handling in kernel replace test

dustymabe:
The major=$(. /usr/lib/os-release && echo "${CPE_NAME}" | grep -Eo '[0-9]{1,2}') was yielding '10 1' so it was breaking this test. Let's use the new match_maj_ver function that Huijing recently added.


aaradhak (approved):
/lgtm

PR #1816, opened 2025-05-09
NO-JIRA: README.md: add Mermaid diagram

jlebon:
This is a simple diagram of the node image and what it's made up of.


jlebon:
image

Could probably be made cleaner with more Mermaid-foo, but eh... it gets the point across.

jbtrystram:
Nice !

PR #1819, opened 2025-05-13
NO-JIRA: c9s/c10s repos: use basearch var for extras repo

jcapiitao:
estras-common repo was missing for s390x arch [1], but it's now fixed.

[1] https://pagure.io/centos-infra/issue/1635


jbtrystram (approved):
Nice ! 
/lgtm

PR #1824, opened 2025-06-23
COS-3294: extensions/rhel-10.1: drop kernel-rt-kvm package from RHEL 10 builds

marmijo:
The kernel-rt-kvm package is removed in RHEL 10. Going forward, KVM modules will no longer be shipped in a separate package for RT kernels.

See: https://gitlab.com/cki-project/kernel-ark/-/merge_requests/3837


dustymabe (approved):
/lgtm

PR #1825, opened 2025-06-23
COS-3294: extensions/rhel-10.1: skip crun-wasm dependency due to missing repo

marmijo:
The el10 `crun-wasm` package isn't available yet because the `server-ose-4.20` repo doesn't exist for RHEL 10. Let's comment it out until the repo exists.

### DIFF
--- a/extensions/rhel-10.1.yaml
+++ b/extensions/rhel-10.1.yaml
@@ -23,13 +23,14 @@ repos:
   # - rhel-10.1-fast-datapath
 
 extensions:
-  # https://issues.redhat.com/browse/RFE-4177
-  wasm:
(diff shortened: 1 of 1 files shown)


jbtrystram:
We pull those from the 9.6 repos : 
https://github.com/openshift/os/blob/9093cb73ab43c0b03b85549c420f593cd6e86835/extensions/rhel-10.1.yaml#L13-L16

jlebon:
We should figure this out (either get it built for el10, or just drop it), though let's not block the el10 enablement work on this.

/lgtm

travier (approved):
/lgtm

PR #1826, opened 2025-06-24
OCPEDGE-1706: [TNF] OCP Two Node with Fencing symlink begone!

jaypoulz:
The latest builds of OCP 4.20 now come with an updated resource-agents RPM that includes the podman-etcd agent.

This means that the symlink now conflicts with the resource agent specified in the default path.


jlebon:
Woohoo, nice!

/approve
/lgtm

PR #1834, opened 2025-07-01
NO-JIRA: denylist: add ostree.remote to denylist for 4.18

aaradhak:
The ostree.remote test has been failing on 4.18 build, let us add this test to the denylist for now

Ref: https://github.com/coreos/rhel-coreos-config/issues/34


tlbueno:
I think it can be already closed as the infra is back

PR #1836, opened 2025-07-02
COS-3418: add kola-denylist for skipping kola tests

Roshan-R:
This adds the initial `kola-denylist.yaml` and manifest files enabling us to run kola and OpenShift-specific tests in the pipeline. The denylist includes `ext.config.version.rhaos-pkgs-match-openshift` which is currently failing.

Note: the tracker link included is a placeholder and will be updated once the issue is properly tracked.

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -1,0 +1,5 @@
+- pattern: ext.config.version.rhaos-pkgs-match-openshift
+  tracker: https://github.com/openshift/os/issues/1847
+  streams:
+    - 4.19-9.6
+    - 4.20-9.6


manifest-rhel-10.1.yaml
```diff
@@ -0,0 +1,8 @@
+metadata:
```
jlebon, line 1: Hmm, we shouldn't need these files. What are you hitting?

kola-denylist.yaml
```diff
@@ -0,0 +1,2 @@
+- pattern: ext.config.version.rhaos-pkgs-match-openshift
+  tracker: https://issues.redhat.com/projects/OCPBUGS/issues/OCPBUGS-58271
```
jlebon, line 2: That bug looks unrelated.

jbtrystram: Yeah, we pinged ART about this issues and put a placeholder bug in the meantime. I guess a comment would have been useful

manifest-rhel-10.1.yaml
jbtrystram: The issue is that kola run need the variables here to know what tests to skip in the denylist based on the `os` variable. Without these we were not able to have kola run the tests properly

Roshan-R: > Hmm, let's tweak kola instead to make this cleaner. E.g. something like `--denylist-stream 4.19-9.6`? And then in the `kola-denylist` we can use `stream:` as usual to filter entries. (And obviously, kola wouldn't error out any more if there is no `manifest.yaml` available).

AFIAK kola makes use of the manifest file to get the stream and also the osversion. Should we then add a `--denylist-osversion` flag to set the variable?

jlebon: We probably just need one of the two for our purposes. It's kinda arbitrary which one we use. I suggested `stream:` since that's closer to the OCP release streams nomenclature, but fine also with using `osversion:`, which matches what we use in rhel-coreos-config.

kola-denylist.yaml
jlebon: It could also just be a ticket against this repo.

jlebon (approved):
Nice!

jbtrystram (approved):

jbtrystram (approved):
/lgtm
/approve

jbtrystram:
/retitle COS-3418: add kola-denylist for skipping kola tests

PR #1837, opened 2025-07-02
OCPBUGS-47495: 40rhcos-fips: include fips.so in initrd

mike-nguyen:
Ignition panics when fips is enabled with golang 1.22. Golang 1.21 trusts openssl exclusively to tell whether FIPS mode is on or off but golang 1.22 looks at the system FIPS configuration directly. If the system is in FIPS mode but the initrd doesn't include the openssl FIPS module, golang 1.21 would assume it is not in FIPS mode whereas golang 1.22 sees that the openssl configuration doesn't match the system configuration and panics.


mike-nguyen:
Thanks for the review @jbtrystram !

PR #1842, opened 2025-07-08
COS-3424: Revert ostree.remote snooze - 4.12

tlbueno:
remove ostree.remote test as it is working again after fedora DC move completes

Ref: coreos/rhel-coreos-config#34


aaradhak (approved):

mike-nguyen:
This only affects tests and no production code so adding the jira/valid-bug label so it can merge.

PR #1848, opened 2025-07-17
NO-JIRA: extensions-rhel-9.4: Unpin `libreswan`

c4rt0:
Removing the pin to use the latest version of libreswan.

### DIFF
--- a/extensions-rhel-9.4.yaml
+++ b/extensions-rhel-9.4.yaml
@@ -15,9 +15,7 @@ extensions:
   # https://github.com/coreos/fedora-coreos-tracker/issues/1504
   ipsec:
     packages:
-      # pin to 4.6-3.el9_0.3 for now for https://issues.redhat.com/browse/OCPBUGS-43498
-      # we can revert once that's fixed in latest libreswan
-      - libreswan-4.6-3.el9_0.3
+      - libreswan
       - NetworkManager-libreswan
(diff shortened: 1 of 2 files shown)


jbtrystram:
The pin was removed on OVN : https://github.com/openshift/ovn-kubernetes/pull/2593/commits/764685a5631fffa992875afa49483b59062df75d

jbtrystram (approved):
/approve

c4rt0:
> 4.18 is based on RHEL 9.4. Should this be changing extensions-rhel-9.4.yaml?

Hmm, that's true. Perhaps in this PR I could add both, as work on https://issues.redhat.com/browse/OCPBUGS-43498 has been completed and addressed both rhel-9.4 & rhel-9.6 (4.19)

dustymabe:
> > 4.18 is based on RHEL 9.4. Should this be changing extensions-rhel-9.4.yaml?
> 
> Hmm, that's true. Perhaps in this PR I could add both, as work on https://issues.redhat.com/browse/OCPBUGS-43498 has been completed and addressed both rhel-9.4 & rhel-9.6 (4.19)

You can update them both, but only the rhel 9.4 one will get used. We should definitely retitle this PR to mention 9.4 and not 9.6.

sdodson:
/lgtm
Looks like all comments have been addressed

PR #1851, opened 2025-08-06
[release-4.19] NO-JIRA: tests/kola: drop rhel-matches-rhcos-build

jlebon:
This test is no longer valid since our versioning scheme changed:

https://github.com/openshift/enhancements/blob/master/enhancements/rhcos/split-rhcos-into-layers.md#etcos-release

It was removed in 4.20/master as part of
https://github.com/openshift/os/pull/1811 which would be nice to backport, but it's not strictly necessary either.

For now just cherry-pick this to 4.19.


jlebon:
I think let's just get this in to unbreak the pipeline.

Issue #1852, opened 2025-08-06
[4.19-9.6, 4.20-9.6]: s390x: `rhcos.network.init-interfaces-test` timing out


marmijo:
This test recently started running in the `build-node-image` job, see: https://github.com/coreos/fedora-coreos-pipeline/pull/1178, and has been failing since on the `s390x` arch for the `4.19-9.6` and `4.20-9.6` streams.

```
--- FAIL: rhcos.network.init-interfaces-test (611.87s)
         harness.go:1872: mach.Start() failed: machine "0756884f-e506-4bc6-b471-43b1c7ab7d3f" failed to start: ssh journalctl failed: time limit exceeded
```

#### Console Log Analysis
I've reviewed the console logs and believe the machine is actually booting successfully. The log for a test run shows the machine reaching a login prompt and configuring network interfaces.
- **Successful boot**: The log ends with a console login prompt, although for the `ibm-p8-kvm-03-guest-18` user, not the `core` user.
- **Network configuration**: The `rhcos.network.init-interfaces-test` test configures the network via Ignition. The log shows that  systemd successfully started the Network Manager service. The final log output also shows three network interfaces (`enc1.0001`, `enc1.0002`, and `enc3`) with assigned IP addresses, confirming the network was successfully configured and is operational.
- **coreos-teardown-initramfs**: There are some messages from this service: `RTNETLINK answers: Operation not supported` that I'm not sure are related. The console log shows that the system proceeds to successfully boot despite these messages.

Console.txt from `2025-08-06`:
- [console.txt](https://github.com/user-attachments/files/21627973/console.txt)

PR #1854, opened 2025-08-06
NO-JIRA: denylist: add `rhcos.network.init-interfaces-test` entry

marmijo:
This test has been failing since we started running openshift-specific kola tests in the `build-node-image` job. Let's denylist it for now so we can unblock the pipeline while we investigate
https://github.com/openshift/os/issues/1852

Also remove the stream definitions from the `ext.config.version.rhaos-pkgs-match-openshift` entry. An entry has been added to the `release-4.19` branch to properly denylist the test for 4.19-9.6. See: https://github.com/openshift/os/pull/1853


prestist (approved):
LGTM!

PR #1860, opened 2025-09-11
NO-JIRA: bump version to 4.21

mike-nguyen:
Note: c9s.repo and c10s.repo do not have an okd-4.21 repo yet so it will remain untouched.


mike-nguyen:
Fixed CI so this one needs another review

PR #1877, opened 2025-12-15
NO-JIRA: drop use of rhel-9.6-appstream-containernetworking repo

dustymabe:
cri-o has been rebuilt for RHEL 10 and nothing from this repo is getting pulled in any longer in node-image el10 builds. Let's drop this definition from the config.


dustymabe:
I'll also do a followup in rhel-coreos-config to see if I can get the label added back for SCOS.

PR #1880, opened 2026-01-08
[4.21] OCPBUGS-72560: package: add ose-crio-credential-provider

haircommander:
which is used to allow pull secrets to  be used for mirrored registries


travier (approved):
/lgtm
/label verified

PR #1888, opened 2026-01-15
[release-4.21] NO-JIRA: Remove (commented) WASM mentions

ravanelli:
- WASM support is no longer available in RHEL 9.8 and RHEL 10.0/10.1/10.2. Remove the corresponding extension mentions from RHCOS.


marmijo (approved):
/lgtm

PR #1894, opened 2026-02-10
NO-JIRA: extensions: add a label for extensions image discovery

dustymabe:
The MCO team was using the absence of the `ostree.linux` label to distinguish an extensions image from a node image in the release payload [1]. Let's make it easier to distinguish an extensions image by adding a label here.

[1] https://github.com/openshift/machine-config-operator/blob/6d48b64d76ed56318fbfdfc6a61c57ab93d96f9b/pkg/osimagestream/image_data.go#L15


jlebon (approved):
Seems fine to me!

PR #1895, opened 2026-02-12
NO-JIRA: packages-openshift: Fix repo versions

ravanelli:
- Fix wrong versions of repos under rhel-10.2

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -25,13 +25,9 @@ conditional-include:
       repos:
         - rhel-10.2-baseos
         - rhel-10.2-appstream
-        #- rhel-10.2-early-kernel
+        - rhel-10.2-early-kernel
         - rhel-10.2-fast-datapath
-        # XXX Not built for rhel 10 yet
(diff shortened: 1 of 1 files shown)


packages-openshift.yaml
```diff
@@ -30,8 +30,8 @@ conditional-include:
         # XXX Not built for rhel 10 yet
         #- rhel-10.2-fast-datapath
```
joelcapitao: maybe we should uncomment this. Also, note that `rhel-10.2-fast-datapath` is already defined above.

ravanelli:
> @ravanelli the extensions container build in CI is failing with:
> 
> ```
> error: Packages not found: kernel-devel-5.14.0-678.el9, kernel-headers-5.14.0-678.el9, kernel-rt-core-5.14.0-678.el9, kernel-rt-devel-5.14.0-678.el9, kernel-rt-modules-5.14.0-678.el9, kernel-rt-modules-extra-5.14.0-678.el9
> ```

It needs a new rhel-9.8 build to get the packages sync, it is running, almost at the end

Issue #992, opened 2022-09-07
Replace all use of Dockerfile by Containerfile


travier:
Let's make it explicit that this is not about Docker but about containers.

Conan-Kudo:
Not that I don't have a problem with this, but has anyone attempted to make _Docker/Moby_ support this filename by default too?

travier:
This is outside of the scope of this project issue tracker.

Conan-Kudo:
Then let's not bother renaming the files until the issue is resolved where it should be.

travier:
Changing the names that we use here is in scope. Changing the behavior of Docker is something else. We don't ship or use Docker in OCP or OKD.

PR #1896, opened 2026-02-13
NO-JIRA: Rename extensions/Dockerfile to Containerfile

cverna:
Replace use of Dockerfile with Containerfile to make it explicit that this is about containers, not Docker.

Fixes: https://github.com/openshift/os/issues/992

Assisted-by: OpenCode (Opus 4.5)


cverna:
Will need https://github.com/coreos/fedora-coreos-pipeline/pull/1297 after this is merged.

cverna:
/retitle NO-JIRA: Rename extensions/Dockerfile to Containerfile

PR #1900, opened 2026-02-27
COS-3912: denylist: drop rhcos.network.init-interfaces-test

nikita-dubrovskii:
Fix went upstream.

Issue: https://github.com/openshift/os/issues/1852


dustymabe:
> @marmijo FYI ^^ - we recently did some work on `ext.config.version.rhaos-pkgs-match-openshift` and I don't think that test is running at all. I think we'd need to add the `openshift` tag to that test?

https://github.com/openshift/os/pull/1906

Issue #1847, opened 2025-07-10
[4.19-9.6 and 4.20-9.6]: `ext.config.version.rhaos-pkgs-match-openshift` failure


Roshan-R:
The test fails due to package versions of rhaos packages not matching with the OpenShift version of the container.
 
The logs when testing against `4.20-9.6`
```
14:50:09  Jul 01 09:20:06 qemu0 systemd[1]: Started kola-runext-2.service.
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ NAME='Red Hat Enterprise Linux CoreOS'
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ VERSION='9.6.20250629-0 (Plow)'
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ ID=rhel
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ ID_LIKE=fedora
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ VERSION_ID=9.6
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ PLATFORM_ID=platform:el9
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ PRETTY_NAME='Red Hat Enterprise Linux CoreOS 9.6.20250629-0 (Plow)'
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ ANSI_COLOR='0;31'
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ LOGO=fedora-logo-icon
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ CPE_NAME=cpe:/o:redhat:enterprise_linux:9::baseos
14:50:09  Jul 01 09:20:06 qemu0 kola-runext-rhaos-pkgs-match-openshift[1655]: ++ HOME_URL=https://www.redhat.com/
[...]

PR #1920, opened 2026-03-31
[release-4.20] NO-JIRA: rhaos-pkgs-match-openshift: exclude known package mismatches

marmijo:
Introduce `RHAOS_PACKAGE_EXCEPTIONS` and build a regex from it to filter out packages that are known to not match the target version of openshift. Use a list-based structure so additional packages can be excluded in the future by appending to the exception array.

Exclude the toolbox and conmon rhaos packages, which are not expected to match.

Also remove the `rhaos-pkgs-match-openshift` test from the denylist.

See: https://github.com/openshift/os/issues/1847

This backports:
- https://github.com/openshift/os/commit/ef8a3bbd2f960958b36b64854d2ae5ad3c4b16a9
- https://github.com/openshift/os/commit/1abf0f6381b932c5b2a05bdb5d0407cfa76306cd

### DIFF
--- a/kola-denylist.yaml
+++ b/kola-denylist.yaml
@@ -2,5 +2,3 @@
 # coreos-assembler to automatically skip some tests. For more information,
 # see: https://github.com/coreos/coreos-assembler/pull/866.
 
-- pattern: ext.config.version.rhaos-pkgs-match-openshift
-  tracker: https://github.com/openshift/os/issues/1847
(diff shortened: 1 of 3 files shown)


Rolv-Apneseth (approved):
/lgtm

aaradhak:
@sdodson Do you think you ll be able to add these labels below for this PR to move forward?

/label backport-risk-assessed
/label jira/valid-bug

travier (approved):
/lgtm

PR #1923, opened 2026-04-03
OCPBUGS-77244: 9.8: override runc package to pull in newer fixes

haircommander:
Related to OCPBUGS-77244


travier (approved):
/lgtm

travier:
@haircommander Can you ping us / revert this PR once the build lands in RHEL? Maybe we should add some check here but I'm not sure how we could do that :/

sdodson:
@travier @haircommander I don't think this worked, or maybe we need to force a new 9.8 build even though there weren't changes for the versionlocks to be set properly? From the latest 4.22-9.8 node-image-build I just kicked off hoping to see this change effective

```
 14:23:02  Adding versionlock on: runc-4:1.4.0-2.el9.*
...
14:23:07  Package runc-4:1.4.0-2.el9.s390x is already installed.
...
repeat for other three arches
```

travier:
Looks like this landed in RHEL already: https://github.com/openshift/os/pull/1925

travier:
But yes, it looks like this did not work.

PR #1924, opened 2026-04-10
NO-JIRA: packages-openshift.yaml: Include runc only in rhel-9.8 and c9s

PeaceRebel:
runc doesn't exist in other versions, at least in rhel-10.2, causing build failure.

Fixes: https://github.com/openshift/os/pull/1923

PR #1925, opened 2026-04-10
NO-JIRA: Revert "OCPBUGS-77244: 9.8: override runc package to pull in newer fixes"

travier:
Reverts openshift/os#1923

See: https://github.com/openshift/os/pull/1924 & https://github.com/openshift/os/pull/1923#issuecomment-4201514875


travier:
CC @haircommander

travier:
This is still not using the right version:
```
2026-04-10T13:28:47.588753634Z Adding versionlock on: runc-4:1.4.0-2.el9.*
```

travier:
Ah, that's the one from the base image and we likely don't have an updated build yet to use here.
/unhold

PR #1929, opened 2026-04-13
WIP Updates for 4.23 node image Konflux build

thegreyd:
- Update ocp_version to 4.23 and switch rhel-9.8 repo IDs to match the repo names injected at build time in ocp-build-data - https://github.com/openshift-eng/ocp-build-data/pull/9978
- Drop rhel-9.8-early-kernel as the kernel is now covered by rhel-9-server-ose-rpms.

rh-pre-commit.version: 2.3.2
rh-pre-commit.check-secrets: ENABLED

### DIFF
--- a/packages-openshift.yaml
+++ b/packages-openshift.yaml
@@ -2,7 +2,7 @@ metadata:
   # This should match the /etc/os-release manipulation we do below when
   # injecting `OPENSHIFT_VERSION`. It's used by CI to determine the repos to
   # inject when building the layered image.
-  ocp_version: "4.22"
+  ocp_version: "4.23"
 
 conditional-include:
   - if:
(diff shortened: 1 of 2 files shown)


locriandev:
Green build: [rhcos-node-image-container-v4.23.0-202604220859.p2.gc621854.assembly.stream.el9](https://art-build-history-art-build-history.apps.artc2023.pc3z.p1.openshiftapps.com/build?nvr=rhcos-node-image-container-v4.23.0-202604220859.p2.gc621854.assembly.stream.el9&record_id=3c1a3176-55db-7de4-01b3-9e1f45ca61af&group=openshift-4.23&outcome=success&type=image)

packages-openshift.yaml
```diff
@@ -12,22 +12,9 @@ conditional-include:
     - osversion != "centos-10"
     include:
       repos: [ENOEXIST] # We want an error in this case
-  - if: osversion == "rhel-9.8"
```
travier, line 15: How do you inject the repos? Don't we still need them here for the CI or local builds?

locriandev: For Konflux, we inject them in the Dockerfile we produce by reconciling the upstream one, [here](https://github.com/openshift-priv/os/commit/491d1ae676c905b5046383d50f434de7f2a6f47d#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R14).

For CI builds, things get more complicated. For "normal" images, we use well-known CI images that we produce and publish to floating tags like `registry.ci.openshift.org/ocp/4.22:base-rhel9`. These images have a dnf wrapper script we use to inject the needed repos at build time. Do you need CI builds at all in this case? We would be mirroring the result of our Konflux builds to any imagestream you'd need.

thegreyd:
Closed in favor of https://github.com/openshift/os/pull/1932

PR #1940, opened 2026-06-04
NO-JIRA: Set io.openshift.os.streamclass label on built images

dustymabe:
This is currently being done in the build-node-image job [1]
but the more appropriate place to put it is in the openshift/os
definitions here.

Now that we have build-args we are more easily able to add
conditional configuration values like this so let's move the
configuration of this label from the build-node-image job
to the definitions here.

[1] https://github.com/coreos/fedora-coreos-pipeline/commit/b5669ddd1e6e24eacc01b2bf07522deb78bbc76c


extensions/centos-9.yaml
joelcapitao, line 1: With this commit we can also move https://github.com/openshift/os/blob/176d885f2a299d37b6d443deaedfae8059bd63e0/extensions/Containerfile#L13 to `stream10` for consistency and to avoid confusion.

joelcapitao (approved):
LGTM (my in-line comment is not a blocker)

joelcapitao:
/lgtm
waiting for https://github.com/openshift/release/pull/80122 to be be merged

docs/building.md
```diff
@@ -37,19 +37,31 @@ and metadata for that build. Choose the appropriate one for your target:
 To build:
 
 ```
-podman build . --build-arg-file build-args-c10s-5.0.conf \
+podman build --build-arg-file build-args-c10s-5.0.conf \
```
jbtrystram, line 40: I feel moving the `. ` to the end will make it more copy/paste error prone as it's easier to leave it out.. but that's doesn't matter too mcuch I guess

jbtrystram (approved):
LGTM

PR #1941, opened 2026-06-15
NO-JIRA: Add configs for 4.23; update build-args file naming

dustymabe:
With these configs we can now build the 10.2-4.23 and 9.8-4.23 node images.


ravanelli (approved):
/lgtm

Issue #1911, opened 2026-03-05
kata-containers package is missing in rhel-10.2-server-ose-4.22 repo


ravanelli:
The kata-containers package is not currently available in the RHEL-10.2 repository

PR #1942, opened 2026-06-17
NO-JIRA: extensions: templatize OCP version in repo names; drop kata from extensions

dustymabe:
Replace hardcoded OCP version in server-ose repo names with an __OCP_VERSION__ placeholder that gets substituted at build time via OPENSHIFT_VERSION (already defined in build-args-*.conf files).

This allows the same extensions YAML files to be used for building against different OCP versions (e.g. 4.23 and 5.0) without duplicating the entire file just to change a repo name.

Assisted-by: <anthropic/claude-opus-4.6>

---

extensions: drop kata containers from extensions for now

The package is rarely in the places we need it and when we open requests
to get it built for newer versions that go unanswered [1][2].
    
Drop it until someone cares enough to fix it.
    
[1] https://github.com/openshift/os/issues/1911
[2] https://issues.redhat.com/browse/KATA-4726

### DIFF
--- a/extensions/Containerfile
+++ b/extensions/Containerfile
@@ -5,6 +5,7 @@ RUN mkdir /os
 WORKDIR /os
 ADD . .
 ARG OPENSHIFT_CI=0
+ARG OPENSHIFT_VERSION=overridden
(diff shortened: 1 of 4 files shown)


Rolv-Apneseth (approved):
/lgtm

PR #1943, opened 2026-06-18
ART-19543: extensions: use dnf for ART builds, keep rpm-ostree for prow/local

locriandev:
## Summary

- Switch ART builds to use `dnf download` instead of `rpm-ostree compose extensions`
- Keep prow and local builds using `rpm-ostree compose extensions` as before
- Resolves repo naming conflicts for ART builds that inject repos via `/etc/yum.repos.d/`

## Details

ART builds inject repository configurations via `/etc/yum.repos.d/` which have different names than those defined in the extensions YAML files. This mismatch causes `rpm-ostree compose extensions` to fail because it expects exact repo name matches.

By switching to `dnf download`, we allow ART builds to use whatever repo names are present in the environment, while prow and local builds continue with the existing `rpm-ostree` workflow unchanged.

### DIFF
--- a/extensions/build.sh
+++ b/extensions/build.sh
@@ -6,16 +6,36 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
     ci/get-ocp-repo.sh ocp.repo
 fi
 
-# just to parse the treefile, rpm-ostree still wants to read referenced "externals" (e.g. passwd, group)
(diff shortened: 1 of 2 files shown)


locriandev:
Superseded by https://github.com/openshift/os/pull/1945

PR #1945, opened 2026-06-24
ART-19543: Refactor extensions build to use dnf download instead of rpm-ostree compose

locriandev:
## Summary
Refactors the extensions build process to use `dnf download` directly instead of `rpm-ostree compose extensions`, improving flexibility and simplifying the build process for RHEL 10.2 and other OS streams.

### Build Script Refactor (`extensions/build.sh`)
- **Replace rpm-ostree**: Switched from `rpm-ostree compose extensions` to direct `dnf download` commands
- **Parameterized repos**: Introduced `YUM_REPO_NAMES` build arg to specify repos at build time
- **Direct RPM downloads**: Each extension now downloads RPMs directly with `--resolve --alldeps`
- **Architecture-aware**: Handles arch-specific extensions (kernel-rt for x86_64, kernel-64k for aarch64)
- **Version pinning**: Kernel-related packages pinned to installed kernel EVR to ensure compatibility

### New Build Configuration Files
- `build-args-5.0-9.8.conf`: Repo names for RHEL 9.8
- `build-args-5.0-10.2.conf`: Repo names for RHEL 10.2
- `build-args-5.0-c10s.conf`: Repo names for CentOS Stream 10

### Containerfile Update
- Added `ARG YUM_REPO_NAMES=overridden` to accept repo names as build argument
- Maintains backward compatibility with secret.repo mount

### Removed Files
- Deleted YAML extension definitions (no longer needed with dnf approach):
  - `extensions/rhel-9.8.yaml`
  - `extensions/rhel-10.2.yaml`
  - `extensions/centos-10.yaml`


travier:
I would prefer to keep the split between the code/script doing the download and the configuration file with the extensions for each varsion/variant.

This is mentioned in:

> Replace the __OCP_VERSION__ placeholder with the actual OpenShift version.
> This allows the same YAML file to be used across different OCP versions
> (e.g. 4.23 and 5.0) without duplication.

The script could be a dnf download function with all the parameters and a loop over the entries in the YAML config.

extensions/build.sh
```diff
@@ -6,16 +6,89 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+# Helper function to download packages for an extension
+download_extension() {
+    local extension_name=$1
+
+    # Get packages for this extension from JSON, performing variable substitution
+    local packages
+    packages=$(jq -r ".extensions[\"${extension_name}\"].packages[]" "$extensions_json" | while read -r pkg; do
+        # Perform variable substitution
+        eval "echo \"$pkg\""
+    done)
```
dustymabe: not sure why we need the while loop here, but can probably fix it later.

extensions/build.sh
```diff
@@ -6,16 +6,89 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+arch=$(uname -m)
+destdir=/usr/share/rpm-ostree/extensions/
+mkdir -p "${destdir}"
+
+# Determine which extensions JSON file to use based on OS
+extensions_json="extensions/${ID}-${VERSION_ID}.json"
+
+# Get kernel version for packages that need to match the base kernel
+# Include epoch (0:) so dnf can disambiguate name from version in NEVRA format
+kernel_evr=$(rpm -q --queryformat '%{VERSION}-%{RELEASE}' kernel)
```
dustymabe: I think we should just versionlock everything in the node image and not worry about `kernel_evr` substitution here.

i.e I think a `dnf versionlock` should do it like we do for the node image:

https://github.com/openshift/os/blob/d75a447b705c33cb96a43355eae4355c609d0784/build-node-image.sh#L35-L36

extensions/build.sh
```diff
@@ -6,16 +6,89 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+    local packages
+    packages=$(jq -r ".extensions[\"${extension_name}\"].packages[]" "$extensions_json" | while read -r pkg; do
+        # Perform variable substitution
+        eval "echo \"$pkg\""
+    done)
+
+    # Download the packages
+    if [ -n "$packages" ]; then
+        echo "Downloading extension: ${extension_name}"
+        dnf --repo="${YUM_REPO_NAMES}" download --resolve --alldeps \
```
dustymabe: I'm not sure we need `--alldeps` here.

When you do a build with this does `glibc` and `glibc-devel` get downloaded? or just `glibc-devel`?

extensions/build.sh
```diff
@@ -6,16 +6,89 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+
+    # Get packages for this extension from JSON, performing variable substitution
+    local packages
+    packages=$(jq -r ".extensions[\"${extension_name}\"].packages[]" "$extensions_json" | while read -r pkg; do
+        # Perform variable substitution
+        eval "echo \"$pkg\""
+    done)
+
+    # Download the packages
+    if [ -n "$packages" ]; then
```
dustymabe: If there were no packages to download we should error.

extensions/build.sh
```diff
@@ -6,16 +6,89 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+            fi
+        done
+        if [ "$variant_match" = false ]; then
+            echo "Skipping ${extension} (not for ${ID})"
+            continue
+        fi
+    fi
+
+    # Download the extension packages
+    download_extension "$extension"
```
dustymabe: I would argue that we should only call `dnf download` once instead of multiple times, but we can fix that later.

extensions/build-args-5.0-10.2.conf
```diff
@@ -0,0 +1,2 @@
+# The names of the yum repos to use for the extensions image build.
+YUM_REPO_NAMES=rhel-10.2-baseos,rhel-10.2-appstream,rhel-10.2-server-ose-5.0,rhel-10.2-highavailability,rhel-10.2-fast-datapath,rhel-10.2-nfv
```
dustymabe: Since f6d3dba we use the top level build args file for extensions image builds too so let's not create a new build-args file here. Let's add this info there and use a different var name:

```
EXTENSIONS_YUM_REPO_NAMES=
```

?

dustymabe (commented):
A step in the right direction. I can see a vision for how we can achieve the goal here cleanly.

Here's an initial round of review, but I'll have more questions as we go.

extensions/build.sh
locriandev: Done

extensions/build.sh
locriandev: Removed

extensions/build.sh
locriandev: Done

extensions/build.sh
locriandev: Done

extensions/build-args-5.0-10.2.conf
locriandev: Done

extensions/build.sh
```diff
@@ -6,16 +6,121 @@ if [ "${OPENSHIFT_CI}" != 0 ]; then
...
+            if [ "$ID" = "$ext_variant" ]; then
+                variant_match=true
+                break
+            fi
+        done
+        if [ "$variant_match" = false ]; then
+            echo "Skipping ${extension} (not for ${ID})"
+            continue
+        fi
+    fi
```
travier: IIUC, that should let us merge all the extensions definitions into a single YAML file?
It does not appear to be used in the config right now.

travier (commented):
Thanks, this is looking great.

extensions/build.sh
locriandev: Yes, the `variants` field enables merging all extension definitions into a single YAML in the future.  I'm deferring the actual YAML consolidation to keep this change focused. We can migrate to a unified file in a follow-up once this filtering logic is proven.

travier: OK, then if it's not used yet then let's remove this for now and we can add the code when it's used?

travier (commented):
One nit but LGTM. CI is failing on the repos not being properly overridden via the build-arg missing. Not sure what happens.

extensions/build.sh
locriandev: Done with c46c6192857095e58c2b9ec4e9a6faf7d1346ff2

travier (approved):
/lgtm
/hold until the release PR is merged
Thanks

dustymabe:
Starting reviewing this today. I found some things that need to be fixed but can possibly do those as a followup if you want to merge this before.

dustymabe:
ok I pushed a bunch of fixes here (see individual commit messages). Highlights are:

1. dnf4 `dnf download` doesn't respect versionlocking :(
2. Needed to replace `match-base-evr` logic
3. Housekeeping on c10s repos
4. some docs updates

travier (approved):
/lgtm

dustymabe:
/unhold

release PR https://github.com/openshift/release/pull/81704 merged

PR #1952, opened 2026-08-05
OCPBUGS-104572: NetworkManager-ovs has moved up to RHCOS

sdodson:
NetworkManager-ovs was being installed in the node-image layer (openshift/os) from OCP repos. Because the base image version-locks all its packages before the node layer runs, NetworkManager-ovs could end up at a different version than NetworkManager itself. Move it to the base image so both are sourced from the same RHEL repos and stay in sync.

Depends on https://github.com/coreos/rhel-coreos-config/pull/296


sdodson:
/hold
until the dependent PR merges

Rolv-Apneseth (approved):
/lgtm

aaradhak (approved):

PR #1955, opened 2026-08-07
Revert "Revert "OCPBUGS-104572: NetworkManager-ovs has moved up to RHCOS""

sdodson:
Reverts openshift/os#1954


sdodson:
Now that I've verified that both rhel-9.8 and rhel-10.2 have NM-ovs added at their layers it's safe to pull this back in.

PR #1958, opened 2026-08-10
NO-JIRA: Use --nobest when installing node packages

dustymabe:
Even if the RHCOS base image we are FROMing is a little out of date let's continue best effort if the older versions of the packages we need (based on versionlocking) are available in the repos.

### DIFF
--- a/build-node-image.sh
+++ b/build-node-image.sh
@@ -36,7 +36,10 @@ mkdir -p /var/opt
 dnf --disablerepo=* versionlock add '*'
 
 # Install the OCP packages. Limit to appropriate repos for this stream.
-dnf --repo="${YUM_REPO_NAMES}" install -y \
+#
+# Leverage --nobest so older versions of packages can be used when
+# versionlocking would require that.
+dnf --repo="${YUM_REPO_NAMES}" install --nobest -y \
     cri-o cri-tools conmon-rs \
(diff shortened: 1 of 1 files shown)


sdodson:
/jira backport release-4.22,release-4.21,release-4.20,release-4.19

sdodson:
Hmm, I thought this would've fixed this error. https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_os/1958/pull-ci-openshift-os-master-images/2086825258005827584#1:build-log.txt%3A1072

?

dustymabe:
> Hmm, I thought this would've fixed this error. https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_os/1958/pull-ci-openshift-os-master-images/2086825258005827584#1:build-log.txt%3A1072
> 
> ?

That particular error is from the extensions image build, which [already has --nobest](https://github.com/openshift/os/blob/bf90b219ae4ba42e07bf8c010b725401b5402cc8/extensions/build.sh#L115).

The only explanation I have is that the prow repos don't have the older versions of the kernel in them.

dustymabe:
Yeah. the switch to using `dnf` here for the node image and extensions image only happened recently and won't apply to older branches.

PR #1966, opened 2026-09-05
NO-JIRA: extensions/Dockerfile: add extensions.json to last layer

cverna:
Backport of commit 31816ac from master. On release branches the extensions container is still built by coreos-assembler, which bind-mounts /tmp/extensions.json into the build to propagate the extensions metadata into the build's meta.json, so keep writing to that path.

In addition, bake extensions.json into the final container image, so that it is easy for anyone but particularly ART to query metadata about the extensions container, e.g. via `oc image extract` or `podman run $img cat /usr/share/rpm-ostree/extensions.json`.

Assisted-by: AI


cverna:
This would be useful to help with verifying CVEs in the extensions container, testing on 4.18 before backporting further

jmarrero (approved):

